PR #10, opened 2022-08-01
update writer

giovp:


spatialdata/_core/_spatialdata.py
```diff
@@ -1,59 +1,83 @@
...
-        self.points = points
+
+        elems = []
+
+        images, images_transform, elem_images = _validate_dataset(images, images_transform)
+        labels, labels_transform, elem_labels = _validate_dataset(labels, labels_transform)
+        points, points_transform, elem_points = _validate_dataset(points, points_transform)
+        shapes, shapes_transform, elem_shapes = _validate_dataset(shapes, shapes_transform)
+
+        if images is not None and images_transform is not None:
```
LucaMarconato, line 57: it should never happen that one is None and the other not, I would replace this by 
if (... is None) + (... is None) % 2 == 0:
    code in the if block
else:
    exception
    
same for the ifs belows

spatialdata/_constants/_constants.py
```diff
@@ -0,0 +1,11 @@
+from enum import unique
+
+from spatialdata._constants._enum import ModeEnum
+
+
+@unique
+class ShapesType(ModeEnum):
```
LucaMarconato, line 7: to discuss tomorrow if we support only circles or not

spatialdata/_core/_spatialdata.py
```diff
@@ -1,59 +1,83 @@
...
+    shapes: Optional[Mapping[str, AnnData]]
+    elems: List[Optional[Set[str]]]
 
     def __init__(
         self,
-        adata: Optional[AnnData] = None,
-        regions: Mapping[str, Any] = MappingProxyType({}),
-        images: Mapping[str, Any] = MappingProxyType({}),
+        tables: Optional[AnnData] = None,
+        labels: Optional[Mapping[str, Any]] = MappingProxyType({}),
```
LucaMarconato, line 40: To decide if we will add polygons and "star regions" as 2 (+ 2 because of the transforms) different arguments, or if we just allow for a mapping of regions and a mapping of transforms

spatialdata/_core/_spatialdata.py
```diff
@@ -1,59 +1,83 @@
...
+        labels_transform: Optional[Mapping[str, Any]] = None,
+        images: Optional[Mapping[str, Any]] = MappingProxyType({}),
         images_transform: Optional[Mapping[str, Any]] = None,
-        points: Optional[AnnData] = None,
+        points: Optional[Mapping[str, AnnData]] = None,
```
LucaMarconato, line 44: from the specs points was not a mapping, but we can change it (I think it makes sense to have it a mapping)

spatialdata/_core/_spatialdata.py
```diff
@@ -206,10 +277,10 @@ def h(s: str) -> str:
 
         if self.points is not None:
             n = 1
-            descr += f"\n{h('level0')}points with n_obs x n_vars = {self.points.n_obs} x {self.points.n_vars}"
+            # descr += f"\n{h('level0')}points with n_obs x n_vars = {self.points.n_obs} x {self.points.n_vars}" # TODO: returns error atm.
```
LucaMarconato, line 280: gonna have a look

spatialdata/_core/mixin/io_mixin.py
```diff
@@ -1,15 +1,13 @@
 from abc import abstractmethod
-from typing import Any, Optional
+from typing import Any
 
 
 class IoMixin:
```
LucaMarconato, line 5: Initially this class would have been associated to "Regions" and all the explicit building blocks that can be written to and from zarr. Since atm we don't have such classes (because we use the first-dispatch strategy), this class seems superfluous. But it will be useful when of we implement `SpatialDataPool` (or if we go for having classes for each building block).

spatialdata/_core/_spatialdata.py
giovp: makes sense, updarted

spatialdata/_core/_spatialdata.py
giovp: to be discussed, let's keep it for now

spatialdata/_core/_spatialdata.py
giovp: agree, let's keep it a mnapping

spatialdata/_core/mixin/io_mixin.py
giovp: to me it can be removed for now, but feel free to keep it if you think it's useful!

giovp:
@LucaMarconato think I finished it, there are many todos still and lots to work on but I would merge this in your branch for now as it contains a number of updates on testing and pre-commits. RE: pre-commits I really think we should already work with pre-commit checks active.

random list of things I'd prioritize next:
- [ ] move repr out of the class in separate module (see how it's done in mudata, think it's nice)
- [ ] reader (and transformation)
- [ ] get transformations out of objects when writing and save them according to ome-zarr-py
- [ ] cleanup (temp folder, napari viewer etc.)
   - If you think napari is very important to have it working at this stage, we could consider to have a dev branch in napari-spatialdata and start building minimal interface there.
- [ ] minimal test for Transform
  - Transform seems a pretty important class that is prob worth discussing in details.
  - I really like that it's a separate class so it'd be easier to incorporate changes from upstream (e.g. ngff specs)

I think it'd make sense to split this, let me know what you wish to take up and I'll take the rest!

giovp:
p.s. CI tests are not runnning cause the PR is not against main, they should run once this is merged

spatialdata/_core/_spatialdata.py
```diff
@@ -27,57 +26,57 @@
...
+        if labels is not None:
             self.labels = {
                 k: self.parse_image(labels, labels_transform)
                 for k, (labels, labels_transform) in zip(labels.keys(), zip(labels.values(), labels_transform.values()))
             }
             elems.append(elem_labels)
 
-        if points is not None and points_transform is not None:
+        if points is not None:
             self.points = {k: self.parse_tables(points[k], points_transform[k]) for k in points.keys()}
```
LucaMarconato, line 72: I checked all the changes, all good. I will just rename parse_tables to parse_points

LucaMarconato:
Great job, just merged in the `io` branch

PR #14, opened 2022-08-14
Refactor writer and add reader

giovp:


giovp:
ok, I think there is something to work with, to try it out:

```python
from spatialdata._core.spatialdata import SpatialData
import numpy as np
from anndata import AnnData

rng = default_rng()
sdata = SpatialData(
    images={
        "image1": rng.normal(size=(5, 100, 100)),
        "image2": rng.normal(size=(5, 100, 100)),
    },
    labels={"image1": rng.normal(size=(100, 100))},
    points={"image1": AnnData(rng.normal(size=(200, 200)), dtype=np.float_)},
    tables={"image1": AnnData(rng.normal(size=(100, 100)), dtype=np.float_)},
)


zarr_store = "./mydata.zarr"

# write
sdata.write(zarr_store)

# read
from spatialdata._io.read import read_zarr

sdata = read_zarr("./mydata.zarr")
```

important points to consider
- [ ] transforms are not considered ever, highest priority is to add them, and follow as close as possible how they are stored/read by ngff
- [ ] right now tables are copied to same element (what we used to call building block) and also stored in `sdata.tables`. I think it'd make sense to decide now re views v. copy.
- [ ] Feedback welcome everywhere ofc, but the thing I'm the least confident about is reading. Again, I've tried to follow as close as possible the classes/readers of `ome_zarr_py` but I have many doubts whether I made the right choices.

polygons are not implemented neither for reading or for writing.

This PR supersedes #11 and if there is an agreement on high level design choices, I'd suggest to merge it and split workload again.

spatialdata/_core/elements.py
```diff
@@ -0,0 +1,177 @@
...
+        """Construct Element from file."""
+        # TODO: maybe split into local and remote?
+
+    @abstractproperty
+    def shape(self) -> Tuple[int, ...]:
+        """Return shape of element."""
+
+
+class Image(BaseElement):
+    def __init__(self, image: ArrayLike, transform: Transform) -> None:
```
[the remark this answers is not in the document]
kevinyamauchi: actually, I now see that the `parse_image` method below does coerce the array to a `xarray.DataArray`. I think we should update the type then.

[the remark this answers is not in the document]
kevinyamauchi: also, the type of `self.data` should be declared.

spatialdata/_constants/_key.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, f: Callable[..., str]):
+        self.f = f
+
+    def __get__(self, obj: Any, owner: Any) -> str:
+        return self.f(owner)
+
+
+# TODO: use this class for various properties
+class Key:
+    """Class which manages keys in :class:`anndata.AnnData`."""
```
kevinyamauchi, line 14: What is this for?

spatialdata/_constants/_enum.py
```diff
@@ -0,0 +1,66 @@
...
+    __error_format__ = "Invalid option `{0}` for `{1}`. Valid options are: `{2}`."
+
+    @classmethod
+    def _format(cls, value: Enum) -> str:
+        return cls.__error_format__.format(
+            value, cls.__name__, [m.value for m in cls.__members__.values()]  # type: ignore[attr-defined]
+        )
+
+
+class ModeEnum(str, ErrorFormatterABC, PrettyEnum, metaclass=ABCEnumMeta):
```
kevinyamauchi, line 65: Not a blocker, but it feels overly complex to use multiple inheritance for an Enum that tracks mode. I think this can be refactored into a more simple structure that just inherits from `PrettyEnum` if the error formatter method is added directly to `PrettyEnum` (i.e., without a mixin) and the `ABCEnumMeta` is part of `PrettyEnum`. If we find there's a need for more granular subclasses, we can move there, but I don't think we want to start out with such complexity.

spatialdata/_core/spatialdata.py
```diff
@@ -1,25 +1,242 @@
...
+
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, SpatialData):
+            return False
+        # new comparison: dumping everything to zarr and comparing bytewise
+        # TODO: do the same but in memory
+        with tempfile.TemporaryDirectory() as tmpdir:
+            self.write(os.path.join(tmpdir, "self.zarr"))
+            other.write(os.path.join(tmpdir, "other.zarr"))
+            return are_directories_identical(os.path.join(tmpdir, "self.zarr"), os.path.join(tmpdir, "other.zarr"))
```
kevinyamauchi, line 174: I recognize this is a temporary solution, but I think I would rather not have an `__eq__` method than have one that writes objects to disk. 

In general, I think we want the `__eq__` methods that do the comparison to live on the elements and this `SpatialData` `__eq__` just call those.

kevinyamauchi (commented):
Thanks @giovp and @LucaMarconato !

Overall, I think this is a good first step. I have left some general comments below and some more specific ones inline. In the future, this would be easier to review as multiple, smaller PRs (e.g., starting with just elements and `SpatialData` and following up with IO).


### Complexity
I am a bit concerned about the amount of complexity we've already introduced. In general, it's much easier to add complexity than reduce it, so I feel we should try to be as simple as possible at this early stage. A few places where we can reduce complexity:

- **type coercion**: for now, I think we should avoid automatically converting too many input types. In general, I find that it adds bloat and can often lead to odd behaviors due to the implicit assumptions that inevitably made. We can add helper conversion functions in the future once we understand the main pathways. 
- **multiple inheritance**: I don't think we really need multiple inheritance for now. We can add it later if we find that we are duplicating a ton of code. It is much more difficult to undersand and maintain code that is inherited from several places.

### Construction pattern

What is our preferred construction pattern for a `SpatialData` object when the data are already in memory (i.e., not from an OME-NGFF file)? Should the user first create Element objects (e.g., `Image`, `Labels` and then pass those to the `SpatialData` constructor? Or should they pass some more basic objects (e.g., numpy arrays) directly to the `SpatialData` constructor and then the constructor does coercion/construction into our Element types?

Personally, I would prefer that for now, we require the user to construct the Element objects and then pass those to the SpatialData constructor. I think we should not implement too much coercion for now, as allowing multiple types and inferring how to coerce them always adds complexity. We can always add convenience functions later, but for now, I think we should focus on making the core solid.

### single dispatch
I am curious about the rationale for using `singledispatch` over a switch case or dictionary based on type? Personally, I find `singledispatch` more difficult to follow the different code branches (e.g., when using a switch case or dictionary, the IDE can link you to the exact function). Is it because we want people to be able to register their own implementations?

If we do use `singledispatch`, can we commit to keeping all of the registered functions together in a single file? Also, can we give the registered names? I think this will help with the readability of the code.

I don't feel strongly about this - I am mostly just curious if there are specific advantages to using `singledispatch`

### Next steps

I think we can merge this pretty quickly. I would like some feedback on my questions/comments, but I don't see any of them as big blockers.

Next, I think we should make a short "design document" that formalizes the `SpatialData` design principles and gives pseudocode examples of how it would be used in an analysis pipeline.  I think we have a lot of pieces of this around, but we should consolidate and make it into a document here in the repo. We should also include a glossary that defines key terms.

Here is a nice example of a design document (we don't have to be this detailed/complete for now): https://napari.org/dev/naps/4-async-slicing.html

spatialdata/_core/transform.py
```diff
@@ -0,0 +1,105 @@
...
+            if scale_factors is not None:
+                assert len(scale_factors) == ndim
+
+        self.translation = translation
+        self.scale_factors = scale_factors
+        self.ndim = ndim
+
+
+@singledispatch
+def set_transform(arg: Any, transform: Optional[Transform]) -> None:
```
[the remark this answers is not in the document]
LucaMarconato: It's from the old code, now it can be removed (now the transform data is stored in the BaseElement, before it was stored inside the object itself, e.g. inside the xarray object)

spatialdata/_core/spatialdata.py
LucaMarconato: I agree, but comparing anndata objects is not possible, so at least for that we may need to compare files on disk.

LucaMarconato (approved):

spatialdata/_core/spatialdata.py
kevinyamauchi: > comparing anndata objects is not possible

@LucaMarconato , can you please expand on this? Do you mean we would need to contribute an `AnnData.__eq__()` method upstream to anndata? Or is there some other limitation?

In any case, I could imagine it would be slow since there's potentially a lot of data to compare. Perhaps we should just remove this method?

giovp: jumping in only on this comment: indeed anndata doesn't have such method, so we'd ahve to do it upstream (although I believe there is plans to do it @ivirshup ). Regardless, I agree we;d want to remove it eventually, but for now I think it can turns out quite handy for prototyping ?

spatialdata/_constants/_enum.py
giovp: ok, I think it's a nice enum improvement (got it from squidpy) for printing the available enums when input is wrong. I'll try to simplify. It's also a minor part of the code that doesn't really relate to anything else beside enum

spatialdata/_constants/_key.py
giovp: this is something that turned out to be quite handy for a bunch of use cases (specifically accessing anndata slots). For now is unused but I wanted to put it there for the first occasion to use it.

spatialdata/_core/elements.py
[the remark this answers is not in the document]
giovp: agree to update the type.

spatialdata/_core/transform.py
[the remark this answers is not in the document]
giovp: are we sure we want to remove all code from transform? can do that, didn't do cause I didn't touch it in this PR (beside pre-commit)

spatialdata/_core/transform.py
```diff
@@ -0,0 +1,105 @@
...
+from typing import Any, Optional
+
+import anndata as ad
+import numpy as np
+import xarray as xr
+
+from spatialdata._types import ArrayLike
+
+
+class Transform:
```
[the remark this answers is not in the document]
giovp: this is very nice, 100% agree, I'll try to set it up

spatialdata/_core/spatialdata.py
```diff
@@ -1,25 +1,242 @@
...
+            }
+
+        if tables is not None:
+            # TODO: concatenate here? do views above?
+            self.tables = {k: self.parse_table(tables[k]) for k in tables.keys()}
+
+    @classmethod
+    def parse_image(cls, data: Any, transform: Optional[Any] = None) -> Image:
+        data, transform = parse_dataset(data, transform)
+        return Image(data, transform)
```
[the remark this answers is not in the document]
giovp: would work for me, will do!

giovp:
thanks a lot @kevinyamauchi for the comments, super useful, here's my reply:

> Complexity
> I am a bit concerned about the amount of complexity we've already introduced. In general, it's much easier to add complexity than reduce it, so I feel we should try to be as simple as possible at this early stage. A few places where we can reduce complexity:

I agree, I'll try to reduce as per your comments in code.

> type coercion: for now, I think we should avoid automatically converting too many input types. In general, I find that it adds bloat and can often lead to odd behaviors due to the implicit assumptions that inevitably made. We can add helper conversion functions in the future once we understand the main pathways.
multiple inheritance: I don't think we really need multiple inheritance for now. We can add it later if we find that we are duplicating a ton of code. It is much more difficult to undersand and maintain code that is inherited from several places.

I agree but I was trying to come from the ome-ngff side e.g. since while reading we always have dask arrays (as per ngff) then shouldn't we always set with a standard for arrays (e.g. xarray) ?

> single dispatch
> ...
> Is it because we want people to be able to register their own implementations?

yes, in my opinion this is the main reason for it.

> If we do use singledispatch, can we commit to keeping all of the registered functions together in a single file? Also, can > we give the registered names? I think this will help with the readability of the code.

> I don't feel strongly about this - I am mostly just curious if there are specific advantages to using singledispatch

👍 💯 I personally like single dispatch, find it clearer than dictionaries or if switches but understand maybe it's overkill in some cases here.

> Next steps
> I think we can merge this pretty quickly. I would like some feedback on my questions/comments, but I don't see any of them as big blockers.

> Next, I think we should make a short "design document" that formalizes the SpatialData design principles and gives pseudocode examples of how it would be used in an analysis pipeline. I think we have a lot of pieces of this around, but we should consolidate and make it into a document here in the repo. We should also include a glossary that defines key terms.

> Here is a nice example of a design document (we don't have to be this detailed/complete for now): https://napari.org/dev/naps/4-async-slicing.html

💯 💯 💯  this is great, I think we can start it before the call next Thu and populate it during/after that.

I'll adress the comments and merge this if you agree, didn't find any major blocker from comments. 

One main thing that I think we need to settle (prob also at call this thu) is the name/number of classes in `Elements` (what we used to call `BuildingBlocks`, changed the name cause it's shorter, ofc we can change it again). Now we have:
- Image
- Labels
- Points (both for transcripts and indexes)
- Polygons
- Tables

imho this is a sound selection that mirrors nicely with ngff and napari. I haven't seen any comment on this so far but just wanted to raise it once again.

spatialdata/_core/elements.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # store the transform objects as a dictionary with keys (source_coordinate_space, destination_coordinate_space)
+    transforms: Optional[Transform] = None
+
+    @abstractmethod
+    def to_zarr(self, group: zarr.Group, name: str, scaler: Optional[Scaler] = None) -> None:
+        """Write element to file."""
+
+    @abstractmethod
+    def transform(self, new_coordinate_space: str, inplace: bool = False) -> "BaseElement":
```
kevinyamauchi, line 37: What do you think about changing this to `transform_to()`? I think `BaseElement.transforms` is a bit to similar to `BaseElement.transform()`.

```python

# get the local to global transformation
global_transform = image.transforms["local_to_global"]

# transform to the global coordinate system
new_image = image.transform_to("global", in_place=False)
```

spatialdata/_core/spatialdata.py
kevinyamauchi: Overall, my concern here is that the `__eq__` method is part of the public API and I think this implementation is really not ideal (e.g., this will cause problems if a user calls this on reasonably large data). Even though we are early stage, my understanding is that we want to shop `SpatialData` around to potential users relatively soon. Thus, I think we need to be careful about what we put on our public API. 

> indeed anndata doesn't have such method, so we'd ahve to do it upstream (although I believe there is plans to do it @ivirshup ). Regardless, I agree we;d want to remove it eventually, but for now I think it can turns out quite handy for prototyping ?

Can you expand on what you would like to use this for? Could this be a private utility function? If anndata doesn't have an `__eq__` implementation, that is a fairly strong signal to me that we don't need one either (yet).

spatialdata/_constants/_key.py
kevinyamauchi: Got it. Yes, please remove it before merging. I think we want to start out as lean as possible to reduce the clean up work we need to do in the future.

kevinyamauchi:
Thanks for the feedback, @giovp !

> I agree, I'll try to reduce as per your comments in code.

Thanks!

> 👍 💯 I personally like single dispatch, find it clearer than dictionaries or if switches but understand maybe it's overkill in some cases here.

Okay, works for me. Happy to give it a go!

> I agree but I was trying to come from the ome-ngff side e.g. since while reading we always have dask arrays (as per ngff) then shouldn't we always set with a standard for arrays (e.g. xarray) ?

For loading OME-NGFF from disk, I think we can/should do the type coercion because we know exactly how to do it (thanks to the specification). The point I was trying to make is that I want to avoid "automagical" type coercion in the SpatialData and Element constructors for now because it generally requires implicit assumptions about the input data. For example, in the test below, what is the coordinate space and what are the axis names for `"image"`?

https://github.com/scverse/spatialdata/blob/9906b438c0b726aae2e608706491cfd3d86c6b84/tests/_core/test_instantiation.py#L30-L33


> One main thing that I think we need to settle (prob also at call this thu) is the name/number of classes in Elements (what we used to call BuildingBlocks, changed the name cause it's shorter, ofc we can change it again). Now we have:
>
>   Image
>    Labels
>    Points (both for transcripts and indexes)
>    Polygons
>    Tables

This looks good to me! I agree it would be good to get some feedback at the community meeting.

imho this is a sound selection that mirrors nicely with ngff and napari. I haven't seen any comment on this so far but just wanted to raise it once again.

> I'll adress the comments and merge this if you agree, didn't find any major blocker from comments.

Sounds good! Please make an issue for any comments you do not address (fine to punt if you think it's too much for now - I just don't want to lose track of things). Thanks!

kevinyamauchi (approved):
> I'll adress the comments and merge this if you agree, didn't find any major blocker from comments.

Sounds good! Please make an issue for any comments you do not address (fine to punt if you think it's too much for now - I just don't want to lose track of things). Thanks!

spatialdata/_core/spatialdata.py
ivirshup: I think if AnnData gets an `__eq__` method, it's probably not going to return a single boolean. It'd probably be broadcasted over all the elements.

For something that returns a bool, it'd be a separate function. I think it's likely we'll have this one first since we've already got `assert_equal` for testing. It'll probably be slow and kept in a testing module.

LucaMarconato: Ok thanks for the comments, I'll turn it into a private method.

kevinyamauchi: Sounds good. Thanks everyone!

spatialdata/_core/spatialdata.py
```diff
@@ -1,25 +1,242 @@
...
         self,
-        adata: AnnData,
-        regions: Mapping[str, Any] = MappingProxyType({}),
-        images: Mapping[str, Any] = MappingProxyType({}),
-        points: Optional[AnnData] = None,
+        images: Optional[Mapping[str, Any]] = MappingProxyType({}),
+        labels: Optional[Mapping[str, Any]] = MappingProxyType({}),
+        points: Optional[Mapping[str, AnnData]] = MappingProxyType({}),
+        polygons: Optional[Mapping[str, AnnData]] = MappingProxyType({}),
+        tables: Optional[Mapping[str, AnnData]] = MappingProxyType({}),
```
ivirshup, line 35: Why are these `Optional`? They have default values.

I think it simplifies the model to say they must be `Mapping`s. E.g. cannot be `None`, and are instead just empty dicts.

```suggestion
        images: Mapping[str, Any] = MappingProxyType({}),
        labels: Mapping[str, Any] = MappingProxyType({}),
        points: Mapping[str, AnnData] = MappingProxyType({}),
        polygons: Mapping[str, AnnData] = MappingProxyType({}),
        tables: Mapping[str, AnnData] = MappingProxyType({}),
```

ivirshup (changes requested):
# Tests

Needs more tests. I don't see any tests for IO, and there are a few things in IO which I think don't work. However, it's unclear to me whether this is true, since I don't see an example of how methods are meant to be called.

Really this is the main blocking thing for me.

# Complexity and Construction pattern

I agree with Kevin on this.

In particular, I strongly that the base `SpatialData` constructor only accepts core elements for now, and basically just validates that they are related to each other. It won't try to convert/ coerce other classes. I think we can make some convenience methods for creating `SpatialData` objects for the tests.

# Classes

## `BaseElement`

I don't the current structural `BaseElement` class makes sense to me.

What is it meant to encapsulate?

### `Image(BaseElement)`

This defines a `.table` attribute. But `Image` inherits from this and does not have a `.table` attribute.

### `Labels(BaseElement)`
Why would this have a `table` attached to it, if the `table` is in a separate mapping on the `SpatialData` object? Are they meant to be the same table



## xarray-dataclass

Personally, I have started to come around on the design of [`spatial-image`](https://github.com/spatial-image/spatial-image) once I understood that the dataclasses here aren't instantiated, but are used to validate and create xarray objects (via https://github.com/astropenguin/xarray-dataclasses/).


# Extra files

Is it really useful to have files like `spatialdata/temp/napari_viewer.py` in the main branch, or can we keep these out of this PR?

kevinyamauchi:
Thanks for the reviews, @ivirshup !

> Needs more tests. I don't see any tests for IO, and there are a few things in IO which I think don't work. However, it's unclear to me whether this is true, since I don't see an example of how methods are meant to be called.
>
> Really this is the main blocking thing for me.

I agree that we want solid test coverage for IO and this PR doesn't provide that. Personally, at this early stage, I am okay with making an issue and circling back on tests if this PR is blocking other progress. However, tests are going to basically be our only form of documentation in the near future, so it's probably helpful to get them in ASAP. Also, backfilling tests is pretty painful...

I will defer to the group here, as I am not sure what you all are used to.


### Classes

> I don't the current structural BaseElement class makes sense to me.

I think you raise good points here. I think I had initially suggested the BaseElement because I thought the elements would share more functionality, but as seen here, that's perhaps not the case. Your point about `Image` not having a table also goes towards not having a base class. I would be okay with not having a base class and then refactoring later if we find that we are duplicating code.

> Labels(BaseElement)
>
> Why would this have a table attached to it, if the table is in a separate mapping on the SpatialData object? Are they meant to be the same table

Good point. I missed this.

My understanding is that the Elements would have their annotations stored together with the data. In this case, the table annotates the labeled regions. I don't think `SpatialData` should accept bare tables. What was the use case you had imagined for passing tables directly to `SpatialData`, @giovp and @LucaMarconato ?

ivirshup:
@kevinyamauchi 

> I agree that we want solid test coverage for IO and this PR doesn't provide that. Personally, at this early stage, I am okay with making an issue and circling back on tests if this PR is blocking other progress.

I'd be okay with lighter than usual coverage, but I think zero coverage isn't workable. I think it's really important for the implementer to put some tests down, so we have some expectations for API usage and what currently works.

> My understanding is that the Elements would have their annotations stored together with the data. In this case, the table annotates the labeled regions. I don't think SpatialData should accept bare tables. What was the use case you had imagined for passing tables directly to SpatialData, giovp and LucaMarconato ?

I thought the main case here would be one table referring to multiple Regions? E.g. either multiscale regions (cell and tissue level) or multiple slides.

I would think that none of the Region elements should have an annotation table attached, but tables in a `SpatialData` must refer to regions that exist within it. So you check that the `"region"`, `"region_key"`, etc. make sense at construction time.

spatialdata/_core/transform.py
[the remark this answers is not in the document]
giovp: removed

spatialdata/_constants/_enum.py
giovp: agree, if you don't mind I'll keep it as is and leave it as a todo for later

spatialdata/_core/spatialdata.py
giovp: I'll remove it here.

spatialdata/_io/format.py
```diff
@@ -0,0 +1,93 @@
...
+        ----------
+        tables
+            Tables.
+
+        Returns
+        -------
+            Nothing.
+        """
+        if tables is None:
+            return
```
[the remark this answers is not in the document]
giovp: it should, good catch

giovp:
hi all, thanks a lot for the feedback, I addressed most of your comments, most important point:
- I removed adding the table to `Elements` and only allow to pass single anndata to table (after discussion with @ivirshup )
- I kep the `BaseElement` class for now, as it might turn out handy, but it'd be easy to remove
- I added one test for readwrite

> My understanding is that the Elements would have their annotations stored together with the data. In this case, the table annotates the labeled regions. I don't think SpatialData should accept bare tables. What was the use case you had imagined for passing tables directly to SpatialData, @giovp and @LucaMarconato ?

@kevinyamauchi I think I agree with the answer from @ivirshup , pasted below:

> I thought the main case here would be one table referring to multiple Regions? E.g. either multiscale regions (cell and tissue level) or multiple slides.

> I would think that none of the Region elements should have an annotation table attached, but tables in a SpatialData must refer to regions that exist within it. So you check that the "region", "region_key", etc. make sense at construction time.

Although I also think that we might want to have tables associated to `Elements` in the case of `SpatialDataContainer`, and then to add a method to it like `to_spatialdata()` which would generate the current `SpatialData` which has ` single table in it, and for which some of the slicing operations we discussed would be implemented (and it would have more constraints than the container).
For this PR, I'd suggest to keep it like this, but I think that this is a major design choice that we should discuss this Thu and potentially carry out `SpatialDataContainer` dev in parallel, where the tight integration between `Labels` and `Tables` is present.

> Personally, I have started to come around on the design of [spatial-image](https://github.com/spatial-image/spatial-image) once I understood that the dataclasses here aren't instantiated, but are used to validate and create xarray objects (via https://github.com/astropenguin/xarray-dataclasses/).

I agree, also looking more into it, and playing around with ome_zarr readers, we prob should support multiscale representation from the start.

I'll open issues and once CI pass I'll merge this, as I believe all the major concerns are addressed (some in full, other in parts).

giovp:
closes #11

ivirshup:
@giovp, could you point me at the IO test?

giovp:
@ivirshup apologies, I completely forgot to push the new folder, it is now in #21

PR #21, opened 2022-08-17
add io test

giovp:


ivirshup (approved):
I'd like to merge.

Question though: Why do the test directories have names like `_io`? Idk that we need to specify private v public tests.

giovp:
I believe @LucaMarconato merged this in #22  already, can we close it?

thanks for feedback, good point we can remove private/public

ivirshup:
Can we merge it instead? I would much rather have the test on main.

LucaMarconato:
Yes we can merge. I added an extra step to the roundtrip (writing back to disk and checking the written output bytewise) and this triggers a nasty bug. Gonna open a issue later, but we meanwhile we can merge this.

PR #22, opened 2022-08-17
transformations (wip)

LucaMarconato:
- io for transform
- fixed bugs from previous pull request (appeared when running on the merfish dataset from spatialdata-sandbox)


kevinyamauchi:
@LucaMarconato , does this need review or are you still working on this? No pressure - just checking in to see if you need anything from me.

LucaMarconato:
Hi, there is still quite some work to be done. I am going through all the examples in `spatialdata-sandbox`, converting to OME-NGFF, reading them into a `SpatialData` object and passing them to the new `napari-spatialdata` plugin. You can see the current status of the datasets that work by running `examples/view_spatialdata.py` in `napari-spatialdata`.

LucaMarconato:
@giovp there are two flake8 issues in `spatialdata._constants._enum.py`. I am not familiar with that code, can you check it please?

LucaMarconato:
@giovp this is now ready for review. It implements the new transform specs, tests and io. I have tested it also with the `merfish`, `visium` and `mibitof` datasets from `spatialdata-sandbox`, which is now updated to use the new specs and I was also updated `napari-spatialdata` (`spatialdata` brunch) to read the transformations.

What's missing is the support for coordinate spaces and especially for axes annotations. For this reason the `merfish` dataset can be viewed in napari, but the `visium` one at the moment no. I will work on integrating your coordinate spaces in a new branch.

spatialdata/_core/elements.py
```diff
@@ -48,26 +51,46 @@ def shape(self) -> Tuple[int, ...]:
...
+        # at the moment we don't use the compressor because of the bug described here (it makes some tests the
+        # test_readwrite_roundtrip fail) https://github.com/ome/ome-zarr-py/issues/219
         write_image(
             image=self.data.data,
             group=group,
-            axes=["c", "y", "x"],  # TODO: infer before.
+            axes=self.axes,
             scaler=scaler,
+            coordinate_transformations=[[coordinate_transformations]],
+            storage_options={"compressor": None},
```
giovp, line 86: shouldn't this be exposed, even if just by kwargs?

spatialdata/_core/elements.py
```diff
@@ -48,26 +51,46 @@ def shape(self) -> Tuple[int, ...]:
...
         )
 
+    def _infer_axes(self, shape: Tuple[int]) -> List[str]:
+        # TODO: improve (this information can be already present in the data, as for xarrays, and the constructor
+        # should have an argument so that the user can modify this
+        return ["c", "y", "x"][3 - len(shape) :]
```
giovp, line 92: I would add the argument now and also would somehow document the c,y,x or better make it a constant (In _constants)

spatialdata/_core/elements.py
```diff
@@ -84,23 +107,38 @@ def shape(self) -> Tuple[int, ...]:
...
 
 class Labels(BaseElement):
-    def __init__(self, labels: DataArray, transform: Transform) -> None:
-        self.data: DataArray = labels
+    def __init__(self, labels: DataArray, transform: BaseTransformation) -> None:
+        if isinstance(labels, DataArray):
+            self.data = labels
+        elif isinstance(labels, np.ndarray):
+            self.data = DataArray(labels)
+        elif isinstance(labels, dask.array.core.Array):
```
giovp, line 115: nitpick, would do `from dask.array.core import Array as DaskArray`

spatialdata/_core/elements.py
```diff
@@ -84,23 +107,38 @@ def shape(self) -> Tuple[int, ...]:
 
 
 class Labels(BaseElement):
-    def __init__(self, labels: DataArray, transform: Transform) -> None:
-        self.data: DataArray = labels
+    def __init__(self, labels: DataArray, transform: BaseTransformation) -> None:
```
giovp, line 110: input type of this is the kind of thing we need to validate.

spatialdata/_core/elements.py
```diff
@@ -84,23 +107,38 @@ def shape(self) -> Tuple[int, ...]:
...
+        # at the moment we don't use the compressor because of the bug described here (it makes some tests the
+        # test_readwrite_roundtrip fail) https://github.com/ome/ome-zarr-py/issues/219
         write_labels(
             labels=self.data.data,
             group=group,
             name=name,
             axes=["y", "x"],  # TODO: infer before.
             scaler=scaler,
+            storage_options={"compressor": None},
+            coordinate_transformations=[[coordinate_transformations]],
```
giovp, line 141: same here se compressor and also described above.

giovp: but also, maybe this can be handled in separate IO PR

spatialdata/_core/elements.py
```diff
@@ -149,26 +192,68 @@ def from_path(
...
+                vertices = np.squeeze(vertices, 0)
+                assert len(vertices.shape) == 2
+                coordinates.append(vertices)
+            else:
+                print(f'ignoring "{region["type"]}" from geojson')
+
+        string_coordinates = [Polygons.tensor_to_string(c) for c in coordinates]
+        a = AnnData(shape=(len(names), 0), obs=pd.DataFrame({"name": names, "spatial": string_coordinates}))
+        return a
+
```
giovp, line 247: approving this but ofc will need rewrite

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+    "Coordinates",
+    "VectorField",
+    "InverseOf",
+    "Bijection",
+    "ByDimension",
+    "compose_transformations",
+]
+
+
+class BaseTransformation(ABC):
```
giovp, line 32: nitpick, can we call it `BaseTransform` to be consistent? also shorted

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+    @abstractmethod
+    def inverse(self) -> BaseTransformation:
+        pass
+
+
+def compose_transformations(transformation0: BaseTransformation, transformation1: BaseTransformation) -> Sequence:
+    return Sequence([transformation0, transformation1])
+
+
+def get_transformation_from_json(s: str) -> BaseTransformation:
```
giovp, line 61: nitpick, would make these `def _compose...` and would move to end of file.

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+        cls = Bijection
+    elif type == "byDimension":
+        cls = ByDimension  # type: ignore
+    else:
+        raise ValueError(f"Unknown transformation type: {type}")
+    del kw["type"]
+    if "input" in kw:
+        del kw["input"]
+    if "output" in kw:
+        del kw["output"]
```
giovp, line 104: what is this? why `del` ?

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+    def __init__(self) -> None:
+        self._ndim = None
+
+    @property
+    def ndim(self) -> Optional[int]:
+        return self._ndim
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "type": "identity",
```
giovp, line 118: would also make these constants or somehow put them in schema. Would basically avoid any type of hard coding

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+    elif type == "vectorField":
+        cls = VectorField  # type: ignore
+    elif type == "inverseOf":
+        cls = InverseOf
+    elif type == "bijection":
+        cls = Bijection
+    elif type == "byDimension":
+        cls = ByDimension  # type: ignore
+    else:
+        raise ValueError(f"Unknown transformation type: {type}")
```
giovp, line 99: would make all this as constants, also not sure about camelCase but I guess it's how they are defined in the specs?

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+        return points + self.translation
+
+    def inverse(self) -> BaseTransformation:
+        return Translation(translation=-self.translation)
+
+    def to_affine(self) -> Affine:
+        assert self.ndim == 2
+        m: ArrayLike = np.vstack(
+            (np.hstack((np.array([[1.0, 0.0], [0.0, 1.0]]), self.translation.reshape(2, 1))), np.array([0.0, 0.0, 1.0]))
+        )
```
giovp, line 190: I'm not sure what this is, think it should be documented somewhere?

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+        assert type(scale) == np.ndarray
+        assert len(scale) == ndim
+        self.scale = scale
+        self._ndim = ndim
+
+    @property
+    def ndim(self) -> Optional[int]:
+        return self._ndim
+
+    def to_dict(self) -> Dict[str, Any]:
```
giovp, line 222: general comment borderline nitpick, but I would explicilty type dictionaries, not with `Any`. This one especially should just be string or sequences of floats

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+class Affine(BaseTransformation):
+    def __init__(
+        self,
+        affine: Optional[Union[ArrayLike, List[Any]]] = None,
+    ) -> None:
+        """
+        class for storing scale transformations.
+        """
+        if affine is None:
+            affine = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]], dtype=float)
```
giovp, line 251: why not using `np.diag` ?

giovp: or better `np.eye` or `np.identity` ?

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+    def inverse(self) -> BaseTransformation:
+        # naive check for numerical stability
+        a = self.affine[0, 0]
+        b = self.affine[0, 1]
+        m = self.affine[0, 2]
+        c = self.affine[1, 0]
+        d = self.affine[1, 1]
+        n = self.affine[1, 2]
+        det = a * d - b * c
+        closed_form = np.array([[d, -c, 0], [-b, a, 0], [b * n - d * m, c * m - a * n, det]])
```
giovp, line 287: remember you explaining this, I wonder if it makes sense to have it as separate function called by the method and documented. Also if there is no already something like this in scipy?

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+        self._ndim = 2
+
+    @property
+    def ndim(self) -> Optional[int]:
+        return self._ndim
+
+    def to_dict(self) -> Dict[str, Any]:
+        return {
+            "type": "rotation",
+            "rotation": self.rotation.ravel().tolist(),
```
giovp, line 322: sometimes it is `.tolist()` some other it is `.ravel().tolist()` would converge in one.

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+            assert rotation.shape == (2, 2)
+
+        self.rotation = rotation
+        self._ndim = 2
+
+    @property
+    def ndim(self) -> Optional[int]:
+        return self._ndim
+
+    def to_dict(self) -> Dict[str, Any]:
```
giovp, line 319: I'm getting more convinced that it makes sense to type dictionaries.

spatialdata/_core/transform.py
```diff
@@ -1,86 +1,526 @@
...
+        return {
+            "type": "bijection",
+            "forward": self.forward.to_dict(),
+            "inverse": self._inverse.to_dict(),
+        }
```
giovp, line 466: I feel we might want to put transform returns somewhere else (maybe a dataclass) ? and only populate them in the class method call.

spatialdata/temp/bug_dask_zarr.py
```diff
@@ -0,0 +1,58 @@
+##
```
giovp, line 1: what is this?

spatialdata/temp/bug_dask_zarr_report.py
```diff
@@ -0,0 +1,94 @@
+##
```
giovp, line 1: whta is this?

spatialdata/utils.py
```diff
@@ -19,15 +26,106 @@ def phase3(self) -> None:
...
+        #
+        # loc = ZarrLocation(_root_dir2)
+        # reader = Reader(loc)()
+        # nodes = list(reader)
+        # assert len(nodes) == 1
+        # node = nodes[0]
+        # im_read2 = node.load(Multiscales).array(resolution="1", version=CurrentFormat().version)
+        # print(im_read[:5, :5, 0].compute())
+        # print(im_read2[:5, :5, 0].compute())
+        ##
```
giovp, line 65: I'm not sure hwat to do with this text, I would just remove otherwise not comment

giovp (changes requested):
really great work, impressive to see all the transforms already being laid out. Left some comments but to summarize:
- I think we want to have a way to store the transforms types and variables somewhere else, and in the class only materialize them. They could be as a schema or as a dataclass?> but basically I would avoid instantiating dictionaries in method call, and rather separate the transform implementations from the transforms specs
- there are a lot of code changes, and there is a lot of IO. I will try to get IO done by end of the week and I will probably branch out from this PR to incorporate your current changes.
- beside my comments, I think that this PR should:
  1. be re-reviewed after the proposal discussion
  2. be reviewed by someone else as well as there are quite some code changes.

spatialdata/_core/elements.py
LucaMarconato: Yes, and also `chunks` needs to be exposed. It's crucial to let the user choose a good chunk size (we can also make a good guess) to have good performance also for massive images.

LucaMarconato:
I am working on this pr. Could you please give more details re this:
> I think we want to have a way to store the transforms types and variables somewhere else, and in the class only materialize them. They could be as a schema or as a dataclass?> but basically I would avoid instantiating dictionaries in method call, and rather separate the transform implementations from the transforms specs

My idea is to actually keep the logic of the specs in the classes, since it is not just about data storage but also what each type of transformation can do, like an `Identity` can't have a `ndim` method telling the dimensionality since the specs don't specify information about that for identity transformations, but a `ndim` method is available for `Rotation` since the specs require the full matrix to be specified.

But what I would do is that I would separate how the specs dictate how to store the data (i.e. as json/dict) and the variables that are inside the classes. So for example `Scale` would now access in the constructor just a vector of scaling factors, not the corresponding dict that the NGFF specs describes.

spatialdata/_core/elements.py
LucaMarconato: ✔️ exposed `compressor` and `chunks`

spatialdata/_core/elements.py
LucaMarconato: skipping this from this pr since it is implemented in another pr that I'll open after this is merged

spatialdata/_core/elements.py
LucaMarconato: 👍

spatialdata/_core/elements.py
LucaMarconato: done

spatialdata/_core/elements.py
LucaMarconato: Agree, it needs to be rewritten. Also, I moved it spatialdata-io (keeping it here in this pr).

spatialdata/_core/transform.py
LucaMarconato: I actually removed since it was redundant: if private, so not exposed to the user, then `Compose` does the same job (and more).

spatialdata/_core/transform.py
LucaMarconato: from the transform specs, there are cases where `input` and `output` (the input and output coordinate systems of the transformation) are specified in the dict describing the transformation. But in one edge case they are not specified: when a transformation is used within a sequence. In this file I am trying to separate the concept of transformation and of input/output coordinate system (as it is done in the specs), that's why if input and output are present we remove them, since the constructors of the various classes don't take them.

LucaMarconato: We can safely use `del` since `kw` is a copy of the input dict.

spatialdata/_core/transform.py
LucaMarconato: copilot typed it for me 😁 for the ndim case I'll use `np.eye`

spatialdata/_core/transform.py
LucaMarconato: Good point. For the ndim case we'll have to use something like `numpy.linalg.inv` anyway. I'll modify the code.

spatialdata/_core/transform.py
LucaMarconato: This is due to the fact that the specs sometimes store the matrix as a list and not a list of lists, and without the last row of the affine matrix, which is always [0, 0, 0, ..., 0, 1]. I will replace every class to accept as parameter to `__init__`, and store, a real affine matrix, and leave the job of removing the last row, and eventually flattening the list, to another separate part of the code.

spatialdata/_core/transform.py
LucaMarconato: Agree, I will do it.

spatialdata/_core/transform.py
LucaMarconato: Agree, I'll do it.

spatialdata/_core/transform.py
LucaMarconato: Yes, I agree we should separate the two things.

spatialdata/temp/bug_dask_zarr.py
LucaMarconato: It's to use [*Python cell mode* in PyCharm](https://plugins.jetbrains.com/plugin/7858-pycharm-cell-mode). Basically it allows to send code blocks to an open console/debug console. It gives the advantages of Jupyter-like cells without having to abandon `.py` files. I sometimes use it while debugging, can be deleted.

spatialdata/utils.py
LucaMarconato: removed

LucaMarconato:
Merge postponed.
Before merging (an updated version of) this pr, the following will be isolated in a separate pr, improved, and merged.
- [x] coordinate systems
- [ ] coordinate transformations
- [ ] io branch from @giovp 

## About this pr.
This pull request implements a first working version of coordinate systems, transformations, as long as all that spatialdata needs to perform basic io operatations with its elements.

spatialdata/_core/transform.py
LucaMarconato: Follow up on this, I tried doing this by introducing the (ugly)
```
Transformation_t = Dict[str, Union[str, List[int], List[str], List[Dict[str, Any]]]]
```
and replacing all the typing annotation with `Transformation_t`, but I have just realized that this is suboptimal since there is still a `Any` inside, and also this was giving a ton of weird mypy errors. The problem here is that `Sequence` and `MapAxis` transformations can contain infinitely nested transformations. Not sure if there is a way around this to support for these recursive dictionary types.

Meanwhile I have kept `Transformation_t` around the code, but I have redefined it as `Dict[str, Any]`.

LucaMarconato: I would maybe keep it like this and eventually if we want to give a better typing I would open a relative issue for the future.

spatialdata/_core/transform.py
LucaMarconato: I didn't use dataclasses but I went toward what you said by doing the following.
The conversion between how things are stored on disc (json representation of python dicts with native types) and how are stored in memory (transformation and coordinate system classes) is now done only in the functions `from_dict()` and `to_dict()`. Before it was messy because the constructor of each transformation class could take as input both the dict representation or explicit constructor parameters (example: before a `Sequence` could be defined by a list of `BaseTransformation` or by a list of `Dict`, now only by a list of `Transformation`; another example: before a `Affine` could take as input either the matrix representation of the affine transformation, either the list of list of float specified by the NGFF specs, which is basically the matrix without the last row. Now only the first matrix representation is allowed).

I don't like that `from_dict()` is in `BaseTransformation` and has a long list of `if else`, while `to_dict()` is specified in each class, we could make this uniform.

LucaMarconato:
@giovp ready for review! Please see the two comments regarding your previous review.

Additional comments on the pr.
1. When constructing a transformation one can decide to omit, or to specify, the input and/or output coordinate systems. 
1.1. By default they are None, and if the user don't specify them, there will be an error when trying to apply the transformation to points.
1.2. If they are specified, then a validation step will be performed when trying to apply the transformation to points, whenever this is possible. For instance an error will be raised if one tries to apply an Identity/Translate/Scale/Rotation transformation between different input and output coordinate systems. Instead when applying a Affine transformation, only the dimensionality of the axes is checked, but the transformation will be considered valid both if the input->output are xy->xy and xy->yx. There is no way to know which of the two is correct, it is responsibility of the user.
1.3. The coordinate system can be a string. This allows to specify a coordinate system by name. It will be responsibility of SpatialData to fill it with the appropriate `CoordinateSystem` object, otherwise an error will be raised when trying to apply the transformation to some points.
2. The class `Sequence` is complex. The reason is that by design the transformations that compose a `Sequence` object can be declared without specifying the input/output coordinate systems. These systems can be sometimes inferred (like if we apply a scale we know that input and output axes must coincide). Other times this is not possible. Instead of just applying the sequence of transformations without checking if the inferred coordinate systems make sense, I have written a system to infer them and check them when possible, by populating the `.input_coordinate_system` and `.output_coordinate_system` realtime when applying a `Sequence` transformations. These changes are then reverted back when the transformation is applied. In this way we leave the object untoched but we are able to catch bugs. I will link a discussion with John regarding this topic and add it to the comments.

ivirshup:
### `from_dict`

Point of feedback regarding `from_dict`, and an alternative to the big `if ... else` block:

Instead of doing one big block, I would have something more like:

```python
TRANSFORMATION_REGISTRY = {
    "Rotation": Rotation.from_dict,
    ...
}

def transformation_from_dict(arg: dict) -> BaseTransformation:
    return TRANSFORMATION_REGISTRY[arg["type"]](arg)
```

This lets us keep code factored out, and allows registering more transformations as they're created.

LucaMarconato:
The branch has already been merged to `feature/transform_and_schema`, which has been merged to `main`. The code review has been implemented, so no need to merge and now closing this pr.

PR #40, opened 2022-10-22
schema and dataclasses (wip)

giovp:


giovp:
polygons IO only works with `pip install git+https://github.com/shapely/shapely.git@refs/pull/1559/head --no-binary shapely` (branch where `to/from_ragged_array` is added).

giovp:
open questiopns:
- [ ] can we have a SpatialData with only one table (no labels, regions etc.) ?
- [ ] if `region: str` what is the table annotating?
- [ ] can we/shall we check for consistency of `instance_key` ?

spatialdata/_core/models.py
```diff
@@ -0,0 +1,421 @@
...
+def _(
+    data: np.ndarray,  # type: ignore[type-arg]
+    shape_type: Literal["Circle", "Square"],
+    shape_size: np.float_,
+    transform: Optional[Any] = None,
+    **kwargs: Any,
+) -> AnnData:
+
+    # TODO: AnnData(obsm={"spatial": data}) doesn't work, shall we change?
+    adata = AnnData(np.empty(shape=data.shape), obsm={"spatial": data})
```
giovp, line 319: @ivirshup this `AnnData(obsm={"spatial": data})` doesn't work. Quite important if we want to save shapes/points in anndata with coordinates in obsm (and X potentially `None` ). What should we do?

spatialdata/_io/write.py
```diff
@@ -66,133 +70,231 @@ def _write_metadata(
...
-        name=name,
-        scaler=scaler,
-        chunks=chunks,
+    sub_group = group.require_group("shapes")
+    coordinate_transformations = [[shapes.uns.get("transform").to_dict()]]
+    attr = shapes.uns.get("spatialdata_attrs")
+    # TODO: do not save copy but pop transform?
+    shapes_copy = shapes.copy()
+    shapes_copy.uns.pop("transform")
+    write_adata(sub_group, name, shapes_copy)
```
giovp, line 227: basically if `adata.uns["transform"] = spatialdata.Transform(...)` it can't ofc be serialized so it needs to be popped. However, that would change the in-memory representation. Can only copy or someone has better idea?

giovp:
zarr group hierarchy of elements:

## Image
```
.
├── image0
│   └── 0
│       └── 0
│           └── 0
├── image01
│   └── 0
│       └── 0
│           └── 0
└── image012
    └── 0
        └── 0
            └── 0
```

## Labels
```
.
├── label0
│   └── labels
│       └── label0
│           └── 0
│               └── 0
├── label01
│   └── labels
│       └── label01
│           └── 0
│               └── 0
└── label012
    └── labels
        └── label012
            └── 0
                └── 0
```

## Shapes
```
.
├── shapes0
│   └── shapes
│       └── shapes0
│           ├── X
│           │   └── 0
...
└── shapes01
    └── shapes
        └── shapes01
            ├── X
            │   └── 0
...
```

## Polygons
```
.
├── polygons0
│   └── polygons
│       └── polygons0
│           ├── coords
│           │   └── 0
│           └── offsets
│               └── 0
├── polygons01
│   └── polygons
│       └── polygons01
│           ├── coords
│           │   └── 0
│           └── offsets
│               └── 0
└── polygons012
    └── polygons
        └── polygons012
            ├── coords
            │   └── 0
            └── offsets
                └── 0
```

## Table
```
.
└── table
    └── table
        ├── X
        │   ├── data
        │   ├── indices
        │   └── indptr
        ├── layers
        ├── obs
```

and the `SpatialData` specific attributes are in `path` in `.zattr`, e.g. for polygons
```
"path": {
                        "geos": {
                            "geometry_name": "POLYGON",
                            "geometry_type": 3
                        }
                    }
```

let's use this for discussion, think we can improve a bit.

spatialdata/_core/models.py
```diff
@@ -0,0 +1,421 @@
...
+    array_type=ArrayTypeSchema(DaskArray),
+    attrs={"transform": Transform_s},
+)
+Labels3D_s = DataArraySchema(
+    dims=DimsSchema((Z, Y, X)),
+    array_type=ArrayTypeSchema(DaskArray),
+    attrs={"transform": Transform_s},
+)
+Image3D_s = DataArraySchema(
+    dims=DimsSchema((Z, C, Y, X)),
```
LucaMarconato, line 58: Let's use czyx to reflect the NGFF 0.4 specs. Will be nice to support arbitrary order of axes but this maybe has to be done on the `SpatialImage` side.

giovp: done

PR #44, opened 2022-10-28
Add initial draft of the design doc

kevinyamauchi:
This PR adds the initial draft of a design document describing the goals and strategy for designing the SpatialData library.

I propose that we review to make sure this captures our current thoughts as of today, but try and merge quickly (before we change our minds 😉 ). We can continue to revisit and modify as we do the implementation.

### DIFF
--- a/docs/design_doc.md
+++ b/docs/design_doc.md
@@ -107,7 +107,7 @@ The `SpatialData` object contains a set of Elements to be used for analysis. Ele
 
 **`NGFFStore`**
 
-The `NGFFStore` is an object representing the on-disk layout of a dataset. The `NGFFStore` parses the files to determine what data are available to be loaded. Initial implementations will target a single Zarr file on disk, but future implementations may support reading from a collection of files.
+The `NGFFStore` is an object representing the on-disk layout of a dataset. The `NGFFStore` parses the files to determine what data are available to be loaded. Initial implementations will target a single Zarr file on disk, but future implementations may support reading from a collection of files. A `SpatialData` object can be instantiated from a `NGFFStore`.
 
 ### Elements
 
--- a/docs/design_doc.md
+++ b/docs/design_doc.md
@@ -1,0 +1,346 @@
+## Abstract
+This documents defines the specifications of SpatialData: a FAIR format for multi-modal spatial omics data. It also describes the initial implementation plan. This is meant to be a living document that can be updated as the project evolves.
+
+## Motivation and Scope
+Recent advances in molecular profiling technologies allow to measure abundance of RNA and proteins in tissue, at high throughput, multiplexing and resolution. The variety of experimental techniques poses unique challenges in data handling and processing, in particular around data types and size. *SpatialData* aims at implementing a performant in-memory representation in Python and an on-disk representation based on the Zarr data format and following the OME-NGFF specifications. By maximing interoperability, performant implementations and efficient (cloud-based) IO, *SpatialData* aims at laying the foundations for new methods and pipelines for the analysis of spatial omics data.
+
+## Goals
+
+The goals define *what* SpatialData will be able to do (as opposed to *how*). Goals can have the following priority levels:
+
+- P0: highest priority, required for successful implementation (i.e., must have)
+- P1: high priority, but not required (i.e., nice to have)
+- P2: nice to have, but not a priority
+
+**1. Load data from modern spatial multiomics experiments**
+
+- P0. Data can be loaded from the OME-NGFF and saved to OME-NGFF.
+    - [x] multiscale images and labels
+    - [x] point clouds
+    - [x] polygon-shaped regions of interest
+    - [x] circle-shaped regions of interest
+    - [x] tables
+    - [x] graphs -> see how it's done in Napari
+- P0. Data can be loaded lazily.
+> [name=Luca] I think that lazy loading for points data has high priority, later we should discuss how to tackle the problem.
+> [name=Isaac] The issue is that fully loading a large FISH experiment into memory takes some time. It isn't "terrible", but is quite a while for opening a viewer. We should delay reading the points until the viewer wants to see them for sure. We can also speed up full reading, but it will always take some time.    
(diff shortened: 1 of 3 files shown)


kevinyamauchi:
I think we will likely want to add a schematic showing the relationships between the different pieces, but I think we can do that in a follow up PR.

Also, I wasn't sure where to put it. Happy to move it elsewhere!

kevinyamauchi:
@LucaMarconato , can you add the rest of the pseudocode we worked out to the bottom of the document?

kevinyamauchi:
Hey everyone. I think this is good to go. I haven't heard any objections and @LucaMarconato has reviewed, so I will merge tomorrow unless I hear otherwise.

docs/design_doc.md
```diff
@@ -0,0 +1,443 @@
...
+
+**`SpatialData`**
+
+The `SpatialData` object contains a set of Elements to be used for analysis. Elements contained within a `SpatialData` object must be able to share a single Region Table. Future work may extend `SpatialData` to allow multiple tables (see discussion [here](https://github.com/scverse/spatialdata/issues/43)). All Elements within a `SpatialData` object can be queried, selected, and saved via the `SpatialData` objects.
+
+**`NGFFStore`**
+
+_Note: previously we used the term `SpatialDataContainer`; the new naming is `NGFFStore`._
```
giovp, line 110: I think we can remove this reference? since there isn't a "before" of this document imho.

docs/design_doc.md
```diff
@@ -0,0 +1,443 @@
...
+
+The `NGFFStore` is an object representing the on-disk layout of a dataset. The `NGFFStore` parses the files to determine what data are available to be loaded. Initial implementations will target a single Zarr file on disk, but future implementations may support reading from a collection of files.
```
giovp, line 112:
```suggestion
The `NGFFStore` is an object representing the on-disk layout of a dataset. The `NGFFStore` parses the files to determine what data are available to be loaded. Initial implementations will target a single Zarr file on disk, but future implementations may support reading from a collection of files. A `SpatialData` object can be instantiated from a `NGFFStore`.
```

docs/design_doc.md
```diff
@@ -0,0 +1,443 @@
...
+By decomposing the data model into building blocks (i.e. Elements) we support operations that do not otherwise fit this model. For instance, we may have data where regions have not been defined yet, or are just dealing with images and points.
+
+#### Assumptions
+
+_SpatialData_ closely follows the OME-NGFF specifications and therefore much of its assumptions are inherited from it. Extra assumptions will be discussed with the OME-NGFF community and adapted to the community-agreed design. The key assumptions are the following:
+
+-   `Images`, `Labels`, `Points`, and `Shapes` MUST have one or more _coordinate systems_ and _coordinate transforms_.
+-   `Tables` CAN NOT have a _coordinate system_ or _coordinate transforms_.
+-   `Labels` and `Shapes` are both instances of `Regions`.
+-   `Regions` are `Elements` and they MAY be annotated by `Tables`.
```
giovp, line 136: think important to clarify here re `Points`. If we unify them as `Circles` and `Points` than we should add that `Points` are instanced of `Regions`.
Also below, since we have only one `Table` then `Points` CANNOT be annotated by `Tables.`

docs/design_doc.md
```diff
@@ -0,0 +1,443 @@
...
+-   Imaging or feature channels.
+-   Z-stacks.
+
+We will support also time-point axes in the future. Furthermore, due to NGFF specs v0.5, such axes will not have name constraints (although they do for first iteration due to NGFF specs v0.4).
+
+The image object itself builds on prior art in image analysis, in particular the [xarray library](https://docs.xarray.dev/en/stable/).
+
+Images have labelled dimensions, and coordinate transformations. These transformations are used to map between pixel space and physical space, and between different physical spaces.
+
+For computational efficiency, images can have a pyramidal or multiscale format. This is implemented as an [xarray datatree](https://github.com/xarray-contrib/datatree).
```
giovp, line 156: think we could add a reference to spatial-image but maybe only later on?

docs/design_doc.md
giovp: ok I now read that shapes will be also circles (I like that) hence the first part of the comment can be discarded. The part about points not being annotated by tables though remains.

docs/design_doc.md
```diff
@@ -0,0 +1,443 @@
...
+
+The label object itself builds on prior art in image analysis, in particular the [xarray library](https://docs.xarray.dev/en/stable/).
+
+For computational efficiency, labels can have a pyramidal or multiscale format. This is implemented as an [xarray datatree](https://github.com/xarray-contrib/datatree).
+
+##### Polygons
+
+A set of (multi-)polygons associated with a set of observations. Each set of polygons is associated with a coordinate system. Polygons can be used to represent a variety of regions of interests, such as clinical annotations and user-defined regions of interest.
+
+The Polygon object is implemented as a geopandas dataframe with [multi-polygons series](https://geopandas.org/en/stable/docs/user_guide/data_structures.html).
```
giovp, line 191: Furtehr to be added:
`The coordinate system and transforms are stored in ``geopandas.DataFrame.attrs``

giovp: same as above for labels and images, we store it in xarray.DataArray.attrs

giovp (approved):
few comments

docs/design_doc.md
kevinyamauchi: Good point. Done.

docs/design_doc.md
kevinyamauchi: Good catch. Updated to say that Points MAY NOT be annotated by tables.

docs/design_doc.md
kevinyamauchi: Done!

docs/design_doc.md
kevinyamauchi: I added a reference here. We can revise once we have more clarity on if/how we will depend on `spatial-image`

kevinyamauchi:
Thanks, @giovp ! I've addressed your comments.

PR #52, opened 2022-11-29
transform and models

giovp:
This PR implements transforms (#22  and model validation and parsing (#40) for SpatialData.


pyproject.toml
```diff
@@ -22,7 +22,20 @@ version = "0.0.1.dev0"
...
+    "xarray",
+    "zarr",
+    "ome_zarr",
+    "spatial_image",
+    "multiscale_spatial_image",
+    "xarray-schema",
+    "geopandas",
+    "pandera",
+    "pygeos",
+    "shapely @ git+https://github.com/shapely/shapely.git@main --no-binary shapely",
```
kevinyamauchi, line 37: @giovp, I think this is causing installation on CI to fail. Does this this work locally with 'pip install .'?

pyproject.toml
```diff
@@ -22,7 +22,20 @@ version = "0.0.1.dev0"
...
+    "xarray",
+    "zarr",
+    "ome_zarr",
+    "spatial_image",
+    "multiscale_spatial_image",
+    "xarray-schema",
+    "geopandas",
+    "pandera",
+    "pygeos",
+    "shapely @ git+https://github.com/shapely/shapely.git@main",
```
kevinyamauchi, line 37: I think this should work

```suggestion
    "geopandas"
    "pandera"
    "shapely==2.0rc1",
```

kevinyamauchi:
CI is running! Looks like updating the `shapely` and `GEOS` dependencies did the trick. 🎉

LucaMarconato:
Two comments on the models @giovp (2 things are missing).
1. in `PointModel` there is the need for another array being passed to the parser, in fact since points can't be annotated by tables, we need to include the annotation when constructing them, which is one extra column specifying the cell type for each point.
2. In Shape model the `shape_size` is a float, while in some datasets the shape is point dependent.

LucaMarconato:
Looking at the design doc I see that only a single value is mentioned. I think we forgot to discuss single number vs array of radiuses. What shall we do? We keep a single number and we update later on or we implement an array of radii? If we go with the first route we could still visualize datasets like this (MERFISH), by rasterizing the "shapes" (a bit of a hack).
image

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+Image2D_s = Image2DModel()
+Image3D_s = Image3DModel()
+Polygon_s = PolygonModel
+Point_s = PointModel()
+Shape_s = ShapeModel()
+Table_s = TableModel()
+
+
+class SpatialData:
+    """Spatial data structure."""
```
kevinyamauchi, line 47: This is our main object, so I think we need a really complete docstring including the parameters for the `__init__` and the properties.

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+        labels: Mapping[str, Any] = MappingProxyType({}),
+        points: Mapping[str, Any] = MappingProxyType({}),
+        polygons: Mapping[str, Any] = MappingProxyType({}),
+        shapes: Mapping[str, Any] = MappingProxyType({}),
+        table: Optional[AnnData] = None,
+        **kwargs: Any,
+    ) -> None:
+
+        if images is not None:
+            self.images = {}
```
kevinyamauchi, line 68: I think this is a bit confusing because the default element property values and `__init__` parameters values are `MappingProxyType`s while then end up getting initialized as regular dictionaries. The `MappingProxyType` would be readonly while the dictionary is mutable.

What is the intended behavior? Should we make the elements properties without settings so that they can't be directly mutated?

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def _gen_repr(
+        self,
+    ) -> str:
+        def rreplace(s: str, old: str, new: str, occurrence: int) -> str:
+            li = s.rsplit(old, occurrence)
+            return new.join(li)
+
+        def h(s: str) -> str:
+            return s
```
kevinyamauchi, line 192: What is this function for?

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+                            descr += f"{h(attr + 'level1.1')}'{k}': {descr_class} " f"shape: {v.shape}"
+                        elif attr == "shapes":
+                            # assuming 2d
+                            descr += f"{h(attr + 'level1.1')}'{k}': {descr_class} " f"shape: {v.shape}"
+                        else:
+                            if isinstance(v, SpatialImage) or isinstance(v, MultiscaleSpatialImage):
+                                descr += f"{h(attr + 'level1.1')}'{k}': {descr_class}"
+                            else:
+                                descr += f"{h(attr + 'level1.1')}'{k}': {descr_class} {v.shape}"
+                        # descr = rreplace(descr, h("level1.0"), "    └── ", 1)
```
kevinyamauchi, line 224: can we remove this comment?

spatialdata/_core/models.py
```diff
@@ -0,0 +1,407 @@
...
+                )
+
+        elif isinstance(attr["region"], str):
+            attr["region_key"] = None
+            attr["instance_key"] = None
+
+        return data
+
+    @classmethod
+    def parse(
```
LucaMarconato, line 375: Some comments:
- `instance_key` is specified always, even when `region` is not a list
- no need to pass `region`, `region_key` and `instance_key` to the validator
I'll take care of modifying the above in the code that I will push.

spatialdata/_core/models.py
```diff
@@ -0,0 +1,407 @@
...
+class TableModel:
+    attrs_key = "spatialdata_attrs"
+
+    def validate(
+        self,
+        data: AnnData,
+        region: Optional[Union[str, Sequence[str]]] = None,
+        region_key: Optional[str] = None,
+        instance_key: Optional[str] = None,
+        **kwargs: Any,
```
LucaMarconato, line 352: Do we need kwargs? Same for the parser.

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+            li = s.rsplit(old, occurrence)
+            return new.join(li)
+
+        def h(s: str) -> str:
+            return s
+            # return hashlib.md5(repr(s).encode()).hexdigest()
+
+        ##
+        descr = "SpatialData object with:"
+        for attr in ["images", "labels", "points", "polygons", "shapes", "table"]:
```
kevinyamauchi, line 197: nitpick: this function would likely be cleaner if we had a separate function for generating reprs for each element. We could then generate the repr for each element type and then concatenate them. This would also allow as to move a lot of the details off of the `SpatialData` class into util functions. I think we should strive to make the `SpatialData` class as simple/clean as possible since this will be most people's main contact with the library.

In general, I think long if/else blocks like this are hard to follow and hard to test.

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+
+        for el in elems:
+            elem_group = root.create_group(name=el)
+            if self.images is not None and el in self.images.keys():
+                write_image(
+                    image=self.images[el],
+                    group=elem_group,
+                    name=el,
+                    storage_options={"compressor": None},
+                )
```
kevinyamauchi, line 137: Can we use the `to_zarr()` methods on each element instead of the `write_*` functions? That way we just need to iterate over all of the elements. Something like:

```python
element_names = ["images", "labels", ...]
for element in element_names:
    element_dict = getattr(self, element)
    if len(element_dict) == 0:
         continue
    elem_group = root.create_group(name=element)
    for k, v in element_dict.items():
        v.to_zarr()
```



```

File: spatialdata/_core/_spatialdata.py [hunk 11]
```diff
@@ -0,0 +1,262 @@
...
+            if self.labels is not None and el in self.labels.keys():
```
kevinyamauchi, line 138: can self.labels be None?

kevinyamauchi: (same for all elements)

kevinyamauchi (commented):
Looking good so far!  A few initial comments on the `SpatialData` class below

It would be nice if there were some tests to show the expected behavior for the __init__ method (e.g., does a `SpatialData` object without labels have None or an empty dict for self.labels).

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+    ) -> str:
+        def rreplace(s: str, old: str, new: str, occurrence: int) -> str:
+            li = s.rsplit(old, occurrence)
+            return new.join(li)
+
+        def h(s: str) -> str:
+            return s
+            # return hashlib.md5(repr(s).encode()).hexdigest()
+
+        ##
```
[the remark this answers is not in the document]
LucaMarconato: Done it (on the code I will push). I removed `return s` in favor of the hash string (this implements the logic to print nested things by later replacing the hash with the correct graphics elements). Using `s` instead of the hash risks collision with an unfortunate choice of names from the user.

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+    def _gen_repr(
+        self,
+    ) -> str:
+        def rreplace(s: str, old: str, new: str, occurrence: int) -> str:
+            li = s.rsplit(old, occurrence)
+            return new.join(li)
+
+        def h(s: str) -> str:
+            return s
+            # return hashlib.md5(repr(s).encode()).hexdigest()
```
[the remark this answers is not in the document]
LucaMarconato: Same as above

spatialdata/_core/_spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: the `##` are for [PyCharm cell mode](https://plugins.jetbrains.com/plugin/7858-pycharm-cell-mode), I use them while  debugging/coding. Feel free to remove them.

spatialdata/_core/_spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: Done it (on the code I will push). I removed return s in favor of the hash string (this implements the logic to print nested things by later replacing the hash with the correct graphics elements). Using s instead of the hash risks collision with an unfortunate choice of names from the user.

spatialdata/_core/_spatialdata.py
LucaMarconato: se comment above, I removed the `return s` in favor of the hash. The return s was for debugging.

spatialdata/_core/_spatialdata.py
LucaMarconato: removed

spatialdata/_core/_spatialdata.py
LucaMarconato: Created a issue on this since not high priority, but to be done since it will improve the code quality

spatialdata/_core/_spatialdata.py
LucaMarconato: Not sure we can, since the elements are of different classes and we need to have per-element control of what is being written (for example `chunks`, `compressor` etc for images and labels)

spatialdata/_core/_spatialdata.py
LucaMarconato: As the code is written now no, it would likely break things, but the constructors for the elements assures that there is always an empty dict when elements are not specified.

spatialdata/_core/models.py
```diff
@@ -0,0 +1,407 @@
...
+    transform_key = "transform"
+
+    def validate(self, data: AnnData) -> None:
+        if self.coords_key not in data.obsm:
+            raise ValueError(f"AnnData does not contain points coordinates in `adata.obsm['{self.coords_key}']`.")
+        if self.transform_key not in data.uns:
+            raise ValueError(f"AnnData does not contain `{self.transform_key}`.")
+
+    @classmethod
+    def parse(
```
LucaMarconato, line 322: Here there is the need for another array being passed to the parser, in fact since points can't be annotated by tables, we need to include the annotation when constructing them, which is one extra column specifying the cell type for each point.

LucaMarconato: I have included this in the parser, gonna push it soon.

LucaMarconato:
I pushed my code changes. Mostly: 
- I made some improvements/refactoring to the tests and fixed related bugs that I subsequently caught.
- I addressed the previous comments that Kevin and I made to the code

I have some extra comments, I created a issue for each of them:
- https://github.com/scverse/spatialdata/issues/55
- https://github.com/scverse/spatialdata/issues/56
- https://github.com/scverse/spatialdata/issues/57

spatialdata/_core/_spatialdata.py
kevinyamauchi: In that case, in this we should remove the `if * is None` condition.

spatialdata/_core/_spatialdata.py
kevinyamauchi: Thanks!

kevinyamauchi:
Quick comment: I would prefer if the single dispatch methods on the model classes had names (e.g., `_from_numpy`). I think using a generic _ is less readable and harder to understand.

kevinyamauchi:
Also, sorry if I missed it, but are there tests for the models?

I'm not sure that I understand how the model `validate()` methods are intended to be used. My previous understanding was that they would take an object and return True/False or raise a validation error if the object is valid.

spatialdata/_core/_spatialdata.py
LucaMarconato: done thanks

spatialdata/_io/write.py
```diff
@@ -60,121 +68,238 @@ def _write_metadata(
...
+    fmt: Format = SpatialDataFormat(),
+    axes: Optional[Union[str, List[str], List[Dict[str, str]]]] = None,
+    **metadata: Union[str, JSONDict, List[JSONDict]],
+) -> None:
+    sub_group = group.require_group("polygons")
+    polygons_groups = sub_group.require_group(name)
+    coordinate_transformations = [[polygons.attrs.get("transform").to_dict()]]
+    print(coordinate_transformations)
+
+    # TODO: save everything or just polygons?
```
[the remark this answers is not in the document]
LucaMarconato: I wrote the code to save all the rest of the columns, it's rudimentary so please have a check, but works.

giovp:
notes from chat with luca:
- [ ] support for various radius in shapes (for IO)
- [ ] check parser for polygons and add test for passing different json

LucaMarconato:
additional notes (todo):
- [ ] finish adjusting the implicit coordinate systems and test with napari
- [ ] change output hierarchy when saving to zarr to have folders (`images`, `labels`, etc.) with elements inside (so no labels inside images, etc)
- [ ] pass lists instead of dictionaries in SpatialData constructor (this avoid having elements with a key which doesn’t correspond to their future file path)
- [ ] implement a first version of the `NGFFStore`

ivirshup (commented):
Made some broad comments. The which reorganizes parsing transformations I could also do as a commit, but it currently breaks mypy.

## On disk format stability

@giovp, for types that aren't in the ngff spec, how are you thinking about the io methods? Stable, temporary?

My concern here is that I think polygons/ points may need some iteration, and how bad is it if the format changes between now and 0.1. Is it okay if you can't read something you wrote last week, pre v0.1.

Probably fine, but would also be nice if you could read it.

## new types

I want to look at the new types in more depth

spatialdata/_core/transformations.py
```diff
@@ -0,0 +1,937 @@
...
+            assert isinstance(t.output_coordinate_system, CoordinateSystem)
+            t_affine = t.to_affine()
+            target_output_indices = [output_axes.index(ax) for ax in t.output_coordinate_system.axes_names]
+            target_input_indices = [input_axes.index(ax) for ax in t.input_coordinate_system.axes_names] + [-1]
+            m[np.ix_(target_output_indices, target_input_indices)] = t_affine.affine[:-1, :]
+        return Affine(
+            m,
+            input_coordinate_system=self.input_coordinate_system,
+            output_coordinate_system=self.output_coordinate_system,
+        )
```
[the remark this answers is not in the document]
ivirshup: @LucaMarconato would you be up for doing this now? Should this be a follow up PR?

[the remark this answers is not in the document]
LucaMarconato: Thanks a lot for this and the other comments. I am fixing some napari related stuff, right after I will read and merge your changes.

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,282 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
giovp: happy to change, @LucaMarconato let me know when

spatialdata/_io/read.py
```diff
@@ -1,18 +1,58 @@
 import os
+from collections.abc import MutableMapping
```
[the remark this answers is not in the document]
giovp: makes sense, happy to change

giovp:
I do not know what's up with codecov. I tried to disable and re-enable github integration with no success.

giovp:
this is how it looks like locally btw and it seems to succeed in uploading but for whatever reason, it doesn't?
```
---------- coverage: platform darwin, python 3.9.12-final-0 ----------
Name                                     Stmts   Miss  Cover
------------------------------------------------------------
spatialdata/__init__.py                      7      0   100%
spatialdata/_constants/__init__.py           0      0   100%
spatialdata/_constants/_constants.py        10     10     0%
spatialdata/_constants/_enum.py             41     41     0%
spatialdata/_core/__init__.py                0      0   100%
spatialdata/_core/_spatialdata.py          144      4    97%
spatialdata/_core/coordinate_system.py      74      8    89%
spatialdata/_core/core_utils.py             83      7    92%
spatialdata/_core/models.py                268     78    71%
spatialdata/_core/transformations.py       431     20    95%
spatialdata/_io/__init__.py                  2      0   100%
spatialdata/_io/format.py                   62     33    47%
spatialdata/_io/read.py                    134     13    90%
spatialdata/_io/write.py                   129      9    93%
spatialdata/_logging.py                     14      1    93%
spatialdata/_types.py                        8      3    62%
spatialdata/utils.py                        35     13    63%
------------------------------------------------------------
TOTAL                                     1442    240    83%
```

giovp:
ok this seems to be the issue (no solution): https://github.com/codecov/codecov-action/issues/844

spatialdata/_io/format.py
```diff
@@ -52,19 +52,100 @@ def validate_shapes_parameters(
...
+        # calculate minimal 'scale' transform based on pyramid dims
+        for shape in shapes:
+            assert len(shape) == len(data_shape)
+            scale = [full / level for full, level in zip(data_shape, shape)]
+            from spatialdata._core.transformations import Scale
+
+            coordinate_transformations.append([Scale(scale=scale).to_dict()])
+        return coordinate_transformations
+
+    def validate_coordinate_transformations(
```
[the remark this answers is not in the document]
giovp: Yes, I think we can remove it here. I think this was a leftover from taking the formatter from ngff

giovp:
ok done some big cleanup, still few comments and other things but here's latest updates:
- shapes sizes are now stored in `adata.obs` to accommodate for variable shape size across array ( the type could also be variable then e.g. square and circles together but I don't think there are use cases so would keep it as is).
- added formatted and version for shapes, polygons, points (and tables) with spatialdata specific versioning, independent of ngff
- cleanup of read and write, now the folder hierarchy is cleaner and consistent across elements and it is the following:
```
store.zarr/images/sample_1/...
store.zarr/images/sample_2/...
store.zarr/labels/sample_1/...
store.zarr/polygons/...
....
```
one big thing that is still missing (just noticed) is that multiple transform objects per elements are not yet supported. Not difficult but still to do.

spatialdata/_io/write.py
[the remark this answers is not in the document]
LucaMarconato: rolled back, and new code from Giovanni using `.index` to store the instance_values

spatialdata/_core/transformations.py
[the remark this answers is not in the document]
LucaMarconato: Accepted changes, I quickly tried fixing `mypy` but I failed, gonna try later. But also, if you think you can quickly fix it, please go ahead.

spatialdata/_core/_spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: refactoring in https://github.com/scverse/spatialdata/issues/61

spatialdata/_io/format.py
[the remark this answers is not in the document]
LucaMarconato: I think is not needed to add a validation for coordinate transformations and coordinate systems because we are not reusing existing types for storing the information (like `AnnData`, `DataArray` etc) but instead we use a constructor to initialize the objects (even when we call `from_dict()`).

spatialdata/_core/_spatialdata.py
```diff
@@ -0,0 +1,262 @@
...
+
+        if self.table is not None:
+            write_table(table=self.table, group=root, name="table")
+
+    @property
+    def table(self) -> AnnData:
+        return self._table
+
+    @classmethod
+    def read(cls, file_path: str) -> SpatialData:
```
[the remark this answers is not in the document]
LucaMarconato: true, modified

spatialdata/_core/models.py
LucaMarconato: removed

spatialdata/_core/models.py
LucaMarconato: modified and pushed

pyproject.toml
LucaMarconato: resolved after the discussion in Zulip

spatialdata/_core/_spatialdata.py
LucaMarconato: Done.

LucaMarconato:
I addressed all the comments. There are just a few more points to address and then we can merge:
- [ ] finish implementing the tests for models (@giovp)
- [ ] use `GeoDataFrame` instead of `AnnData` for points (me)

tests/_core/test_transformations.py
```diff
@@ -0,0 +1,361 @@
...
+        pass
+
+    # wrong points shapes
+    transformation.output_coordinate_system = output_cs
+    with pytest.raises(ValueError):
+        assert np.allclose(transformation.transform_points(original.ravel()), transformed.ravel())
+    with pytest.raises(ValueError):
+        assert np.allclose(transformation.transform_points(original.transpose()), transformed.transpose())
+    with pytest.raises(ValueError):
+        assert np.allclose(transformation.transform_points(np.expand_dims(original, 0)), np.expand_dims(transformed, 0))
```
kevinyamauchi, line 69: I'm not sure I understand these tests. Is the `assert np.allclose()` necessary? Should it just be 

```python
with pytest.raises(ValueError):
    transformation.transform_points(...)
```

giovp:
ok all models test are now there. the table model is a bit flaky but I think fine for now, expecting changes as we play around with datasets. @LucaMarconato IMHO merge whenever you want

tests/_core/test_transformations.py
LucaMarconato: you are right, changed it

LucaMarconato: pushed by accident in `napari_fix`, so gonna be merged when we merge that branch (very soon)

LucaMarconato:
Great @giovp, good for me. Please @kevinyamauchi or @ivirshup merge anytime

LucaMarconato (approved):

kevinyamauchi (approved):
Thanks for this massive effort @giovp and @LucaMarconato ! This looks like a great start. I propose to optimistically merge this tonight.

I have made a few comments below that I will turn into issues so we can address them as the code base stabilizes.

Thanks again ! 🚀 


- There are a lot of commented out blocks of code. We should delete them.
- Many functions are missing docstrings
- The tests look fairly comprehensive, but it is difficult to tell what they are testing. I think we should both break them up into smaller tests and add comments/docstrings to explain what they are testing. Tests are an important form of documentation for contributors - especially contributors who are not part of writing the original code!

spatialdata/_core/models.py
```diff
@@ -0,0 +1,578 @@
+"""This file contains models and schema for SpatialData"""
```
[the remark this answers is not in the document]
kevinyamauchi: Made an issue #66

spatialdata/_core/coordinate_system.py
```diff
@@ -87,12 +87,12 @@ def name(self) -> Optional[str]:
         return self._name
 
     @property
-    def axes_names(self) -> List[str]:
-        return [ax.name for ax in self._axes]
+    def axes_names(self) -> Tuple[str, ...]:
```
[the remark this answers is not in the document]
kevinyamauchi: Made an issue #67

spatialdata/_core/transformations.py
```diff
@@ -0,0 +1,911 @@
...
+    # None or strings. Anyway, in order to be able to apply a transformation to a DataArray, we need to know the name of
+    # the input and output axes (information contained in the CoordinateSystem object). Therefore, the following
+    # variables will be populated with CoordinateSystem objects when both the coordinate_system and the transformation
+    # are known.
+    # Example: as a user you have an Image (cyx) and a Point (xy) elements, and you want to contruct a SpatialData
+    # object containing the two of them. You also want to apply a Scale transformation to the Image. You can simply
+    # assign the Scale transformation to the Image element, and the SpatialData object will take care of assiging to
+    # "_input_coordinate_system" the intrinsitc coordinate system of the Image element,
+    # and to "_output_coordinate_system" the global coordinate system (cyx) that will be created when calling the
+    # SpatialData constructor
```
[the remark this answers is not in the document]
LucaMarconato: The code logic changed a bit. I added a todo to see in a while if this is still relevant and either remove it or put it in the docstring.

spatialdata/_core/transformations.py
```diff
@@ -0,0 +1,911 @@
...
+        output_coordinate_system: Optional[Union[str, CoordinateSystem]] = None,
+    ) -> None:
+        """
+        class for storing rotation transformations.
+        """
+        super().__init__(input_coordinate_system, output_coordinate_system)
+        self.rotation = self._parse_list_into_array(rotation)
+
+    @classmethod
+    def _from_dict(cls, d: Transformation_t) -> Self:  # type: ignore[valid-type]
```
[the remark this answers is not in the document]
LucaMarconato: Yes but `mypy` complains when it see `Self`. Maybe someone have a look please.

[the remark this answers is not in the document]
giovp: this is indeed weird but I think Self might be valid in newer python versions?

giovp:
LGTM!

PR #76, opened 2022-12-12
Fix/codecov experiment

ivirshup:
Experimenting to get #71 working


ivirshup:
@giovp, it takes a little bit for the coverage action to run, and it's not added to the checks until after it runs.

Maybe we should make it a required check?

.github/workflows/test.yaml
```diff
@@ -45,20 +45,18 @@ jobs:
               run: |
                   python -m pip install --upgrade pip wheel
                   pip install pytest-cov
-                  pip install codecov
             - name: Install dependencies
               run: |
-                  pip install ".[dev,test]"
+                  pip install -e ".[dev,test]"
```
giovp, line 50: do you think this made it work? or what was it otherwise?

giovp (approved):

giovp:
> Maybe we should make it a required check?

we could make it a required check

.github/workflows/test.yaml
ivirshup: Yes, it wasn't collecting the coverage otherwise. I can locally reproduce that installing without `-e` prevents collection of the coverage. 

I vaguely remember this being an issue before. I think it came up for anndata – though it's not using `-e` now and I can't tell what's different 😅. There's also some reason we make the cookiecutter do an editable install too.

I think it had something to do with the install being a wheel, and being able to trace the code being tested to the wheel?

PR #77, opened 2022-12-12
Make geopandas use shapely on import

### LINKED ISSUE #73: force shapely 2.0?
@giovp

When we import spatial data we get a warning from geopandas saying:

```python
UserWarning: Shapely 2.0 is installed, but because PyGEOS is also installed, GeoPandas will still use PyGEOS by default for now. To force to use and test Shapely 2.0, you have to set the environment variable USE_PYGEOS=0. You can do this before starting the Python process, or in your code before importing geopandas:

import os
os.environ['USE_PYGEOS'] = '0'
import geopandas

In a future release, GeoPandas will switch to using Shapely by default. If you are using PyGEOS directly (calling PyGEOS functions on geometries from GeoPandas), this will then stop working and you are encouraged to migrate from PyGEOS to Shapely 2.0 (https://shapely.readthedocs.io/en/latest/migration_pygeos.html).
```

Should we just go ahead and set:

```python
import os
os.environ['USE_PYGEOS'] = '0'
```

Before importing geopandas? + warn if it's already been imported in this environment?

This could then be documented in the installation instructions.

I'm thinking this could cut down on debugging against both versions.=

ivirshup:
Fixes #73

Now I do the change dynamically, instead of modifying the environment. I also tell people how to undo the modification in case it's actually important.

### DIFF
--- a/spatialdata/_compat.py
+++ b/spatialdata/_compat.py
@@ -1,0 +1,25 @@
+def _check_geopandas_using_shapely() -> None:
+    """
+    Checking that geopandas is using shapely, until it becomes the default.
+    """
+    import warnings
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            "Shapely 2.0 is installed, but because PyGEOS is also installed",
+            category=UserWarning,
+        )
+        import geopandas
+
+        if geopandas.options.use_pygeos is True:
+            geopandas.options.use_pygeos = False
+
+            warnings.warn(
+                (
+                    "Geopandas was set to use PyGEOS, changing to shapely 2.0 with:"
+                    "\n\n\tgeopandas.options.use_pygeos = True\n\n"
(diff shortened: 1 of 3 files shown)


spatialdata/_compat.py
```diff
@@ -0,0 +1,25 @@
...
+            geopandas.options.use_pygeos = False
+
+            warnings.warn(
+                (
+                    "Geopandas was set to use PyGEOS, changing to shapely 2.0 with:"
+                    "\n\n\tgeopandas.options.use_pygeos = True\n\n"
+                    "If you intended to use PyGEOS, set the option to False."
+                ),
+                UserWarning,
+            )
```
giovp, line 25: think this is fine but we have a logger that could also be used.

giovp (approved):

LucaMarconato:
FYI, Shapely 2.0 is now out! https://shapely.readthedocs.io/en/latest/release.html

Issue #64, opened 2022-12-09
The parsers ignore eventual transformations already present in the element


LucaMarconato:
- [x] The parsers ignore eventual transformations already present in the element
Easy fix

LucaMarconato:
Also the parser leads to an output coordinate system which is ignoring the units already presents, probably also the name.

LucaMarconato:
Fixed by a long time, by the function `_parse_transformations()` in `models.py`, and tested extensively in `_parse_transformation_from_multiple_places()` in `test_models.py`.

PR #83, opened 2022-12-15
Improvements and bugfix

LucaMarconato:
Small improvements and bugfix:
- [ ] added property `coordinate_system` to `SpatialData`
- [ ] fixed bug on physical units descried in https://github.com/scverse/spatialdata/issues/64#issuecomment-1351823357
- [ ] fixed some typings of `Mapping` vs `Dict`


kevinyamauchi (commented):
This is looking good, @LucaMarconato ! Can you please be sure to add tests that fail on main for the bugs (i.e., tests that would have caught the bug in case it pops up again) that you're fixing? Thanks!

LucaMarconato:
Sure! For this particular case I will add in general more comprehensive tests for the coordinate systems after I write some helper functions. But for future bugs I'll do that.

PR #93, opened 2022-12-22
Bounding box query on points

kevinyamauchi:
This PR adds the infrastructure to perform a bounding box query on a SpatialData object. This implements it for points point to demonstrate the functionality. If we like the API, we can extend to other elements in follow up PRs.

See this comment for an overview: https://github.com/scverse/spatialdata/pull/93#issuecomment-1363292335

### DIFF
--- a/spatialdata/_core/_spatial_query.py
+++ b/spatialdata/_core/_spatial_query.py
@@ -77,7 +77,7 @@ class SpatialQueryManager:
 
 def _get_spatial_axes(
     coordinate_system: CoordinateSystem,
-) -> List[str]:
+) -> list[str]:
     """Get the names of the spatial axes in a coordinate system.
 
     Parameters
--- a/spatialdata/_core/_spatial_query.py
+++ b/spatialdata/_core/_spatial_query.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING
 
 import numpy as np
 import pyarrow as pa
@@ -17,6 +17,12 @@ class BaseSpatialRequest:
     """Base class for spatial queries."""
 
     coordinate_system: CoordinateSystem
+
+    def __post_init__(self) -> None:
+        # validate the coordinate system
+        spatial_axes = _get_spatial_axes(self.coordinate_system)
+        if len(spatial_axes) == 0:
+            raise ValueError("No spatial axes in the requested coordinate system")
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


kevinyamauchi:
here's a gist demonstrating the API:

https://gist.github.com/kevinyamauchi/5005c37556a339481255673e1ffdf089

kevinyamauchi:
We can add additional queries to the SpatialQuery manager as we develop them. We can also use the SpatialQueryManager to cache things, build spatial indices, etc. I prefer to encapsulate the functionality in a class because it allows us to more easily swap backends in the future and keeps the API smaller on the SpatialData object.

spatialdata/_core/_spatial_query.py
```diff
@@ -0,0 +1,92 @@
...
+    max_coordiate : np.ndarray
+        The coordinate of the upper right hand corner (i.e., maximum values)
+        of the bounding box
+    """
+
+    min_coordinate: np.ndarray  # type: ignore[type-arg]
+    max_coordinate: np.ndarray  # type: ignore[type-arg]
+
+
+def _get_spatial_axes(
```
LucaMarconato, line 48: I'll move this as a method in `CoordinateSystem`

LucaMarconato:
Looks very good to me! I answer your questions:
1) there are no convenient APIs to do that yet, but it's one of my first next points in my todo list to make a pr to make that convenient.
2) we could make such a function and I like the idea but I would wait a bit to see how the elements evolve. I would start by making some helper functions to tell which type an element is in a separate pr, and then continue from there. I added this pint to my non-urgent todos.
3) I like to have a query object inside the `SpatialData` object, but to reflect the [design doc (see last example)](https://github.com/scverse/spatialdata/blob/main/docs/design_doc.md) here I propose to actually have a class `QueryManager` instead of/in addition to the `SpatialQueryManager` and to rename the attribute `spatial` to `query`. The rationale is that every type of query, spatial or non-spatial (e.g. filtering by cell-type), would have a common syntax: `sdata.query.my_query_method(my_request)`.

kevinyamauchi:
Thanks for the review, @LucaMarconato !

> there are no convenient APIs to do that yet, but it's one of my first next points in my todo list to make a pr to make that convenient.

Okay, in that case, I will punt and make an issue on the transformation stuff until that is merged. I will add some comments and stubs to document my thoughts.

> we could make such a function and I like the idea but I would wait a bit to see how the elements evolve. I would start by making some helper functions to tell which type an element is in a separate pr, and then continue from there. I added this pint to my non-urgent todos.

Makes sense to me! I will make an issue for this.

> I like to have a query object inside the SpatialData object, but to reflect the [design doc (see last example)](https://github.com/scverse/spatialdata/blob/main/docs/design_doc.md) here I propose to actually have a class QueryManager instead of/in addition to the SpatialQueryManager and to rename the attribute spatial to query. The rationale is that every type of query, spatial or non-spatial (e.g. filtering by cell-type), would have a common syntax: sdata.query.my_query_method(my_request).

Sounds good. We may want the `QueryManager` to contain a `SpatialQueryManager`, `ObsQueryManager`, etc. to keep things tidy, but we can always add that hierarchy in later.

spatialdata/_core/_spatial_query.py
kevinyamauchi: Done!

kevinyamauchi:
@LucaMarconato , I have addressed your comments. It would be great to try and merge soon so that we can move on to adding the bounding box query for other element types.

LucaMarconato (approved):

Issue #47, opened 2022-11-02
Using xarray to make transformations less ambiguous


LucaMarconato:
Cost/benefit analysis of using `xarray` to annotate the dimensions of vectors (see code examples below).
Pro:
- would make transformations less ambiguous
- possibility to apply transformations to points of any shapes, for instance (n, 2) and (2, n).

Cons:
- extra verbosity for the use

Current approach: initially I am using `numpy` arrays for the vectors describing points and performing a shape check (e.g. 2D translations can't be applied to a set of points stored as a `numpy` array of shape `(2, n)`, but only with a shape `(n, 2)` (excluding edge cases like n=2).
Later we can discuss the preferred approach.

Code examples from @ivirshup of using `xarray.DataArray` to annotate the dimensions of vectors.
```
import xarray as xr

da = xr.DataArray([1,2,3], coords={"dim": ["x", "y", "z"]})

da.sel({"dim": "x"})
<xarray.DataArray ()>
array(1)
Coordinates:
    dim     <U1 'x'
```

```
import xarray as xr, numpy as np

N_POINTS = 10
points = xr.DataArray(
    np.random.random((N_POINTS, 3)),
    coords={
        "points": range(N_POINTS),
        "dim": ["x", "y", "z"],
    },
)
points
<xarray.DataArray (points: 10, dim: 3)>
array([[0.97379558, 0.29238395, 0.37392129],
       [0.62459118, 0.13214354, 0.0514482 ],
       [0.41742495, 0.59828447, 0.79762354],
       [0.87583248, 0.71631053, 0.30032732],
       [0.93012066, 0.73074886, 0.46370001],
       [0.74505332, 0.84503157, 0.52606046],
       [0.85789939, 0.44050558, 0.23522551],
       [0.35954281, 0.11675383, 0.23321456],
       [0.99866969, 0.83124205, 0.55161158],
       [0.25951685, 0.09205387, 0.00736066]])
Coordinates:
  * points   (points) int64 0 1 2 3 4 5 6 7 8 9
  * dim      (dim) <U1 'x' 'y' 'z'
```

```
translation = xr.DataArray([0, 1, 0], coords={"dim": ["x", "y", "z"]})
points + translation
```

ivirshup:
[...]

PR #100, opened 2023-01-08
3 improviements to transforms: 1. helper function to get an affine tr…

LucaMarconato:
…ansform from input and output coordinate system; 2. fixed but in sequence (now inferring coordinate systems from its components); 3. added missing test for inferencen of intermediate coordinate systems for nested sequences

### DIFF
--- a/spatialdata/_core/transformations.py
+++ b/spatialdata/_core/transformations.py
@@ -473,16 +473,6 @@ class Affine(BaseTransformation):
             input_coordinate_system=self.output_coordinate_system,
             output_coordinate_system=self.input_coordinate_system,
         )
-        # old code, manually inverting a 2d (3x3) affine matrix
-        # a = self.affine[0, 0]
-        # b = self.affine[0, 1]
-        # m = self.affine[0, 2]
-        # c = self.affine[1, 0]
-        # d = self.affine[1, 1]
-        # n = self.affine[1, 2]
-        # det = a * d - b * c
-        # closed_form = np.array([[d, -c, 0], [-b, a, 0], [b * n - d * m, c * m - a * n, det]])
-        # return Affine(affine=closed_form)
 
     def _get_and_validate_axes(self) -> tuple[tuple[str, ...], tuple[str, ...]]:
         input_axes, output_axes = self._get_axes_from_coordinate_systems()
@@ -500,6 +490,33 @@ class Affine(BaseTransformation):
             self.affine,
             input_coordinate_system=self.input_coordinate_system,
             output_coordinate_system=self.output_coordinate_system,
+        )
+
+    @classmethod
+    def _affine_matrix_from_input_and_output_axes(cls, input_axes: tuple[str], output_axes: tuple[str]) -> ArrayLike:
+        from spatialdata._core.core_utils import C, X, Y, Z
+
+        assert all([ax in (X, Y, Z, C) for ax in input_axes])
+        assert all([ax in (X, Y, Z, C) for ax in output_axes])
+        m = np.zeros((len(output_axes) + 1, len(input_axes) + 1))
+        for output_ax in output_axes:
+            for input_ax in input_axes:
+                if output_ax == input_ax:
+                    m[output_axes.index(output_ax), input_axes.index(input_ax)] = 1
+        m[-1, -1] = 1
+        return m
+
+    @classmethod
+    def from_input_output_coordinate_systems(
+        cls,
+        input_coordinate_system: Optional[Union[str, CoordinateSystem]],
+        output_coordinate_system: Optional[Union[str, CoordinateSystem]],
+    ) -> Affine:
+        input_axes = input_coordinate_system.axes_names
+        output_axes = output_coordinate_system.axes_names
+        m = cls._affine_matrix_from_input_and_output_axes(input_axes, output_axes)
+        return cls(
+            affine=m, input_coordinate_system=input_coordinate_system, output_coordinate_system=output_coordinate_system
         )
 
 
@@ -569,9 +586,19 @@ class Sequence(BaseTransformation):
         output_coordinate_system: Optional[Union[str, CoordinateSystem]] = None,
     ) -> None:
         super().__init__(input_coordinate_system, output_coordinate_system)
-        # we can treat this as an Identity if we need to
+        # we can decide to treat an empty sequence as an Identity if we need to
         assert len(transformations) > 0
         self.transformations = transformations
+        if (cs := self.transformations[0].input_coordinate_system) is not None:
+            if self.input_coordinate_system is not None:
+                assert cs == self.input_coordinate_system
+            else:
+                self.input_coordinate_system = cs
+        if (cs := self.transformations[-1].output_coordinate_system) is not None:
+            if self.output_coordinate_system is not None:
+                assert cs == self.output_coordinate_system
+            else:
+                self.output_coordinate_system = cs
 
     @classmethod
     def _from_dict(cls, d: Transformation_t) -> Self:  # type: ignore[valid-type]
@@ -633,7 +660,7 @@ class Sequence(BaseTransformation):
                 f"replaced by the CoordinateSystem named after the string before calling this function."
             )
         else:
-            assert isinstance(input_cs, CoordinateSystem)
+            assert isinstance(input_cs, CoordinateSystem), input_cs
             assert input_cs == latest_output_cs
         output_cs = t.output_coordinate_system
         expected_output_cs = Sequence._inferring_cs_infer_output_coordinate_system(t)
--- a/tests/_core/test_transformations.py
+++ b/tests/_core/test_transformations.py
@@ -40,14 +40,20 @@ def _test_transformation(
     test_affine_inverse: bool = True,
     test_inverse: bool = True,
 ):
-    # missing input and output coordinate systems
-    with pytest.raises(ValueError):
-        assert np.allclose(transformation.transform_points(original), transformed)
+    # missing input and output coordinate systems.
+    # If the transformation is a Sequence, it can have the input_coordinate system specified (inherited from the
+    # first component). In this case the test is skipped
+    if transformation.input_coordinate_system is None:
+        with pytest.raises(ValueError):
+            assert np.allclose(transformation.transform_points(original), transformed)
 
     # missing output coordinate system
+    # If the transformation is a Sequence, it can have the output_coordinate system specified (inherited from the
+    # last component). In this case the test is skipped
     transformation.input_coordinate_system = input_cs
-    with pytest.raises(ValueError):
-        assert np.allclose(transformation.transform_points(original), transformed)
+    if transformation.output_coordinate_system is None:
+        with pytest.raises(ValueError):
+            assert np.allclose(transformation.transform_points(original), transformed)
 
     # wrong output coordinate system
     transformation.output_coordinate_system = wrong_output_cs
@@ -287,7 +293,7 @@ def test_sequence():
         output_coordinate_system=cyx_cs,
     )
 
-    def _manual_xy_to_cyz(x: np.ndarray) -> np.ndarray:
+    def _manual_xy_to_cyx(x: np.ndarray) -> np.ndarray:
         return np.hstack((np.zeros(len(x)).reshape((len(x), 1)), np.fliplr(x)))
 
     _test_transformation(
@@ -316,12 +322,37 @@ def test_sequence():
                 xy_to_cyx,
             ]
         ),
-        original=_manual_xy_to_cyz(original),
-        transformed=_manual_xy_to_cyz(transformed),
+        original=_manual_xy_to_cyx(original),
+        transformed=_manual_xy_to_cyx(transformed),
         input_cs=cyx_cs,
         output_cs=cyx_cs,
         wrong_output_cs=xyc_cs,
         test_inverse=False,
+    )
+
+    # test sequence inside sequence, with full inference of the intermediate coordinate systems
+    # two nested should be enought, let's test even three!
+
+    _test_transformation(
+        transformation=Sequence(
+            [
+                Scale(np.array([2, 3])),
+                Sequence(
+                    [
+                        Scale(np.array([4, 5])),
+                        Sequence(
+                            [Scale(np.array([6, 7]))],
+                        ),
+                    ],
+                ),
+            ]
+        ),
+        original=original,
+        transformed=original * np.array([2 * 4 * 6, 3 * 5 * 7]),
+        input_cs=yx_cs,
+        output_cs=yx_cs,
+        wrong_output_cs=xy_cs,
+        test_inverse=True,
     )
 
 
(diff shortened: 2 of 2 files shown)


LucaMarconato:
As mentioned in my previous post, a downside of " setter for elements to check that the transfomration makes sense (if not, tries to fix it)" is that very verbose transformations are generated.

For instance this code here:

LucaMarconato:
After discussing with @giovp I will refactor into separate classes to divide what is purely a NGFF transform from what is built on top of it for ergonomics. The reasons are the following:
1) the NGFF classes can be in principle moved to `ome-zarr-py` or an object-oriented representation of the storage, which is something that we want;
2) this approach will greatly simplify the code complexity. The NGFF classes will be used only when reading or writing. The in-memory representation will take advantage of the new classes.
3) The new classes will be based on `xarray` like described here https://github.com/scverse/spatialdata/issues/47 and the order of axes will be irrelevant. Also, all the transformations will have non-ambiguous information for the axis (on the contrary NGFF Sequence transform allow for containing transformations which don't specify the axes. This increases the code complexity.
4) New functionalities that will be easily supported will be the possibility to concatenate any sequence of transformations independently of the axes they operate on, and assign any transformation to any element, again independently of the axes they work on.

The old version of the code can be found in the [previous commit](https://github.com/scverse/spatialdata/pull/100/commits/20f2e3a79bb3417db7c6dadaa5e93b68237f43b4), as well as in a [separate branch](https://github.com/scverse/spatialdata/tree/archive/only_ngff_transforms) created from it. If the new code works as planned I'll delete the branch (which I already tagged for archiving purposes and [can be found here](https://github.com/scverse/spatialdata/commit/20f2e3a79bb3417db7c6dadaa5e93b68237f43b4)).

LucaMarconato:
Also, before the review I would suggest to look at this other PR and merge it to main, so that the diff is a bit smaller: https://github.com/scverse/spatialdata/pull/113

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+        Notes
+        -----
+        - You can also use the static method SpatialData.set_transformation_in_memory() to set the transformation of an
+        element when it is not backed
+        """
+        self.set_transformation_in_memory(element, transformation, target_coordinate_system)
+        self._write_transformations_to_disk(element)
+
+    @staticmethod
+    def remove_transformation_in_memory(
```
giovp, line 409: same for this and below

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+
+        Notes
+        -----
+        - You can also use the static method SpatialData.remove_transformation_in_memory() to remove the transformation of
+        an element when it is not backed
+        """
+        self.remove_transformation_in_memory(element, target_coordinate_system)
+        self._write_transformations_to_disk(element)
+
+    @staticmethod
```
giovp, line 457: same again for this, I think there should be 1 method only. Also, when do we need to "remove" transformations>?

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+        ):
+            return Identity()
+        else:
+            g = self._build_transformations_graph()
+            src_node: Union[int, str]
+            if has_type_spatial_element(source_coordinate_system):
+                src_node = id(source_coordinate_system)
+            else:
+                src_node = source_coordinate_system
+            des_node: Union[int, str]
```
giovp, line 543: `des_node` -> `tgt_node` ? although I like the german influence 😅

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+                    # an expection and ask the user to be more specific
+                    paths_with_length_1 = [p for p in paths if len(p) == 2]
+                    if len(paths_with_length_1) == 1:
+                        path = paths_with_length_1[0]
+                    else:
+                        # error 1
+                        raise RuntimeError(
+                            "Multiple paths found between the two coordinate systems. Please specify an intermediate "
+                            "coordinate system."
+                        )
```
giovp, line 564: maybe it's worth here to print the path to tie it to what went wrong?

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+        elements: dict[str, dict[str, SpatialElement]] = {}
+        for element_type, element_name, element in self._gen_elements():
+            transformations = self.get_all_transformations(element)
+            if coordinate_system in transformations:
+                if element_type not in elements:
+                    elements[element_type] = {}
+                elements[element_type][element_name] = element
+        return SpatialData(**elements, table=self.table)
+
+    def transform_element_to_coordinate_system(
```
giovp, line 611: I would remove this as API wise I don't super get it, will elaborate in comment

giovp (changes requested):
So one major comment I have so far (by just checking the `_spatialdata.py` file) is the type of API we want to have. Copying it from the design doc

```python
import spatialdata as sd

sdata = SpatialData(...)
points = sd.transform(sdata.points["image1"], tgt="tgt_space")
sdata = sd.transform(sdata, tgt="tgt_space")
```

I think we should have 1 way to apply transform to either spatialdata or elements, and maybe we also shouldnt' allow user to pass directly the element, but only the name and type (e.g. {"labels":"sample1"}. I think this would prevent mistakes in copy/viewes of elements from the ones already in spatialdata and the ones transformed. wdyt ?

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+        target_coordinate_system: Optional[str] = None,
+    ) -> None:
+        """
+        Set/replace a transformation of an element, writing it to disk if the SpatialData object is backed.
+
+        Parameters
+        ----------
+        element
+            The element to replace the transformation for.
+        transformation
```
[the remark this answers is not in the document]
LucaMarconato: Unfortunately not, one is a static method, the other is a normal method.

LucaMarconato:
Here is the example notebook showcasing transformations (and also a bit of io, mostly related to transformations).

https://github.com/scverse/spatialdata-notebooks/blob/main/notebooks/transformations_examples.ipynb

pyproject.toml
```diff
@@ -70,6 +71,8 @@ xfail_strict = true
 addopts = [
 #    "-Werror",  # if 3rd party libs raise DeprecationWarnings, just use filterwarnings below
     "--import-mode=importlib",  # allow using test files with same name
+    "-x",  # exit on first test if it fails
```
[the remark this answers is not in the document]
LucaMarconato: Right, removed.

spatialdata/_core/_spatialdata.py
```diff
@@ -179,28 +200,45 @@ def _add_labels_in_memory(self, name: str, labels: Union[SpatialImage, Multiscal
...
         self._points[name] = points
 
     def is_backed(self) -> bool:
         """Check if the data is backed by a Zarr storage or it is in-memory."""
         return self.path is not None
 
+    def _get_group_for_element(self, name: str, element_type: str) -> zarr.Group:
+        store = parse_url(self.path, mode="r+").store
+        root = zarr.group(store=store)
+        assert element_type in ["images", "labels", "points", "polygons", "shapes"]
```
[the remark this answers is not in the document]
LucaMarconato: added a todo in the code

spatialdata/_core/_spatialdata.py
```diff
@@ -242,12 +280,391 @@ def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> za
...
+                from spatialdata._io.write import (
+                    overwrite_coordinate_transformations_non_raster,
+                )
+
+                overwrite_coordinate_transformations_non_raster(group=group, axes=axes, transformations=transformations)
+            else:
+                raise ValueError("Unknown element type")
+
+    @staticmethod
+    def get_all_transformations(element: SpatialElement) -> MappingToCoordinateSystem_t:
```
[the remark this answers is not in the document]
LucaMarconato: Maybe we could do it. I would keep the default behaviour of target_coordinate_system to be `None` (well actually "global", gonna change it now). We can have an argument `return_all: bool = False` and change the return type to `Union[BaseTransformation, dict[str, BaseTransformation]`.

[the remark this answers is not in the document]
LucaMarconato: We could also make `target_coordinate_system: Optional[str] = DEFAULT_COORDINATE_SYSTEM` (="global") and return all the transformations if it is set to None by the user (but it feels a bit weird).

spatialdata/_core/_spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: In particular, the non-static method takes a `SpatialElement` as input and finds it inside the `SpatialData` object. In this way we can support changing also the disk content automatically. The disk is updated when `sdata` is backed, that is when `sdata.path` is not `None`.

spatialdata/_core/_spatialdata.py
LucaMarconato: changed, both you and copilot voted for `tgt_node` ahah

spatialdata/_core/_spatialdata.py
LucaMarconato: Good point, added it. Now the exception looks like this:
```
>                       raise RuntimeError(
                            "Multiple paths found between the two coordinate systems. Please specify an intermediate "
                            f"coordinate system. Available paths are:{s}"
                        )
E                       RuntimeError: Multiple paths found between the two coordinate systems. Please specify an intermediate coordinate system. Available paths are:
E                           my_space0 -> /images/image2d_multiscale -> global
E                           my_space0 -> /labels/labels2d -> global
```

LucaMarconato: Actually I changed it to this so it's clearer what the user should plug in the function call:
```
E                       RuntimeError: Multiple paths found between the two coordinate systems. Please specify an intermediate coordinate system. Available paths are:
E                           my_space0 -> <sdata>.images['image2d_multiscale'] -> global
E                           my_space0 -> <sdata>.labels['labels2d'] -> global
```

spatialdata/_core/_spatialdata.py
LucaMarconato: Ok thanks, answering in the comment.

LucaMarconato:
> So one major comment I have so far (by just checking the _spatialdata.py file) is the type of API we want to have. Copying it from the design doc
> 
> import spatialdata as sd
> from spatialdata import SpatialData
> 
> sdata = SpatialData(...)
> points = sd.transform(sdata.points["image1"], tgt="tgt_space")
> sdata = sd.transform(sdata, tgt="tgt_space")
> I think we should have 1 way to apply transform to either spatialdata or elements, and maybe we also shouldnt' allow user to pass directly the element, but only the name and type (e.g. {"labels":"sample1"}. I think this would prevent mistakes in copy/viewes of elements from the ones already in spatialdata and the ones transformed. wdyt ?

I think that we need two functions:
1. one for getting a transformation between one target and one source, and this would need to be a method of `SpatialData` since it needs to build the digraph of the transformations of the elements that are inside. This is currently the function `map_coordinate_systems()`. We can change the name, it's a bit confusing. Maybe `map()`?
2. one method to actually transform one element to a coordinate system. Again this needs to be a method of `SpatialData` because it needs to call the function above internally.

This is the reason why I couldn't go with this idea from the design doc:
```
points = sd.transform(sdata.points["image1"], tgt="tgt_space")
```

Still we could change the second function to this:
```
points = sd.transform(sdata, element_type='labels', element_name='cells', tgt="tgt_space")
```

Here are the pro and cons of this second approach.
Pro:
* we can rewrite and `set_transformation()`, `remove_transformation()` and `remove_all_transformations()` ato take as first argument the `sdata` object and then the string description of the element like above. In this way we can remove the `*_in_memory()` methods since all the functions can be static methods (or just global functions). I really like this.
Con:
* passing both the `element_type` and the `element_name` is (ihmo very) cumbersome. One way to get around this would be to require unique names for elements. I explore this possibility in this issue that I have just created: https://github.com/scverse/spatialdata/issues/124

My proposal is to require unique names for elements and after that to change the second function as described. I would do this in a separate pr (I don't think we will have problems of copy/views, things should work right now).

kevinyamauchi (approved):
Thanks @LucaMarconato ! I think this is a solid step in the right direction. I am approving so that we can keep things moving forward. I think we will have to continue to iterate on API, etc., but it will be easier to do so when we actually use it to make vignettes, etc.

> one for getting a transformation between one target and one source, and this would need to be a method of SpatialData since it needs to build the digraph of the transformations of the elements that are inside. This is currently the function map_coordinate_systems(). We can change the name, it's a bit confusing. Maybe map()?

Perhaps it could be called something like `get_transformation_between_coordinate_systems()`? Probably too long, but more descriptive. In any case, I agree that `map_coordinate_systems()` is probably not the most clear name. 

Does this method actually need to be on `SpatialData`? It seems to me that one could still construct the directed graph of the transformations on a given element. Perhaps I'm missing something though.

spatialdata/_core/_spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: I have removed these functions and added new global methods with better signature (I put them in `_spatialdata.py`, but probably we want to move them somewhere when we will refactor the import structure of the subpackages.

spatialdata/_core/_spatialdata.py
LucaMarconato: I have removed these functions and added new global methods with better signature (I put them in `_spatialdata.py`, but probably we want to move them somewhere when we will refactor the import structure of the subpackages.

LucaMarconato:
Thanks for the comments @kevinyamauchi, I have removed the functions `get`/`set`/`remove` transformations (total of 8 functions) and replaced with 3 functions living outside the `SpatialData` class. I have also moved `map_coordinate_systems` out of the `SpatialData` class and renamed it to `get_transformation_between_coordinate_systems()`.

LucaMarconato:
I will now merge the new points io pr, it will take some times to fix all the tests, but then we'll be ready to merge!

spatialdata/_core/_spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: replace with better methods

giovp:
@LucaMarconato quick thing I noticed from working in #132 : the `PointsModel.parse` still requires an `annotation` to be passed in the case of `nd.array` input. I think this should not be enforced since the `feature_key` is not mandatory. Was it fixed here or shall I push a quick fix in main?

LucaMarconato:
Hi, true I didn't fix it actually. A quick fix would be great thanks.

PR #105, opened 2023-01-18
Add image and polygons bounding box query

kevinyamauchi:
This PR adds bounding box queries for the image, labels and polygons


kevinyamauchi:
It looks like the test failures are unrelated (they are writers). @LucaMarconato  or @giovp , have you seen this before?

kevinyamauchi:
I have created an example of using the bounding box query on a SpatialData object containing images, points, and polygons here: https://gist.github.com/kevinyamauchi/29698d258077cf15c26c1ed042f81bb0

I think this is ready for review! The test failures appear unrelated. The pre-commit failure is not from code I touched. Should I re-run pre-commit on those files locally and push?

spatialdata/_core/_spatial_query.py
```diff
@@ -85,3 +89,129 @@ def _bounding_box_query_points_dict(
...
+        output_coordinate_system=coordinate_system,
+    )
+
+    new_transformation = Sequence(
+        [translation, initial_transform],
+        input_coordinate_system=coordinate_system,
+        output_coordinate_system=initial_transform.output_coordinate_system,
+    )
+
+    query_result.attrs["transform"] = new_transformation
```
kevinyamauchi, line 153: should I use the set_transform convenience function here?

spatialdata/_core/_spatial_query.py
```diff
@@ -85,3 +89,129 @@ def _bounding_box_query_points_dict(
...
+        # add the
+        selection[axis_name] = slice(min_value, max_value)
+
+    query_result = image.sel(selection)
+
+    # update the transform
+    # currently, this assumes the existing transforms input coordinate system
+    # is the intrinsic coordinate system
+    # todo: this should be updated when we support multiple transforms
+    initial_transform = query_result.transform
```
kevinyamauchi, line 131: should I use the get_transform convenience function here?

spatialdata/_core/_spatial_query.py
```diff
@@ -85,3 +89,129 @@ def _bounding_box_query_points_dict(
...
+
+        # add the
+        selection[axis_name] = slice(min_value, max_value)
+
+    query_result = image.sel(selection)
+
+    # update the transform
+    # currently, this assumes the existing transforms input coordinate system
+    # is the intrinsic coordinate system
+    # todo: this should be updated when we support multiple transforms
```
kevinyamauchi, line 130: this whole section will have to be refactored when #100 lands. However, in order to keep things moving, I would like to merge this and refactor later if possible.

spatialdata/_core/_spatial_query.py
```diff
@@ -85,3 +89,129 @@ def _bounding_box_query_points_dict(
             requested_points[points_name] = points
 
     return requested_points
+
+
+def _bounding_box_query_image(
+    image: Union[MultiscaleSpatialImage, SpatialImage], request: BoundingBoxRequest
```
giovp, line 95: with `__annotations__` we could replace Union with `FirstType | SecondType`. Just a comment, since if we settle with that we'd have to change it everywhere.

spatialdata/_core/_spatial_query.py
```diff
@@ -85,3 +89,129 @@ def _bounding_box_query_points_dict(
...
+    Parameters
+    ----------
+    image : Union[MultiscaleSpatialImage, SpatialImage]
+        The image element to perform the query on.
+    request : BoundingBoxRequest
+        The request for the query.
+
+    Returns
+    -------
+    query_result : Union[MultiscaleSpatialImage, SpatialImage]
```
giovp, line 108: I don't think we use arguments in `Returns`

spatialdata/_core/_spatial_query.py
giovp: I think this will change also because we'll have a mapping in the `attrs["transform"]` so I think for now it's fine?

giovp (changes requested):
really nice and clean code, left some comments re docstrings mostly for consistencies with rest of repo.

spatialdata/_core/_spatial_query.py
kevinyamauchi: sounds good. thanks!

spatialdata/_core/_spatial_query.py
```diff
@@ -85,3 +89,129 @@ def _bounding_box_query_points_dict(
...
+
+
+def _bounding_box_query_image(
+    image: Union[MultiscaleSpatialImage, SpatialImage], request: BoundingBoxRequest
+) -> Union[MultiscaleSpatialImage, SpatialImage]:
+    """Perform a spatial bounding box query on an Image or Labels element.
+
+    Parameters
+    ----------
+    image : Union[MultiscaleSpatialImage, SpatialImage]
```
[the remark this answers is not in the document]
kevinyamauchi: Okay, I will remove them.

spatialdata/_core/_spatial_query.py
kevinyamauchi: Should it be like this then?

```python
Returns
-------
The image contained within the specified bounding box.
```

giovp: Yes I think this is what we use

spatialdata/_core/_spatial_query.py
LucaMarconato: I agree with you

spatialdata/_core/_spatial_query.py
LucaMarconato: I am fine with that, maybe let's open an issue re refactoring them with that and once that pr is approved we can start using the new syntax.

LucaMarconato:
@kevinyamauchi looks very good to me! I think there should be tests for labels since their axes is different from the image ones and since 3d labels are available. After that it's ready to merge!

kevinyamauchi:
> @kevinyamauchi looks very good to me! I think there should be tests for labels since their axes is different from the image ones and since 3d labels are available. After that it's ready to merge!

Good point, @LucaMarconato . I can add this today.

kevinyamauchi:
@giovp , @LucaMarconato , I have added the requested tests and fixed the docstrings!

giovp:
thanks a lot @kevinyamauchi ! now that this is merged, could you by any chance take a look at the new points specs with dask dataframes https://github.com/scverse/spatialdata/pull/113 ? Thanks in advance.

LucaMarconato:
@giovp I have fixed it (it was a minor thing)

PR #113, opened 2023-01-24
update points model

giovp:
This PR aims at implementing `Points` element as `dask.dataframe.DataFrame`


LucaMarconato:
Looks great so far thanks a lot!

giovp:
@kevinyamauchi only tests that are still not passing are the bounding boxes, with error e.g.
```
>           raise AttributeError("'DataFrame' object has no attribute %r" % key)
E           AttributeError: 'DataFrame' object has no attribute 'filter'
```

How do you think is best to proceed? I could see we could merge this in your PR first?

LucaMarconato:
Look very good to me! I have added only two comments.

LucaMarconato:
@giovp I have fixed the failing tests (the tests where expecting to find `pyarrow.Table` objects so I have rewritten them using the `DaskDataframe` APIs.

giovp:
thanks a lot @LucaMarconato ! Unfortunately, I can't find the comments in the review.. do you ming adding them again>?

giovp:
@LucaMarconato thanks a lot for the comments. I think I agree on all and implement them but maybe only one thing I'd like to stress out

>  I can see in the future it's value for gene expression data, especially because we could use it to compute an index on the gene expression.

I don't think its' the future but really the present. The single most important use case about the `aggregate` method is to count genes under each segmentation mask/location and the `feature_key` is a must in that case. That is what builds eventually the gexp matrix and a crucial step in the analysis.

LucaMarconato:
I think that for canonical analyses it is true, but for more general operations that the user may need this may not be correct. For instance the user may define some landmark points to use for alignments.

LucaMarconato (approved):

giovp:
ok if tests pass it's ready to be merged

LucaMarconato:
Super!

Issue #125, opened 2023-02-02
Currently considering only scale in the transformations for multiscales


LucaMarconato:
As shown in the [screenshot here](https://github.com/scverse/spatialdata/issues/39#issue-1415482443), for multiscale images and multiscale labels we two sets of transformations. One at the top level, following the new NGFF specs, and one just between levels of the multiscale following the old specs.

Currently I am not saving the transformations in this format, I will have to adapt the IO to follow the NGFF specs (once they are approved), as described in this other issue: https://github.com/scverse/spatialdata/issues/114

But also in the code I am currently expecting the translation to always be absent. I believe that there is no need for that translation now that we have the new more powerful specs, so we'll have to discuss in the NGFF repo. For the moment I open this issue and I will ignore that.

Issue #39, opened 2022-10-19
Influence on the order of axes when declaring different affine transformations


LucaMarconato:
In short: specifying an affine transformation manually is not intuitive and requires an understanding of the underlining implicit and explicit coordinate spaces. Here I describe the problem and I propose one solution to simplify the exposure to this complexity to the user.

I tag @ivirshup, we started discussing this in the last spatial data call and you asked me to write an issue about it. I would also like to know the opinion of @bogovicj, thanks 😊 

To illustrate this issue I start with a real example. Here is the `.zattr` for some 2D Points (left) and for some 2D Labels (right).
![Untitled-1](https://user-images.githubusercontent.com/2664412/196782580-42192537-390b-4dec-8401-ebe3891d9894.jpg)

Let's look for example at the affine transformation which maps the Points (left) to from the implicit (xy) to the target (cyx) coordinate space.
This affine transformation is the following matrix (the NGFF specs says to drop the last row when saving). I pad the matrix with the coordinate spaces for extra clarify.

```
  x y <- source space
c 0 0 0
y 0 1 0
x 1 0 0
  0 0 1
^
target space
```

As you can see this matrix is different from the one specified for Labels, since the implicit coordinate space for labels is yx and not xy like for point.
So far so good. The problem is when the user wants to specify an affine transformation explicitly things could get confusing. For instance if the user wants to translate the **points** by the vector (x, y) = (a, b), the affine transformation would be
```
  x y
c 0 0 0
y 0 1 b
x 1 0 a
  0 0 1
```
If the user wanted to translate the image, the affine transformation would be this one:
```
  y x
c 0 0 0
y 1 0 b
x 0 1 a
  0 0 1
```
While if the user wanted to translate the points between the coordinate systems (x, y) -> (x, y), the affine transformation would simply be
```
  x y
x 1 0 a
y 0 1 b
  0 0 1
```
This last form of affine transformation is the most common one (for instance ihmo it is the most used in math classes and among softwares, for instance napari uses it or [Apple](https://developer.apple.com/documentation/corefoundation/cgaffinetransform) (minus a transposition)).

What I propose is that by default the user will not have to specify the final form of the affine transformation that is written to disk, but that the user will only write an affine transformation of the form (for the 2D case):
```
  x y
x 1 0 a
y 0 1 b
  0 0 1
```
and the analogous for the 3D case.
[...]

PR #127, opened 2023-02-05
enhance + bugfix of images and labels elements

### LINKED ISSUE #59: write to zarr fails when `SpatialImage` is created from a `xarray.DataArray`
When the `SpatialImage` is created from a `np.ndarray` all is good, when is created from a `xarray.DataArray` an error occurs (see traceback).
I have added tests for this. To reproduce uncomment the parts of `_get_images()` and `_get_labels()` (from `tests/conftest.py`) that start with `out['....._xarray'] = .....`.

```
tests/_io/test_readwrite.py:14 (TestReadWrite.test_images)
self = <tests._io.test_readwrite.TestReadWrite object at 0x7f9f9490edd0>
tmp_path = PosixPath('/private/tmp/pytest-of-macbook/pytest-74/test_images0')
images = SpatialData object with:
└── Images
│     ├── 'image2d_numpy': SpatialImage[cyx] (3, 64, 64)
│     ├── 'image2d_multis...3, 64, 64)
│     └── 'image2d_multiscale_numpy_xarray': MultiscaleSpatialImage[cyx] (3, 64, 64), (3, 32, 32), (3, 8, 8)

    def test_images(self, tmp_path: str, images: SpatialData) -> None:
        """Test read/write."""
        tmpdir = Path(tmp_path) / "tmp.zarr"
>       images.write(tmpdir)

tests/_io/test_readwrite.py:18: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
spatialdata/_core/_spatialdata.py:147: in write
    write_image(
spatialdata/_io/write.py:95: in write_image
    write_image_ngff(
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/ome_zarr/writer.py:455: in write_image
    _write_dask_image(
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/ome_zarr/writer.py:565: in _write_dask_image
    da.compute(*delayed)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/base.py:600: in compute
    results = schedule(dsk, keys, **kwargs)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/threaded.py:89: in get
    results = get_async(
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/local.py:511: in get_async
    raise_exception(exc, tb)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/local.py:319: in reraise
    raise exc
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/local.py:224: in execute_task
    result = _execute_task(task, data)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/core.py:119: in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/array/core.py:4366: in store_chunk
    return load_store_chunk(x, out, index, lock, return_stored, False)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/dask/array/core.py:4348: in load_store_chunk
    out[index] = x
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:1373: in __setitem__
    self.set_basic_selection(pure_selection, value, fields=fields)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:1468: in set_basic_selection
    return self._set_basic_selection_nd(selection, value, fields=fields)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:1772: in _set_basic_selection_nd
    self._set_selection(indexer, value, fields=fields)
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:1844: in _set_selection
    self._chunk_setitems(lchunk_coords, lchunk_selection, chunk_values,
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:2045: in _chunk_setitems
    to_store = {k: self._encode_chunk(v) for k, v in cdatas.items()}
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:2045: in <dictcomp>
    to_store = {k: self._encode_chunk(v) for k, v in cdatas.items()}
../../../../miniconda3/envs/ome/lib/python3.10/site-packages/zarr/core.py:2222: in _encode_chunk
    if ensure_ndarray_like(chunk).dtype == object:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

buf = <xarray.DataArray (c: 3, y: 64, x: 64)>
array([[[-1.45060918e+00,  5.93752346e-01, -3.55374641e-02, ...,
         -5.1...053254e+00, ...,
         -2.45753512e-01,  1.03229240e+00, -8.63332273e-01]]])
Dimensions without coordinates: c, y, x

    def ensure_ndarray_like(buf) -> NDArrayLike:
        """Convenience function to coerce `buf` to ndarray-like array.
    
        Parameters
        ----------
        buf : ndarray-like, array-like, or bytes-like
            A numpy array like object such as numpy.ndarray, cupy.ndarray, or
            any object exporting a buffer interface.
    
        Returns
        -------
        arr : NDArrayLike
            A ndarray-like, sharing memory with `buf`.
    
        Notes
        -----
        This function will not create a copy under any circumstances, it is guaranteed to
        return a view on memory exported by `buf`.
        """
    
        if not is_ndarray_like(buf):
            if isinstance(buf, array.array) and buf.typecode in "cu":
                # Guard condition, do not support array.array with unicode type, this is
                # problematic because numpy does not support it on all platforms. Also do not
                # support char as it was removed in Python 3.
                raise TypeError("array.array with char or unicode type is not supported")
            else:
                # N.B., first take a memoryview to make sure that we subsequently create a
                # numpy array from a memory buffer with no copy
>               mem = memoryview(buf)
E               TypeError: memoryview: a bytes-like object is required, not 'DataArray'

../../../../miniconda3/envs/ome/lib/python3.10/site-packages/numcodecs/compat.py:42: TypeError
```

### LINKED ISSUE #60: bug in read when omero metadata is present
Converting the morphology_mip of the xenium dataset (see spatialdata-sandbox) leads to the following `.zattrs`.

```json
{
  "multiscales" : [ {
    "metadata" : {
      "method" : "loci.common.image.SimpleImageScaler",
      "version" : "Bio-Formats 6.10.1"
    },
    "axes" : [ {
      "name" : "t",
      "type" : "time"
    }, {
      "name" : "c",
      "type" : "channel"
    }, {
      "name" : "z",
      "type" : "space"
    }, {
      "unit" : "micrometer",
      "name" : "y",
      "type" : "space"
    }, {
      "unit" : "micrometer",
      "name" : "x",
      "type" : "space"
    } ],
    "name" : "Image0",
    "datasets" : [ {
      "path" : "0",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 0.2125, 0.2125 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "1",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 0.425, 0.425 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "2",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 0.85, 0.85 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "3",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 1.7, 1.7 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "4",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 3.4, 3.4 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "5",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 6.8, 6.8 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "6",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 13.6, 13.6 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "7",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 27.2, 27.2 ],
        "type" : "scale"
      } ]
    }, {
      "path" : "8",
      "coordinateTransformations" : [ {
        "scale" : [ 1.0, 1.0, 1.0, 54.4, 54.4 ],
        "type" : "scale"
      } ]
    } ],
    "version" : "0.4"
  } ],
  "omero" : {
    "channels" : [ {
      "color" : "FFFFFF",
      "coefficient" : 1,
      "active" : true,
      "label" : "DAPI",
      "window" : {
        "min" : 0.0,
        "max" : 5658.0,
        "start" : 0.0,
        "end" : 5658.0
      },
      "family" : "linear",
      "inverted" : false
    } ],
    "rdefs" : {
      "defaultT" : 0,
      "model" : "greyscale",
      "defaultZ" : 0
    }
  }
}
```

The function `_read_multiscale()` from `spatialdata/_io/read..py` has the following values for `node.metadata`

```python console
In[3]: node.metadata
Out[3]: 
{'axes': [{'name': 't', 'type': 'time'},
  {'name': 'c', 'type': 'channel'},
  {'name': 'z', 'type': 'space'},
  {'unit': 'micrometer', 'name': 'y', 'type': 'space'},
  {'unit': 'micrometer', 'name': 'x', 'type': 'space'}],
 'name': ['DAPI'],
 'coordinateTransformations': [[{'scale': [1.0, 1.0, 1.0, 0.2125, 0.2125],
    'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 0.425, 0.425], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 0.85, 0.85], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 1.7, 1.7], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 3.4, 3.4], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 6.8, 6.8], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 13.6, 13.6], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 27.2, 27.2], 'type': 'scale'}],
  [{'scale': [1.0, 1.0, 1.0, 54.4, 54.4], 'type': 'scale'}]],
 'visible': [True],
 'contrast_limits': [[0.0, 5658.0]],
 'colormap': [[[0, 0, 0], [1, 1, 1]]]}
```

Note how the name is `['DAPI']` instead of `Image0`. Having a list instead of a string triggers a bug downstream (see traceback):
```python console
/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/anndata/experimental/pytorch/_annloader.py:18: UserWarning: Сould not load pytorch.
  warnings.warn("Сould not load pytorch.")
  File "/Users/macbook/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/222.4459.20/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevd.py", line 1496, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/Users/macbook/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/222.4459.20/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/macbook/embl/projects/basel/spatialdata-io/src/spatialdata_io/readers/read_xenium.py", line 189, in <module>
    sdata = SpatialData.read("./data.zarr/")
  File "/Users/macbook/embl/projects/basel/spatialdata/spatialdata/_core/_spatialdata.py", line 242, in read
    sdata = read_zarr(file_path)
  File "/Users/macbook/embl/projects/basel/spatialdata/spatialdata/_io/read.py", line 82, in read_zarr
    images[k] = _read_multiscale(node, fmt)
  File "/Users/macbook/embl/projects/basel/spatialdata/spatialdata/_io/read.py", line 39, in _read_multiscale
    return MultiscaleSpatialImage.from_dict(multiscale_image)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/datatree/datatree.py", line 793, in from_dict
    new_node = cls(name=node_name, data=data)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/multiscale_spatial_image/multiscale_spatial_image.py", line 32, in __init__
    super().__init__(data=data, name=name, parent=parent, children=children)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/datatree/datatree.py", line 326, in __init__
    ds = _coerce_to_dataset(data)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/datatree/datatree.py", line 68, in _coerce_to_dataset
    ds = data.to_dataset()
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/xarray/core/dataarray.py", line 620, in to_dataset
    result = self._to_dataset_whole(name)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/xarray/core/dataarray.py", line 568, in _to_dataset_whole
    if name in self.coords:
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/xarray/core/coordinates.py", line 74, in __contains__
    return key in self._names
TypeError: unhashable type: 'list'
```

We need to check more datasets (with more than one channel) to see where exactly the bug lies (could also be in `multiscale-spatial-image`). For the moment I am using a workaround to deal with this edge case.

Also  'visible', 'contrast_limits' and 'colormap' are wrapped in lists, so probably this bug could manifest also when dealing with them.

### LINKED ISSUE #84: 3D image order axes constraint breaks io
Discovered by @berombau when converting the MERSCOPE dataset.

The function `to_multiscale` from `spatial-image`, requires the `zyxc` order. At the moment this break io. I think @giovp is already aware of the problem.

### LINKED ISSUE #115: Refinements to the schema for `MultiscaleSpatialImage`
Here is how a `MultiscaleSpatialImage` looks like
```
DataTree('None', parent=None)
├── DataTree('scale0')
│       Dimensions:             (c: 3, y: 64, x: 64)
│       Dimensions without coordinates: c, y, x
│       Data variables:
│           image2d_multiscale  (c, y, x) float64 dask.array<chunksize=(3, 64, 64), meta=np.ndarray>
└── DataTree('scale1')
        Dimensions:             (c: 3, y: 32, x: 32)
        Dimensions without coordinates: c, y, x
        Data variables:
            image2d_multiscale  (c, y, x) float64 dask.array<chunksize=(3, 32, 32), meta=np.ndarray>
```
and this is an example of the second level of the multiscale
```
data['scale1']['image2d_multiscale']
Out[4]: 
<xarray.DataArray 'image2d_multiscale' (c: 3, y: 32, x: 32)>
dask.array<mean_agg-aggregate, shape=(3, 32, 32), dtype=float64, chunksize=(3, 32, 32), chunktype=numpy.ndarray>
Dimensions without coordinates: c, y, x
Attributes:
    transform:  Sequence \n    Scale (y, x)\n        [2. 2.]\n    Identity 
```

We need to add the following constraints to the validator:
1) the nodes of the tree are only 'scale0', 'scale1', etc.
2) each node has only one data variable (`image2d_multiscale` in the example above). The name is irrelevant
3) each data variable has the transformation in `.attrs['transform']`. That is, the transformation is not here: `multiscale_data.attrs['transform']`; it's not here: `multiscale_data.attrs['scale0'].attrs['transform']`, but it's here: `multiscale_data.attrs['scale0']['image2d_multiscale'].attrs['transform']`.

Right now this type of validation is scattered around the repo. In particular this is done in `write.py`, in the function `_iter_multiscale()`, here reported
```
def _iter_multiscale(
    data: MultiscaleSpatialImage,
    attr: str,
    key: Optional[str] = None,
) -> list[Any]:
    # TODO: put this check also in the validator for raster multiscales
    name = None
    for i in data.keys():
        variables = list(data[i].variables)
        if len(variables) != 1:
            raise ValueError("MultiscaleSpatialImage must have exactly one variable (the variable name is arbitrary)")
        if name is not None:
            if name != variables[0]:
                raise ValueError("MultiscaleSpatialImage must have the same variable name across all levels")
        name = variables[0]
    if key is None:
        return [getattr(data[i][name], attr) for i in data.keys()]
    else:
        return [getattr(data[i][name], attr).get(key) for i in data.keys()]
```
and it is done in `get_transform()` from `core_utils.py`:
```
    d = dict(e['scale0'])
    assert len(d) == 1
    xdata = d.values().__iter__().__next__()
    t = get_transform(xdata)
```

giovp:
This PR does the following:

### Add support for coordinates in `(multiscale)spatial-image`.

without coordinates, it doesn't make much sense to use an `xarray.DataTree` cause many methods do not work across scales. The implementation is the following:
- create a `(multiscale)spatial-image` from the parser
- re-calculate coordinates based on this function

https://github.com/scverse/spatialdata/blob/f65151e9a237e6172d25525a2b17f5a3aa40faa7/spatialdata/_core/core_utils.py#L307-L310

- during IO, coordinates are not saved but re-calculated at input. This makes the coordinates consistent (also across downscaling methods from `spatial-image`). 

One important thing to take into accounts, is that the coordinats always refer to the implicit (pixel) coordinate systems. They are nonetheless useful as they unlock the power of xarray operations on data(trees).

### Add support for 3d Images

This was commented out due to mixing `spatial-image` class. This was addressed here https://github.com/spatial-image/spatial-image/pull/16 . This class is also supported in multiscale case.

### Enhance schema for raster elements

In particular, it now performs correct validation of scales and re-implements the validation for `"transform"` being present in the right place in `attrs`.

It closes the issues listed here:

- close #59 
- close #84 
- close #115 
- close #60 

### Open questions
One thing I'd like to address here, or potentially in separate PR, is the use of the `name` as attribute in `(multiscale)spatial-image`. It's not a specific attribute of spatial-image but of xarray dataarray or datatree. For instance, in the case of `datatree` it's used to access the `DataArray` and the desired node. E.g.

```python
img = Image2DModel(..., scale_factors=[2, 4])
img["scale0"]["image"] # the name in this case is `image`
>>> DataArray ....
```

Right now, we are not very consistent about this through out repo. In particular, the transformations i think expect the fact that `name="image"` yet in many other parts of the repo that is not the case. Furthermore, if the user passes the name, the name is not saved (yet it is used in creating the spatial-image object). 

It's also unclear how this interplay with the key of the image element in spatial data. e.g.

```python
img = Image2DModel(..., scale_factors=[2, 4], name="myimage")
sdata =SpatialData(images={"myrealimage":img}) 
sdata.images["myrealimage"]["scale0"]["image"] # the name in this case is still `image`
>>> DataArray ....
```

I don't have any preference on how to handle this, but we should be consistent, so I suggest either of these two implementations:
- either make name arbitrary, but then save and read accordingly if specified by user. Also transform should be agnostic to that.
- do not allow to pass name, and always use the name `"image"` (default) across repo (and so we can avoid to save/read).


LucaMarconato:
I think it's a good idea to keep coords. But one important thing, we need to understand/decide what's the relationship with transformations. Scale and translations play around well, but when rotations are involved things get dirty.

I would start with this:
1) the coordinates we save are the one for the local coordinate system. We gain no practical advantage with `SpatialImage` in this way, but we can use, as you said, methods that work across scales in `MultiscaleSpatialImage`.

We could extend to this:
2) the implicit coordinate system doesn't need to start from 0 and have units equal to pixels. This is still supported by the ngff specs (and would address this issue: https://github.com/scverse/spatialdata/issues/125). Implications of this are:
*) when cropping or scaling sometimes we can update the xarray object instead of adding a transformation (unless the transformations are more complex like general affine, we can detect this)
*) workflow more intuitive to xarray users that already encode translations and scale into xarray objects, but at the same time the information is sometimes saved as rotations and others as coordinates (see cropping example above), 
*) intrinsic coordinate systems become more powerful, but extrinsic coordinate systems remain important for aligning different objects, rotations and multiple transformations.

LucaMarconato:
Some consideration; just found an implication on this pr. 
In the nanostring cosmx dataset I need to flip each image and labels on the y-axis. I could do it with an affine transformation, or I could do it by transposing the data with dask array (it's a lazy computation so it's fine). But another way for xarray users is to flip only the coordinates ([see stackoverflow](https://stackoverflow.com/questions/54677161/xarray-reverse-an-array-along-one-coordinate)). 
Without this pr the last method can't be applied because coordinates are stripped. With this pr we need to adjust processing code so that it considers the coordinates. I think that if we apply the NGFF transformation to the xarray coordinates things should work.

LucaMarconato:
Also we could decide to avoid saving the coordinates directly but converting them to NGFF into the "dataset transformations" (the translation + scale pair present at each level in the multiscale). The NGFF specs says to apply the new transformations after applying the dataset transformations, so this would be equivalent to apply the NGFF transformations to the xarray coordinates.

In both cases (saving the coordinates directly or converting them to the NGFF dataset transformations), we can only support coordinates that are equivalent to a scale and a translation. Non-linear coordinate displacements (like 0, 1, 2, 3, 10) would break the interplay with the NGFF transformations.

giovp:
> In the nanostring cosmx dataset I need to flip each image and labels on the y-axis. I could do it with an affine transformation, or I could do it by transposing the data with dask array (it's a lazy computation so it's fine). But another way for xarray users is to flip only the coordinates ([see stackoverflow](https://stackoverflow.com/questions/54677161/xarray-reverse-an-array-along-one-coordinate)).

what do you need to flip them for?

> Without this pr the last method can't be applied because coordinates are stripped. With this pr we need to adjust processing code so that it considers the coordinates (basically we apply the NGFF transformations to the coordinates when aligning things).

If by adjust you mean adjust IO then for sure we'll have to adapt

> In both cases (saving the coordinates directly or converting them to the NGFF dataset transformations), we can only support coordinates that are equivalent to a scale and a translation. Non-linear coordinate displacements (like 0, 1, 2, 3, 10) would break the interplay with the NGFF transformations.

that's a good point. ideas could be:
-  only save the coordinate of the highest resolution pyramid, and infer the others at IO according to "scale" (I think only scale relevant for our use of multiscale image). I think this might break the roundtrip as there might be numerical differences popping up, so we'd have to maybe be less restrictive re IO ?
- save the coordinates separately for each group. Probably best option but need to see how it interplays with ome_zarr reader/writers
- simply use the `to_zarr` method of xarray and store in metadata any relevant ome-ngff metadata. I think this would be the quickest option but then we'd really drop any dependency with `ome-zarr-py` (except for format) which could be positive or negative in different ways.

LucaMarconato:
Answering the first 2 points

> what do you need to flip them for? 

The tiles where mapped to the global space in the wrong order. I flipped the y axis of each tile. Another way (better) would have been to adjust the mapping to the global space. We can do this (I write a TODO in the code)

> If by adjust you mean adjust IO then for sure we'll have to adapt

io is not necessary, we need to make the processing methods (and transformations, since the processing methods can be called both on the intrinsic space and any transformed space) aware of the coordinates since we can't work anymore with pure pixel coordinates, but now only with xarray coordinates

LucaMarconato:
Regarding the third part, currently the first option is what is implemented. Referring to [the screenshot in this discussion](https://github.com/scverse/spatialdata/issues/39), I am basically assuming that the transformation corresponding to `"path": "0"` in "Coordinate transformations for the multiscale representation" is translation [0, 0] and scale [1, 1].

We could save the coordinates in that slot by deriving the NGFF transformation that produces the corresponding xarray coordinates. The ones for the next multiscales can be derived from the first (or can also be computed from xarray). I save all the multiscale transformations to the file but when I read I actually load just the top one and re-derive the others: as you pointed out, it should be the same up to some numerical precision errors. 

So the interpretation of the xarray coordinates is that they describe the intrinsic coordinate system (we would never use teh pixel space anymore).

giovp:
> So the interpretation of the xarray coordinates is that they describe the intrinsic coordinate system (we would never use teh pixel space anymore), and the new transformation classes would always operate on the xarray coordinates.

👍

LucaMarconato:
> The tiles where mapped to the global space in the wrong order. I flipped the y axis of each tile. Another way (better) would have been to adjust the mapping to the global space. We can do this (I write a TODO in the code)

I checked, actually flipping the images is correct because the points are aligned with the flipped images.  If change the global positioning of the fov the points become wrong.

giovp:
codecov in this repo is cursed.....

giovp:
@LucaMarconato since `multiscale-spatial-image` there is now an explicit check for scale factors in building the multiscale (I removed the previous check here).

This check is now catching bugs here
https://github.com/scverse/spatialdata/blob/238fb8ca606d55625722521d86288abd3178c1a3/spatialdata/_core/_transform_elements.py#L172-L187

I find reading those lines quite difficult, in particular there are things like
```
factors = shape0 / shape
factors - min(factors)
```

that I don't get whether they are a bug or a leftover. Do you mind taking a look? I fixed 1 but there are still 3 failing

```
FAILED tests/_core/test_transformations_on_elements.py::test_transform_labels_spatial_multiscale_spatial_image - ValueError: Scale factor 10 is incompatible with image shape (10, 148, 148) along dimension `z`.
FAILED tests/_core/test_transformations_on_elements.py::test_transform_elements_and_entire_spatial_data_object[full] - ValueError: Scale factor 10 is incompatible with image shape (10, 64, 128) along dimension `z`.
FAILED tests/_core/test_transformations_on_elements.py::test_transform_elements_and_entire_spatial_data_object[labels] - ValueError: Scale factor 10 is incompatible with image shape (10, 64, 128) along dimension `z`.
```

giovp:
@scverse/spatialdata please refer to the header comment for the description of this PR

https://github.com/scverse/spatialdata/pull/127#issue-1571604280

kevinyamauchi (commented):
Thanks, @giovp ! The code changes look good to me. I am not sure about the other test, but the failing bounding box query tests suggest to me that something has changed in the indexing behavior. The bounding boxes are taken using `image.sel` (see [here](https://github.com/scverse/spatialdata/blob/30897ce960ac9dd80395d44101601e7b005c85c4/spatialdata/_core/_spatial_query.py#L134)).

LucaMarconato:
> I don't have any preference on how to handle this, but we should be consistent, so I suggest either of these two implementations:
> * either make name arbitrary, but then save and read accordingly if specified by user. Also transform should be agnostic to that.
> * do not allow to pass name, and always use the name "image" (default) across repo (and so we can avoid to save/read).

I am not sure which one I prefer, but I would start with the first approach, by making the transformations agnostic to that. I'll change the behavior.

giovp:
> I am not sure which one I prefer, but I would start with the first approach, by making the transformations agnostic to that. I'll change the behavior.

yeah I think when I try that only transformations didn't work

LucaMarconato:
> @LucaMarconato since `multiscale-spatial-image` there is now an explicit check for scale factors in building the multiscale (I removed the previous check here).
> 
> This check is now catching bugs here
> 
> https://github.com/scverse/spatialdata/blob/238fb8ca606d55625722521d86288abd3178c1a3/spatialdata/_core/_transform_elements.py#L172-L187
> 
> I find reading those lines quite difficult, in particular there are things like
> 
> ```python
> factors = shape0 / shape
> factors - min(factors)
> ```
> 
> that I don't get whether they are a bug or a leftover. In general, there are also `print` that should be `logg.warning` and maybe a bit more comments on reasoning would be helpful.
> 
> Do you mind taking a look? I fixed 1 but there are still 3 failing
> 
> ```
> FAILED tests/_core/test_transformations_on_elements.py::test_transform_labels_spatial_multiscale_spatial_image - ValueError: Scale factor 10 is incompatible with image shape (10, 148, 148) along dimension `z`.
> FAILED tests/_core/test_transformations_on_elements.py::test_transform_elements_and_entire_spatial_data_object[full] - ValueError: Scale factor 10 is incompatible with image shape (10, 64, 128) along dimension `z`.
> FAILED tests/_core/test_transformations_on_elements.py::test_transform_elements_and_entire_spatial_data_object[labels] - ValueError: Scale factor 10 is incompatible with image shape (10, 64, 128) along dimension `z`.
> ```
> 
> meanwhile I'll go on and work on IO for channels with omero metadata

Yeah the code was very weird, I removed it altogether. Now I do the transformation for each element of the multiscale and then I assemble the `MultiscaleSpatialImage` object back together.

LucaMarconato:
@giovp I reviewed the code and fixed the tests. There is only a comment that could require changes (the one about the name of the datatree nodes), and a tiny pre-commit fail that I would maybe just silence.

giovp:
@LucaMarconato I'll merge this, now the default name for the dataarray in the datatree is not what we decided as `im` but whatever defaults on spatial-image which is `"image"`

PR #132, opened 2023-02-13
unify polygons and shapes elements

giovp:
## New `Shapes` model

This PR unifies `polygons` and `shape` elements in a single region element: `Shapes`.

### Motivation
As of now, the `Shapes` element was stored as an AnnData, with a bunch of metadata to keep track of the type of shapes and size, and with the centroid coordinates store in in `adata.obsm["spatial"]`. Thinking about it more, it was very much unnecessary:
- since it's a region element, only minimal amount of information should be stored there, as any annotations would go into Tables -> because of this, the AnnData format is unneccesary
- it contained a lot of unnecessary metadata for determining which type of shapes are stored (circle, square only so far).
- it could be simplified by representing it as a geopandas dataframe.

### Solution
Since a `geopandas.GeoDataFrame` is probably the better way to represent `Circles` and `Squares` shapes, instead of `AnnData`, it naturally follows that polygons and shapes should be grouped together. Benefits are the following:
- unified `Shapes` element (less complexity)
- less metadata, offload to shapely handling of different geometry types
- easier access to aggregation functions for `Circles` from geopandas. 

e.g. example for counting points under circles regions @ivirshup 
```python
import geopandas as gpd
from spatialdata._core.models import ShapesModel
from spatialdata._core.models import PointsModel

circles = ShapesModel.parse(np.array([[10, 2], [5, 2]]), geometry=0, radius=[2, 3])
points = PointsModel.parse(
    np.array([[11, 3], [5, 3]]), annotation=pd.DataFrame({"a": [0, 1]})
)
spot = circles["geometry"].buffer(distance=circles["radius"])[0]
gpd.points_from_xy(**points[["x", "y"]].compute()).within(spot)
```

### Summary
Changes are minimal:
- The `Polygons` element (and models, IO, etc.) it's now called `Shapes` and it also contains `Circles`
- `Circles` are represented as a `geopandas.GeoDataFrame` of `shapely.Point` with additional column for `radius`
- The old `Shapes` element is removed
- No additional metadata is introduced, only change from IO side is read/write for `radius` column in the case of `Circles`


giovp:
tests in transforms are failing but I'd wait for #100 to be merged first before fixing, as there will be many changes to be made as well.

LucaMarconato:
@giovp all tests are passing now. I did just fixes, no real change. Anyway if you want to review I suggest to diff the two following commits:
70d3f16 (only merged the branches and ran pre-commits)
e1745f5 (latest commit, all tests passing).

spatialdata/_io/format.py
```diff
@@ -5,11 +5,11 @@
 from pandas.api.types import is_categorical_dtype
 from shapely import GeometryType
 
-from spatialdata._core.models import PointsModel, PolygonsModel, ShapesModel
+from spatialdata._core.models import PointsModel, ShapesModel
 
 CoordinateTransform_t = list[dict[str, Any]]
 
-Polygon_s = PolygonsModel()
+Shapes_s = ShapesModel()
```
kevinyamauchi, line 12: I think `ShapesModel` is duplicated here (see line below)

spatialdata/_core/models.py
```diff
@@ -70,7 +70,7 @@
     "Labels3DModel",
     "Image2DModel",
     "Image3DModel",
-    "PolygonsModel",
+    "ShapesModel",
```
kevinyamauchi, line 73: I think `ShapesModel` is duplicated here (see line below)

spatialdata/_core/models.py
```diff
@@ -306,32 +305,75 @@ def validate(cls, data: GeoDataFrame) -> None:
...
+                - If :class:`geopandas.GeoDataFrame`, it's validated.
+
+            A `radius` array can also be passed to store the radius of the `Circles`.
+
+        geometry
+            Geometry type of the shapes. The following geometries are supported:
+
+                - 0: `Circles`
+                - 3: `Polygon`
+                - 6: `MultiPolygon`
```
kevinyamauchi, line 343: what are the 0, 3, and 6 here?

kevinyamauchi (approved):
This LGTM! I think it makes sense to unify the Circles and polygons here.  I left a couple of minor comments below.

spatialdata/_core/models.py
giovp: thse are defined by `GeometryType` enum by shapely and are used to set the geometry type in the parser.

giovp:
@LucaMarconato thanks a lot for fixing tests!!! I think we can go on and merge this, one comment:
- I saw that pyarrow was re-added, do we need it? Think it's an artefact right?

LucaMarconato:
Thanks for the PR 😊 we need it because it's used in the PR fixing the __repr__ for SpatialData (also, I noticed that we need it in SpatialData io for thr cosmx reader)

giovp:
> Thanks for the PR 😊 we need it because it's used in the PR fixing the `__repr__` for SpatialData (also, I noticed that we need it in SpatialData io for thr cosmx reader)

wait, if it's for the cosmx reader then it should be in spatialdata io

LucaMarconato:
@giovp yes for the moment better to keep it. Dask will allow to get the length of the dataframe soon (https://github.com/dask/dask/issues/9973), at that point we can remove the dependency.

ivirshup:
@LucaMarconato I don't think that dask is going to get this soon. I don't see an active PR for it at the moment, and it's been open for a few years now.

PR #134, opened 2023-02-13
init tests for format module

giovp:


giovp:
@LucaMarconato I would merge it as is and add more tests as I progress with fixing IO for raster (with coordinates) and shapes (in #132 )

PR #48, opened 2022-11-02
legacy code to port

LucaMarconato:
Merge postponed.
Before merging (an updated version of) this pr, the following will be isolated in a separate pr, improved, and merged.
- [x] coordinate systems
- [x] coordinate transformations
- [x] io branch from @giovp 

## About this pr.
This pull request implements a first working version of coordinate systems, transformations, as long as all what spatialdata needs to perform basic io operatations with its elements. Spatialdata-sandbox and napari_spatialdata support this branch. They will also need to be adapted to the new improved code described in the incremental prs listed above.

PR #136, opened 2023-02-13
Old legacy code

LucaMarconato:
This PR ports to the new codebase those missing features that stayed for long sitting in this old pr: https://github.com/scverse/spatialdata/pull/48. After this PR is merged, the old one can be deleted.

In details:
- improved filter_by_coordinate_system() in _spatialdata.py
- concatenation of tables and of `SpatialData` objects
-  the cli support is back (`__main__.py`)

Also there are some tiny helper functions or functions moved from notebooks, in particular:
- `get_transformation_between_landmarks()`, which returns a similarity (anisotropic affine) function between two sets of landmark points
- `align_elements_using_landmarks()`, which prepends the transformation to the previous one present in an element and optionally maps this to a new coordinate system
- `get_table_metadata()`, which return `region`, `region_key`, `instance_key`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
       hooks:
           - id: isort
     - repo: https://github.com/pre-commit/mirrors-mypy
-      rev: v0.991
+      rev: v1.0.0
       hooks:
           - id: mypy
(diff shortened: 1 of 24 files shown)


LucaMarconato:
The review for this branch has been made in https://github.com/scverse/spatialdata/pull/138. I have addressed all the review points, so now merging to main.

Issue #137, opened 2023-02-14
[Possible bug] ContainsGroupError when trying to overwrite SpatialData object (overwriting a backed sdata/element)


melonora:
I am currently retrying writing tiled images and encounter a ```ContainsGroupError``` in version 0.0.1.dev0. 

Specifically running something similar to this twice with the first time no overwrite results in the error:
```
sdata = sd.SpatialData(images=images)
sdata.write(<Path>, storage_options={"chunks": (1,imgs.shape[1]/2,imgs.shape[2]/2)}, overwrite=True)
```

Is this expected behaviour?

LucaMarconato:
Hi, thanks for reporting. The errors is due to the fact that when you save a `SpatialData` object to disk, and when it contains backed elements (Images, Labels, Points), it doesn't load the data in-memory by default but it reads it lazily from disk.

More generally, for the same reason, this happens even when you want to add a single element to an existing `SpatialData` object with `add_image()`, `add_labels()` and `add_points()`.

This is shown in the following example. All works with Shapes (Polygons and Tables), but the equivalent code breaks with Points.

```
from spatialdata import SpatialData
from tests.conftest import _get_shapes, _get_points
import tempfile
import os

with tempfile.TemporaryDirectory() as tmpdir:
    f = os.path.join(tmpdir, "data.zarr")
    sdata = SpatialData(shapes=_get_shapes())
    print(sdata)
    sdata.write(f)
    e = sdata.shapes['shapes_0']
    sdata.add_shapes('new_shapes', e, overwrite=True)

    # ok
    sdata.add_shapes('new_shapes', e, overwrite=True)
    # ok, beacause shapes are not backed
    sdata.add_shapes('new_shapes', sdata.shapes['new_shapes'], overwrite=True)

with tempfile.TemporaryDirectory() as tmpdir:
    f = os.path.join(tmpdir, "data.zarr")
    sdata = SpatialData(points=_get_points())
    print(sdata)
    sdata.write(f)
    e = sdata.points['points_0']
    sdata.add_points('new_points', e, overwrite=True)

    # ok
    sdata.add_points('new_points', e, overwrite=True)
    # exception (data is deleted, we should raise an error before this happens)
    sdata.add_points('new_points', sdata.points['new_points'], overwrite=True)
```

In the latest version from main (I suggest doing a develop install of the repo so that you can install the latest version by just pulling the repo (or changing branch)), the error that you got should not happen: you should get an exception before it happens.

```
[...]

Issue #103, opened 2023-01-14
modules import hierarchy


giovp:
right now, basically every module can be import from spatialdata main module.

e.g. this works:
```python
from spatialdata import SpatialData, Scale, Image2DModel
```

I think it's a bit confusing and I think something like this would be clearer
```python
from spatialdata import Spatialdata
from spatialdata.models import Image2DModel
from spatialdata.transforms import Scale
```

ideas? @scverse/spatialdata

LucaMarconato:
I agree with you, I think the above approach is a good one. Only thing to pay attention is that need to split the imports from `spatialdata._core.core_utils` into the appropriate modules.

LucaMarconato:
Notes:
* `->` indicates a renaming of the currently used nomenclature
* functions/classes that are not listed here, are not public. The one here needs to be in the docs.

# spatialdata

- SpatialData
- read_zarr
- concatenate
- get_dims -> get_axis_names

# spatialdata.transform

- BaseTransformation
- Identity
- MapAxis
- Translation
- Scale
- Affine
- Sequence
- set_transform
- get_transform
- remove_transform
- get_transformation_between_coordinate_systems
- get_transformation_between_landmakrs
- align_elements_using_landmarks

# spatialdata.models

- Labels2DModel
- Labels3DModel
- Image2DModel
- Image3DModel
- PolygonsModel
- ShapesModel
- PointsModel
- TableModel
- get_schema -> get_model

# spatialdata.aggregation

- aggregate_numerical_annotations
- aggregate_categorical_annotations
- aggregate_elements

# spatialdata.query

- bounding_box_query(element: SpatialElement, min_coordinate: ArrayLike, max_coordinate: ArrayLike, coordinate_system: str)

# spatialdata.element_utils

- unpad_raster
- get_table_mapping_metadata

giovp:
sounds good, couple of other things to change to keep track of:
- I would still add an `io` module for readers, writers and format.
- I would still keep spatialdata and models in `_core`
- I would be consistent with `utils.py` file across repo (e.g. always private `_utils.py` and also not stuff like `core_utils.py`

giovp:
pinging @scverse/spatialdata for comments here since this is quite important and should be done by end of this week imhop

LucaMarconato:
> I would still keep spatialdata and models in _core

Basically an empty package `models` whose `__init__.py` imports from `spatialdata._core.models`?

LucaMarconato:
[...]

PR #138, opened 2023-02-14
Io/hot fix

LucaMarconato:
Fixing a critical io bug https://github.com/scverse/spatialdata/issues/137 reported by @melonora and a related bug discussed in the same issue.


LucaMarconato:
@giovp I merged https://github.com/scverse/spatialdata/pull/136 into this PR.

spatialdata/__main__.py
```diff
@@ -0,0 +1,36 @@
+import os
```
giovp, line 1: this is cool, could the usage of this be documented somewhere in docs?

spatialdata/_core/_spatialdata.py
```diff
@@ -324,17 +326,22 @@ def _write_transformations_to_disk(self, element: SpatialElement) -> None:
...
+        coordinate system(s).
 
         Parameters
         ----------
         coordinate_system
-            The coordinate system to filter by.
+            The coordinate system(s) to filter by.
+        filter_table
+            If True, the table will be filtered to only contain regions of an element belonging to the specified
+            coordinate system(s).
```
giovp, line 344: I think this should be True by default. I understand why `False` but I don't think it should be suggested behaviour.

spatialdata/_core/_spatialdata.py
```diff
@@ -430,40 +450,59 @@ def add_image(
...
-            elem_group = self._init_add_element(name=name, element_type="images", overwrite=overwrite)
-            write_image(
-                image=self.images[name],
-                group=elem_group,
-                name=name,
-                storage_options=storage_options,
-            )
-            # reload the image from the Zarr storage so that now the element is lazy loaded, and most importantly,
-            # from the correct storage
+            des_path = os.path.join(str(self.path), "images", name)
```
giovp, line 462: why first casting to `str` and then using join? I think it'd be nicer to have 
```
Path(self.path) / "images" / name
```

giovp: and actually ideally self.path is already a `Path`

giovp: I would also rename variable to `tgt_path`

spatialdata/_core/_spatialdata.py
```diff
@@ -430,40 +450,59 @@ def add_image(
             See https://zarr.readthedocs.io/en/stable/api/storage.html for more details.
         overwrite
             If True, overwrite the element if it already exists.
-        _add_in_memory
+        _called_from_write
             Internal flag, to differentiate between an element added by the user and an element saved to disk by
             write method.
```
giovp, line 455: for this behaviour in general I would like to ask @ivirshup to comment

spatialdata/_core/_spatialdata.py
```diff
@@ -430,40 +450,59 @@ def add_image(
...
+                        name=name,
+                        storage_options=storage_options,
+                    )
+                    src_path = os.path.join(store.path, name)
+                    assert isinstance(self.path, str)
+
+                    if os.path.isdir(des_path) and overwrite:
+                        shutil.rmtree(des_path)
+                    pathlib.Path(des_path).parent.mkdir(parents=True, exist_ok=True)
+                    shutil.move(src_path, des_path)
```
giovp, line 488: this is the way the temporary file is created and handled by spatialdata, @ivirshup what do you think of this? I personally don't feel comfortable adding such behaviour and in general using `shutil` anywhere in the repo. I would rather constrained the behaviour and let user be explicitly about what's in memory and what on disk.

spatialdata/_core/_spatialdata.py
```diff
@@ -711,6 +823,40 @@ def table(self) -> AnnData:
         """
         return self._table
 
+    @table.setter
```
giovp, line 826: this is nice and very much needed

spatialdata/_core/_spatialdata.py
```diff
@@ -711,6 +823,40 @@ def table(self) -> AnnData:
...
+            write_table(table=self.table, group=root, name="table")
+
+    @table.deleter
+    def table(self) -> None:
+        """Delete the table."""
+        self._table = None
+        if self.is_backed():
+            store = parse_url(self.path, mode="r+").store
+            root = zarr.group(store=store)
+            del root["table"]
```
giovp, line 858: could you elaborate on the use case of this?

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,165 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
             transformations.append(g[path[i]][path[i + 1]]["transformation"])
         sequence = Sequence(transformations)
         return sequence
+
+
+##
```
giovp, line 293:
```suggestion

```

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,165 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+        if MERGED_TABLES_REGION_KEY in table.obs:
+            del table.obs[MERGED_TABLES_REGION_KEY]
+
+    merged_table.uns[TableModel.ATTRS_KEY] = attr
+    merged_table.obs[merged_region_key] = merged_table.obs[merged_region_key].astype("category")
+    TableModel().validate(merged_table)
+    return merged_table
+
+
+##
```
giovp, line 399:
```suggestion

```

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,165 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+    Not all tables can be merged, they need to have compatible region, region_key and instance_key values. This function
+    checks this and merges if possible.
+
+    """
+    if len(tables) == 0:
+        return None
+    if len(tables) == 1:
+        return tables[0]
+
+    MERGED_TABLES_REGION_KEY = "annotated_element_merged"
```
giovp, line 318: I think this should be also available in the signature for the user to choose

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,165 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+            key = f"{MERGED_TABLES_REGION_KEY}_{i}"
+
+        all_without = True
+        for table in tables:
+            if key in table.obs:
+                all_without = False
+                break
+        if all_without:
+            MERGED_TABLES_REGION_KEY = key
+            break
```
giovp, line 333: I'm missing what's this code suppose to be doing, could you please elaborate?

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,165 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
             transformations.append(g[path[i]][path[i + 1]]["transformation"])
         sequence = Sequence(transformations)
         return sequence
+
+
+##
+def concatenate_tables(tables: list[AnnData]) -> Optional[AnnData]:
```
giovp, line 294: this I'm also not super sure on its use. My understanding is that it provides a function for the user to concatenate tables __outside__ of SpatialData yet compliant with SpatialData constraints. Why would a user do that instead of concatenating the SpatialData directly (which btw could just be SpatialData objs with only table) ? 
Furthermore, if the user really wants to do this, then the validator would spot errors (maybe a check for nans should be added) and the user would have to fix it this anyway.

In general, I don't see very much how much the user should handle elements outside of spatialdata, and hence I don't see why we should be providing functions for this kind of usage. Maybe worth discussing briefly @scverse/spatialdata ?

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,165 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+    from spatialdata._core._spatialdata import SpatialData
+
+    assert type(sdatas) == list
+    assert len(sdatas) > 0
+    if len(sdatas) == 1:
+        return sdatas[0]
+
+    if not omit_table:
+        list_of_tables = [sdata.table for sdata in sdatas if sdata.table is not None]
+        merged_table = concatenate_tables(list_of_tables)
```
giovp, line 424: if this is where it's used, then I would 💯 make it private func

spatialdata/_core/_transform_elements.py
```diff
@@ -246,3 +248,77 @@ def _(data: AnnData, transformation: BaseTransformation) -> AnnData:
 
     ShapesModel.validate(transformed_adata)
     return transformed_adata
+
+
+def get_transformation_between_landmarks(
+    references_coords: AnnData,
+    moving_coords: AnnData,
+) -> Affine:
```
giovp, line 256: this should be documented. Also, wouldn't this go under some `utils` module?

spatialdata/utils.py
```diff
@@ -137,3 +147,26 @@ def _unpad_axis(data: DataArray, axis: str) -> tuple[DataArray, float]:
...
+    -------
+    The `region`, `region_key`, and `instance_key` values.
+    """
+    from spatialdata._core.models import TableModel
+
+    TableModel().validate(table)
+    region = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY]
+    region_key = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY_KEY]
+    instance_key = table.uns[TableModel.ATTRS_KEY][TableModel.INSTANCE_KEY]
+    d = {TableModel.REGION_KEY: region, TableModel.REGION_KEY_KEY: region_key, TableModel.INSTANCE_KEY: instance_key}
```
giovp, line 171: but isn't this information basically already in `table.uns[TableModel.ATTRS_KEY]` ?

giovp (changes requested):
I must say the temporary file and in-memory/on-disk behavior I'd really like to hear @ivirshup thoughts but I really don't feel comfortable adding anything with `shutils` in the code.

spatialdata/_core/_spatialdata.py
LucaMarconato: One clarification, this code is not about choosing what is in memory and what on disk, but to deal with the following case:
1. the user load a `SpatialData` object from disk. The object contain the image `image0`, this image is a backed, lazy-loaded `xarray.DataArray`.
2. if the user modifies this object, say it doesn't a `log(1 + x)` transform, and wants to save the object into a new slot, like `image1`, this gives no problems
3. if the user instead wants to replace the `image0` slot ([like done from Wouter here](https://github.com/scverse/spatialdata/issues/137)), this requires to first write the data somewhere else, and then replacing the original location. 

Maybe there is a library for doing this kind of "atomic swap" that we can use, so we don't have to call `shutils.rmtree` implicitly. But also, if we define `des_path` right before `rmtree` (now it is defined 10 lines above), and if we do a check like `assert str(des_path.parent).endswith('.zarr')`, then I consider the call to `shutils.rmtree` to be safe. What do you think @ivirshup. Also to the attention of @kevinyamauchi

spatialdata/_core/_spatialdata.py
LucaMarconato: I removed this behavior, I realized it's bad practice because the user may trigger untested behaviors. Now the code used from `write()` is inside `write()`, and the `add_image()`, `add_labels()`, ... functions are not called from `write()`.

spatialdata/_core/_spatialdata.py
LucaMarconato: I removed `shutils`, see comment in the discussion.

spatialdata/_core/_spatialdata_ops.py
LucaMarconato: You are correct, this function is supposed to be private. It will only be called by `concatenate(sdatas)` and not by the user. I will change it to private.

spatialdata/_core/_spatialdata_ops.py
LucaMarconato: Done, thanks.

spatialdata/utils.py
LucaMarconato: True 😅 I remove this function

spatialdata/_core/_transform_elements.py
LucaMarconato: Added a docstring. Kevin and I propose to place the method in `spatialdata.transform`, see here: https://github.com/scverse/spatialdata/issues/103

spatialdata/_core/_spatialdata_ops.py
LucaMarconato: Now also added a comment in the code. Pasting it here:
```
    # 1) if REGION is a list, REGION_KEY is a string and there is a column in the table, with that name, specifying
    # the "regions element" each key is annotating; 2) if instead REGION is a string, REGION_KEY is not used.
    #
    # In case 1), we require that each table has the same value for REGION_KEY (this assumption could be relaxed,
    # see below) and then we concatenate the table. The new concatenated column is correcly annotating the rows.
    #
    # In case 2), we require that each table has no REGION_KEY value. In such a case, contatenating the tables would
    # not add any "REGION_KEY" column, since no table has it. For this reason we add such column to each table and we
    # call it "annotated_element_merged". Such a column could be already present in the table (for instance merging a
    # table that had already been merged), so the for loop before find a unique name. I added an upper bound,
    # so if the user keeps merging the same table more than 100 times (this is bad practice anyway), then we raise an
    # exception.
    #
    # Final note, as mentioned we could relax the requirement that all the tables have the same REGION_KEY value (
    # either all the same string, either all None), but I wanted to start simple, since this covers a lot of use
    # cases already.
```

spatialdata/_core/_spatialdata_ops.py
LucaMarconato: Since `concatenate_table()` will be private we would have to expose this in `concatenate()`, but it's a tricky parameter, I am not sure we want the user to have to think about this when calling the method: sometimes a new column is added, other times the `REGION_KEY` value can be kept the one that was already present. In the comment below where I explain the code with the for loop I have explained that some edge cases have been left out. I would wait until we address those cases, and in that moment I would expose `MERGED_TABLES_REGION_KEY`, since its logic would change. wdyt?

spatialdata/_core/_spatialdata.py
LucaMarconato: In the setter I don't allow to overwrite a table by default to prevent accidental data loss. I do the same in `add_image()`, `add_labels()`, etc. In those functions I could add a `overwrite` parameter, to be sure that the user wanted to replace the element, but with a setter I could not. So if the user tries to replace an existing table I raise an exception in which I ask to explicitly delete the table first.

spatialdata/__main__.py
LucaMarconato: I don't know how to add a documentation for this. Do you have an example in mind I could look at?

spatialdata/_core/_spatialdata.py
LucaMarconato: changed the default to true

spatialdata/_core/_spatialdata.py
LucaMarconato: changed the name of the variable (and all instances of des/destination to tgt/target in `_spatialdata.py`). Also using `Path`.

LucaMarconato:
@giovp all the changes (except the docs for `__main__.py`, see comment) are addressed. Please merge if you feel is good.

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -281,3 +288,183 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+            region = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY]
+            if isinstance(region, str):
+                all_regions.append(region)
+            else:
+                all_regions.extend(region)
+        all_regions = list(set(all_regions))
+        merged_region = all_regions
+
+    attr = {"region": merged_region, "region_key": merged_region_key, "instance_key": merged_instance_key}
+    merged_table = AnnData.concatenate(*tables, join="outer", uns_merge="same")
```
ivirshup, line 401: This is going to start giving deprecation warnings soon. Use `ad.concat`.

spatialdata/_core/_spatialdata.py
```diff
@@ -430,40 +448,52 @@ def add_image(
...
-            write_image(
-                image=self.images[name],
-                group=elem_group,
-                name=name,
-                storage_options=storage_options,
-            )
-            # reload the image from the Zarr storage so that now the element is lazy loaded, and most importantly,
-            # from the correct storage
+            self._add_image_in_memory(name=name, image=image, overwrite=overwrite)
+            with tempfile.TemporaryDirectory() as tmpdir:
```
ivirshup, line 461: Why does adding an image cause any write to happen? Couldn't the image stay where it is, and we just hold a delayed representation here?

Then, when `write` is called explicitly, its written?

spatialdata/__main__.py
ivirshup: What does this have to do with the IO hot fix?

Also, is this functionality being added for the first time here? I think this should probably live in a separate package which relies on spatial data, but provides command line utilities.

spatialdata/_core/_spatialdata.py
giovp: that also a bit concerning to me, I'd rather have a warning saying that it's not backed/saved rather than creating a temporary dir

spatialdata/__main__.py
LucaMarconato: @ivirshup I merged this PR with another one that hasn't merged yet. So the io part is mixed with the rest, sorry.

LucaMarconato: I would keep the CLI momentarily in spatialdata because we have a lot of repos already, but we could move it later on if the CLI gets too big. I plan to just add a few extra functions for now (like exporting the polygons to csv and things like this).

spatialdata/_core/_spatialdata.py
LucaMarconato: @ivirshup the motivation to writing to disk is the following: if a spatialdata object is backed (=it has been written to disk or it has read from disk; note, when I write I immediately re-read to address [this problem](https://github.com/scverse/spatialdata-io/issues/12)), then every time I add something to it: an image, the table, a transformation, I immediately save it to disk; I believe that this makes the life easier to the user, and for us, because we don't have to keep track to what has been saved and what not

@giovp calling write explicitly would not remove the need for the temporary directory. The temporary directory is for performing an atomic save. In this way if the destination of the save is the same as the origin of the backed data, we can still perform the save operation.

LucaMarconato: Consider the following example:

```
sdata = read_zarr(...)
sdata.images['my_image'] = dask.array.log1p(sdata.images['my_image']
```

sdata.images['my_image'] is never loaded in memory, but it saved chunk by chunk from dask. This operation is possible because of the atomic save, otherwise we would not be able to modify the group. This issue is what Wouter reported.

spatialdata/__main__.py
LucaMarconato: Added an issue about documenting the CLI so that we will not forget to do it in the future: https://github.com/scverse/spatialdata/issues/144

giovp:
> 1. Do we need to perform an atomic save using temporary directories? I think this is unavoidable if we want to prevent data corruption

I think for now this is not unavoidable, went back to #137 and for this specific PR I would do the following:
- if the file already exist on disk, simply raise ValueError and force the user to name the element differently/write in a different path
- discuss in separate PR how to handle this behaviour and whether we want to use temporary directories (I would tend to be against it unless there is strong user request).

LucaMarconato:
The problem is not when we call `sdata.write()` (in that case as you pointed out we can see if the target corresponds to where the user wants to save the data and raise an exception), but when we do `add_image()`, `add_labels()` and `add_points()`.

In these three cases we could intercept the fact that the user would be overwriting the data, I show the code below, but it's hacky. With dask objects there is no attribute telling that the data is backed on disk, instead the information is stored in a node in the computational graph that is responsible for reading the data from disk. To see a code example related to this for Points, see this recent issue in which I needed to know the `.parquet` file associated to a `DaskDataFrame` https://github.com/scverse/spatialdata/issues/140, and the code to fix it: and in this PR I show the code to fix it https://github.com/scverse/spatialdata/pull/141. 

Here below I show how the code looks like for images, for two cases of increasing complexity:

### Case 1
```
sdata = read_zarr('data.zarr')
image = sdata.images['image']
sdata.add_image('image', image, overwrite=True)
```
to raise an exception we can inspect the first node of the dask computational graph of `image` (first node = the first dict item of the dict `image.data.dask.layers`), and if its key name starts with `"original-read-zarr-"`, we can inspect it's value, find the path that is being read and raise an exception since it corresponds to
`f"{sdata.path}/images/{image}"`.

### Case 2
```
sdata = read_zarr('data.zarr')
image0 = sdata.images['image0']
image1 = sdata.images['image1']
image2 = image0 + image1
sdata.add_image('image0', image2, overwrite=True)
```
to raise an exception we can inspect all the nodes of the dask computational graph of `image2` that start with `original-read-zarr-`, and check for their path as above.

This is the code that addresses both cases for `SpatialImage` is the following. For `MultiscaleSpatialImage` and `DaskDataFrame` the code is slightly different, but the logic is the same.

```
# not a real Python test, but can be turned into one
def test_wip_io_hotfix(images, labels):
    with tempfile.TemporaryDirectory() as tmpdir:
        images.write(os.path.join(tmpdir, "images.zarr"))
        labels.write(os.path.join(tmpdir, "labels.zarr"))
        image0 = images.images['image2d']
        image1 = images.images['image2d_multiscale']
        image2 = image0[:, :32, :32] + image1['scale1']['image']

        def check(target_sdata, target_name, image):
            for k, v in image.data.dask.layers.items():
                if k.startswith('original-from-zarr-'):
                    mapping = v.mapping[k]
                    path = mapping.store.path
                    if path == os.path.join(target_sdata.path, 'images', target_name):
                        print('raise an exception!')

        # this code tells that we should raise an exception when calling images.add_image(name='image2d', image=image0, overwrite=True)
        check(target_sdata=images, target_name='image2d', image=image0)

        # this code tells that we should raise an exception when calling images.add_image(name='image2d', image=image2, overwrite=True)
        check(target_sdata=images, target_name='image2d', image=image2)
```

LucaMarconato:
I could implement the above check so that we can temporarily fix the problem behind this PR (right now if the user calls `add_image(..., overwrite=True)` the data is lost) and go back to atomic save in another PR (which we should address soon).

In any case, I thought more about atomic save and did some research. I think it is the solution we want and that it is a clean approach. It would not just a raise an exception like above but actually save the data, and it would also protect against errors during writing that otherwise could lead to data corruption. 

There are also some criticisms regarding atomic save: if the move process fails, then the data could get lost, and if we use temporary directories the data is irretrievable. We discussed this with Kevin and we could:
* use a directory know to spatialdata (like ~/spatialdata_cache) so that if something goes wrong the user can find the data there.
* replace "move" with "copy" and then "delete", so that in every given moment the data is available in full in the disk.

LucaMarconato:
More context for you. Here is the content of some objects from the code above.
```python console
In[70]: image2.data.dask.layers
Out[70]: 
{'original-from-zarr-cd7b4593de9695770d47b0911b35b8b7': <dask.highlevelgraph.MaterializedLayer at 0x7fab2ae8a710>,
 'from-zarr-cd7b4593de9695770d47b0911b35b8b7': Blockwise<(('original-from-zarr-cd7b4593de9695770d47b0911b35b8b7', None), ('a1f223a32a269d1e6a4e6213bbd28aa8', ('.0', '.1', '.2'))) -> from-zarr-cd7b4593de9695770d47b0911b35b8b7>,
 'getitem-5e5a71556ac5a9f7612042285545694d': <dask.highlevelgraph.MaterializedLayer at 0x7fab2af01b40>,
 'add-f3c908ac106a2793255112a01caa666b': Blockwise<(('getitem-5e5a71556ac5a9f7612042285545694d', ('.2', '.1', '.0')), (3, None)) -> add-f3c908ac106a2793255112a01caa666b>}
```

```python console
In[71]: image2.data.dask.layers['original-from-zarr-cd7b4593de9695770d47b0911b35b8b7']
Out[71]: <dask.highlevelgraph.MaterializedLayer at 0x7fab2ae8a710>
```

```python console
In[72]: image2.data.dask.layers['original-from-zarr-cd7b4593de9695770d47b0911b35b8b7'].__dict__
Out[72]: 
{'annotations': None,
 'collection_annotations': None,
 'mapping': {'original-from-zarr-cd7b4593de9695770d47b0911b35b8b7': <zarr.core.Array '/0' (3, 64, 64) float64 read-only>}}
```

```python console
In[73]: image2.data.dask.layers['original-from-zarr-cd7b4593de9695770d47b0911b35b8b7'].mapping['original-from-zarr-cd7b4593de9695770d47b0911b35b8b7'].store.path
Out[73]: '/var/folders/gb/1q317sg54wl4gzgpq_gjzw380000gn/T/tmp08rnvend/images.zarr/images/image2d'
```

giovp:
thanks for the detailed explanation!

> There are also some criticisms regarding atomic save: if the move process fails, then the data could get lost, and if we use temporary directories the data is irretrievable. We discussed this with @kevinyamauchi and we could:

> use a directory know to spatialdata (like ~/spatialdata_cache) so that if something goes wrong the user can find the data there.

I think this is the most reasonable solution but:
- I'd like to hear @ivirshup 
- could it be maybe addressed in separate PR?

spatialdata/_core/_spatialdata.py
ivirshup: So `add_image` causes the array represented by `normalized` to compute and write to disk?

I think that's not great, and skips out on a lot of what makes dask useful.

I assume that would run when either `sdata.write(...)` or `sdata.compute(...)` was called.

ivirshup: > atomic save

Atomicity would be great, but also is a high bar and not something I think is in scope for 0.1. TBH, I would say "we should never aim for atomic updates with `zarr.DirectoryStore`". I think we could maybe (at some point in the future) do that with a kerchunk or iceberg style store.

ivirshup:
I strongly agree with @giovp's suggestion here: https://github.com/scverse/spatialdata/pull/138#issuecomment-1435961680

> * if the file already exist on disk, simply raise ValueError and force the user to name the element differently/write in a different path
> * discuss in separate PR how to handle this behaviour and whether we want to use temporary directories (I would tend to be against it unless there is strong user request).

> (1) Do we need to perform an atomic save using temporary directories?

To me, what you're describing is an atomic *update*, which I think is quite a lot of work and out of scope for v0.1. Nice to have, yes. Complicated to make work (especially on diverse storage backends), and difficult to make performant? Extremely.

I think we should broadly avoid inplace updates.

Saving to temporary directories is also not going to be a great way to do this. Lets say I want to add a pyramidal image to my object, so I can view my annotations with napri. One could do this on their laptop, unless you try to write the whole pyramidal image.

One of the biggest goals here should be to **avoid** writing things, and being able to reference remote data.

We also probably shouldn't be relying on the presence of "directories with lots of storage" to write to. One of the big goals here is being ready for cloud storage!

> (2) Do we save the data to disk every time the user adds an element to disk only when the user saves the data.

Only when the user saves data seems very reasonable.

I think a configurable cache may make sense in some circumstances (e.g. intermediate results for a large computation, not "adding an element"), but we can probably rely on fsspec and dask for that.

LucaMarconato:
Thanks for the comments. I will do the following, please tell me if you agree with it.
- [ ] disable the possibility to write a `sdata` object to the path where it is backed;
- [ ] disable the possibility to replace the value of an element, either in-memory or backed, with another value. If the user wants to assign another value to `sdata.images['my_image']`, the previous one needs to be deleted first. To perform this check I will use the strategy described in this comment https://github.com/scverse/spatialdata/pull/138#issuecomment-1435974481. Please tell me if you think it is good. In particular I would like to hear from @kevinyamauchi (who reviewed similar code in the PR for the lengths of Points).

I will then open two issues/Github discussions to keep track of things:
- [ ] an issue for keeping a discussion on the possibility to replace an object in-place using an atomic update.
- [ ] an issue for letting the user save the data by calling explicitly a `save()` as opposed to the current behavior.

For the moment I prefer to keep the current behavior of saving to disk every time the user adds an element. This for two reasons:
- [ ] we don't have a mechanism for knowing what changed and what not
- [ ] when I call `set_transfomation()` I would expect the disk storage to be updated, not just the memory storage. I prefer to act on the safe side and save immediately to disk.

LucaMarconato:
I have just realized two implications on the strategy described in the message above. I am not sure anymore we want to proceed as I described in the previous message.

## 1. updates with non-backed objects.
If the user adds a new column to a table, or if with the user saves a polygon annotation with the napari plugin but then wants to add a new polygons and resave, what will be done is that the data is loaded in-memory. The table or shape elements is removed from spatialdata object in-memory and on-disk, and then the data is written again to disk. Here the data is small (not huge images), so basically what happens here to me sounds like a version of the atomic update where the temporary directory is replaced by just memory! Sounds bad to me! I would be much more comfortable with the atomic update strategy described above (and having a spatialdata cache directory in this case should be fine since the data is small).

## 2. updates with non-backed objects.
I agree with Isaac that since images and labels could be large, we should not encourage the user to change them, and we should not rely on the presence of a directory with a lot of space. So my previous message remains valid: I agree on having a check that raises an exception if the user user to replace and image slot or a labels slot.
But for points it is different. I expect the user to be interested in computing cell types or other metadata for points and wanting to update the points `DaskDataFrame` (as it is possible for the `AnnData` table). Here the data is small: approximately about 300 MB every million points and probably split into different fields of view. In principle an approach like "load everything in memory, delete the previous points slot, save again" would work, but it would take several seconds (like 10 in my machine), and if the user killed the Python process meanwhile the data would be lost! Again here I would be more comfortable with the atomic update approach, and I think it should not be a problem because the data is fairly small.

Please @scverse/spatialdata I would appreciate feedback on this.

LucaMarconato:
> To me, what you're describing is an atomic update, which I think is quite a lot of work and out of scope for v0.1. Nice to have, yes. Complicated to make work (especially on diverse storage backends), and difficult to make performant? Extremely.

I think that if we disable the atomic update for images and labels, and we keep the code of this PR (adding the improvements described in this comment https://github.com/scverse/spatialdata/pull/138#issuecomment-1434346128), then we have a rudimentary but perfectly working and performant version of the atomic update. Rudimentary because we just replace the whole element with the new updated copy, but performant because since the data involved is small, we can afford a full duplication.

LucaMarconato:
This PR contained also code from this other PR https://github.com/scverse/spatialdata/pull/136 (unrelated to IO). I addressed the changes and merged to main.

Thanks for the comments @giovp, answering down here:
> while i understand the use case, precisely because the data is small i think it'd be fine to have everything in memory and then only when the user wants to save, is written to disk.

This keeps open the possibility that the user kills the Python process because it takes too long and loses the data. It is something that probably I would also do as a user in case I accidentally run the wrong operation and it takes too long.

> this also refer to the point on "if set_transformation is called then expect this is also stored to disk", im not sure we should be expecting this and would be fine if that's not the case

I open an issue about this, I would like to gather some feedback from the users. I think that it easy to forget to call a function like `write_element_on_disk()` after calling `set_transformation()`. Especially because `set_transformation()` is already an API that makes the life easier to the user. If I used `my_image.attrs['TRANSFORM']['global'] = ...`, it would be obvious to me that I need to rewrite the element. But if I already call an API for that I think I would expect that the API does the job of writing already for me.

> I don't have experience with this enough to be able to really chime in, i guess i would be fine either way and maybe this is now too much of a blocker. so i would be ok to merge it as is but i would really consider to return to this asap and find a better solution

I would be in favor to this approach; let's wait for the feedback from the others. Meanwhile I have fixed the tests.

LucaMarconato:
I have implemented the changes and now this PR is now ready for review and for being merged @giovp 

In details: I have removed the atomic save approach, restored the check that prevents `sdata` objects to be written into themselves, and added checks to `add_image()`, `add_labels()` and `add_points()` to prevent them to be added into a zarr group that is used for backing (I look at the dask computational graph for doing this). The old code is commented out, I preferred not to remove it because we will have to get back to this and it will make it easier to restore/adapt the old code.

LucaMarconato:
@melonora, after this PR will be merged you won't be able to overwrite the sdata object, or spatial data elements to themselves anymore (please refer to the above for discussion). We will have to get back to this behavior, so please your feedback regarding this (also later on in the future when you will get more familiarity with the package), will be highly valuable and appreciated.

spatialdata/_core/_spatialdata.py
```diff
@@ -425,40 +426,78 @@ def add_image(
             See https://zarr.readthedocs.io/en/stable/api/storage.html for more details.
         overwrite
             If True, overwrite the element if it already exists.
-        _add_in_memory
+        _called_from_write
             Internal flag, to differentiate between an element added by the user and an element saved to disk by
             write method.
```
giovp, line 431: thing this argument is removed right? would remove from docstring then

spatialdata/_core/_spatialdata.py
```diff
@@ -425,40 +426,78 @@ def add_image(
...
         if self.is_backed():
+            files = get_backing_files(image)
+            assert self.path is not None
+            target_path = os.path.join(self.path, "images", name)
+            if target_path in files:
+                raise ValueError(
+                    "Cannot add the image to the SpatialData object because it would overwrite an element that it is"
+                    "using for backing. We are considering changing this behavior to allow the overwriting of "
+                    "elements used for backing. If you would like to support this use case please leave a comment on "
+                    "https://github.com/scverse/spatialdata/pull/138"
```
giovp, line 446: would shorten this error by max 1 line, ok to leave the pull/238 reference

spatialdata/_core/_spatialdata.py
```diff
@@ -425,40 +426,78 @@ def add_image(
...
+            #     for file in os.listdir(str(src_element_path)):
+            #         src_file = src_element_path / file
+            #         tgt_file = tgt_element_path / file
+            #         os.rename(src_file, tgt_file)
+            # from spatialdata._io.read import _read_multiscale
+            #
+            # # reload the image from the Zarr storage so that now the element is lazy loaded, and most importantly,
+            # # from the correct storage
+            # image = _read_multiscale(str(tgt_element_path), raster_type="image")
+            # self._add_image_in_memory(name=name, image=image, overwrite=True)
```
giovp, line 476: I would be against leaving large chunks of uncommented code, although I understand that this is something we want to go back to it quick, so feel free to leave in

spatialdata/_core/_spatialdata.py
```diff
@@ -425,40 +426,78 @@ def add_image(
...
-            # from the correct storage
             from spatialdata._io.read import _read_multiscale
 
+            # reload the image from the Zarr storage so that now the element is lazy loaded, and most importantly,
+            # from the correct storage
             assert elem_group.path == "images"
-            path = os.path.join(elem_group.store.path, "images", name)
-            image = _read_multiscale(path, raster_type="image")
+            path = Path(elem_group.store.path) / "images" / name
+            image = _read_multiscale(str(path), raster_type="image")
```
giovp, line 490: this is nitpick I undersatnd, but it looks like this is because `ZarrLocation` from `ome-zarr-py` doesn't accept paths but only strings? would it be a solution to put the str casting only there, and stick to `Path()` in the repo? instead of os.path.join, used also above

spatialdata/_core/_spatialdata.py
```diff
@@ -591,62 +695,177 @@ def add_shapes(
...
+                #     target_path = tempfile.TemporaryDirectory()
+                #     tmp_zarr_file = Path(target_path.name) / "data.zarr"
+
+        # old code to support overwriting the backing file
+        # if target_path is None:
+        #     store = parse_url(file_path, mode="w").store
+        # else:
+        #     store = parse_url(tmp_zarr_file, mode="w").store
+        # store = parse_url(file_path, mode="w").store
+        # root = zarr.group(store=store)
```
giovp, line 741: same comment here and above

spatialdata/_core/_spatialdata.py
```diff
@@ -591,62 +695,177 @@ def add_shapes(
...
+        if self.is_backed() and self.path != file_path:
+            logger.info(f"The Zarr file used for backing will now change from {self.path} to {file_path}")
+
+        # old code to support overwriting the backing file
+        # target_path = None
+        # tmp_zarr_file = None
+        if os.path.exists(file_path):
+            if parse_url(file_path, mode="r") is None:
+                raise ValueError(
+                    "The target file path specified already exists, and it has been detected not not be "
```
giovp, line 717:
```suggestion
                    "The target file path specified already exists, and it has been detected to not be "
```

spatialdata/_core/_spatialdata.py
```diff
@@ -591,62 +695,177 @@ def add_shapes(
...
+                    "a Zarr store. Overwriting non-Zarr stores is not supported to prevent accidental "
+                    "data loss."
+                )
+            if not overwrite and self.path != str(file_path):
+                raise ValueError("The Zarr store already exists. Use overwrite=True to overwrite the store.")
```
giovp, line 722:
```suggestion
                raise ValueError("The Zarr store already exists. Use `overwrite=True` to overwrite the store.")
```

spatialdata/_core/_spatialdata.py
```diff
@@ -591,62 +695,177 @@ def add_shapes(
...
+        # else:
+        #     self.path = str(tmp_zarr_file)
+        self.path = str(file_path)
         try:
             if len(self.images):
                 root.create_group(name="images")
-                # add_image will delete and replace the same key in self.images, so we need to make a copy of the
+                # add_image_in_memory will delete and replace the same key in self.images, so we need to make a copy of the
                 # keys. Same for the other elements
                 keys = list(self.images.keys())
```
giovp, line 758: it looks like we are not checking that the keys match though right>? shall we be checking for that? otherwise why make a copy?

spatialdata/_core/_spatialdata.py
```diff
@@ -591,62 +695,177 @@ def add_shapes(
...
+                        labels=self.labels[name],
+                        group=elem_group,
+                        name=name,
+                        storage_options=storage_options,
                     )
 
+                    # reload the labels from the Zarr storage so that now the element is lazy loaded, and most importantly,
+                    # from the correct storage
+                    element_path = Path(self.path) / "labels" / name
+                    labels = _read_multiscale(str(element_path), raster_type="labels")
```
giovp, line 793: same here, I'd ttry to be consistent everywhere

giovp (approved):
minor things re doc.

one thing is, why are we copying keys if we then overwrite?

spatialdata/_core/_spatialdata.py
LucaMarconato: Actually `ome-zarr-py` accepts `Path`, I was using an `assert` that was not necessary. I changed this and now I have removed the conversion to `str`. I am still keeping `SpatialData.path`as a `str` (we can change it later)

spatialdata/_core/_spatialdata.py
LucaMarconato: I didn't get exactly what you mean but I fixed this. The copy was to avoid this error: `RuntimeError: dictionary keys changed during iteration`, since `_add_image_in_memory()` (same for the other elements), was doing a `del imags[key]` when `overwrite=True`. I removed this `del` since when we do `image[key] = new_image`, we remove the old content and we don't change the keys.

spatialdata/_core/_spatialdata.py
LucaMarconato: As above, I removed this cast to `str`.

LucaMarconato:
Addressed all the comments. I also fixed a but, I noticed that I had to use `os.path.realpath()` before comparing the path used for backing and the path where we want to write. In this way we resolve for symlinks. Merging now 🚀

Issue #140, opened 2023-02-16
Computing the length of `DaskDataFrame` is slow


LucaMarconato:
`len(points)` is slow because it is needs to compute the value of a `Delayed` object. This is a problem when printing a `sdata` object because the length needs to be computed: `print(sdata)` takes 5.5 seconds (40 million points among 19 dask dataframes).

Even when we load the data from `.parquet`, the `DaskDataFrame` object doesn't know the length (which is stored in `.parquet`).
This is because for dask the dataframe could have been filtered, so the length is treated as unknown.

I have opened an issue in dask, I am not sure how proceed https://github.com/dask/dask/issues/9973.

This is the dask computational graph for an object loaded from disk, one idea could be to look at the data associated to `read-parquet` layer, find the parquet file and read it with `pyarrow.parquet.read_table`, which is ultrafast. But it is difficult to distinguish between objects whose rows may have been filtered and whose who didn't. The computational graph shown here is just for an object that has been read from disk (I think the schema tried to convert things to categorical).

```
points.dask
Out[3]: 
HighLevelGraph with 7 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7fde0a68bfd0>
 0. read-parquet-2f971e340a47781eb685a87a213bf826
 1. getitem-cdfb73fd637e9e0ee5a292183add9458
 2. cat-set_categories-56f20ea96556105e025a60a300d9ea96
 3. assign-182714d359d11968871ea582587d16ec
 4. getitem-97ec6823c9a58fc7a55c932bcef626c3
 5. cat-set_categories-908e5810fd0c7d7bb0d1b4525d366fd7
 6. assign-298db01be23dc397178dae1c8ef8aa3e
```

LucaMarconato:
This is instead the computational graph for an object where the rows have been subset:

```
v[v['target'] == 0].dask
Out[7]: 
HighLevelGraph with 10 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7fdde943ff70>
 0. read-parquet-2f971e340a47781eb685a87a213bf826
 1. getitem-cdfb73fd637e9e0ee5a292183add9458
 2. cat-set_categories-56f20ea96556105e025a60a300d9ea96
 3. assign-182714d359d11968871ea582587d16ec
 4. getitem-97ec6823c9a58fc7a55c932bcef626c3
 5. cat-set_categories-908e5810fd0c7d7bb0d1b4525d366fd7
[...]

PR #141, opened 2023-02-16
computing len of points directly from parquet

LucaMarconato:
Fixes this issue: https://github.com/scverse/spatialdata/issues/140.

A bit hacky, but for the moment I think we don't have many options.

Sorry for the bad name of the PR, I wanted to fix also another thing but I will limit it to the issue linked.


kevinyamauchi (approved):
LGTM! Thanks!

PR #142, opened 2023-02-17
fixed BaseTransformation.transform(); print(sdata) shows coord systems

LucaMarconato:
This PR finishes some work on transformations that was incomplete. In details:
* `print(sdata)` now shows which coordinate systems are available and which elements are in.
* transforming a raster type now prepends the correct translation to the transformed element. This is crucial when the transformation is rotation, so padding is added on the rotated image (for a visualization of this, see last three plots of the [transformations notebook](https://github.com/scverse/spatialdata-notebooks/blob/main/notebooks/transformations_examples.ipynb)).
* added util `_get_affine_for_element()`, which given an element and a transformation, gives an affine transformation with the correct input and output axes.
* added util `_decompose_affine_into_linear_and_translation()`, to split an affine transformation into an affine with no translation part and and a translation part. Useful when plotting.
* fixed bug in `_get_current_output_axes()` discovered when composing an affine 2D -> 3D and a translation 3D -> 3D. Added tests
* Added more tests based on real use cases that I had left commented out in the previous transform PR.

### DIFF
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -873,7 +873,9 @@ class SpatialData:
             gen = self._gen_elements()
             elements_in_cs = []
             for k, name, obj in gen:
-                coordinate_systems = get_transformation(obj, get_all=True).keys()
+                transformations = get_transformation(obj, get_all=True)
+                assert isinstance(transformations, dict)
+                coordinate_systems = transformations.keys()
                 if cs in coordinate_systems:
                     elements_in_cs.append(f"/{k}/{name}")
             if len(elements_in_cs) > 0:
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -253,21 +253,40 @@ class SpatialData:
 
     def _locate_spatial_element(self, element: SpatialElement) -> tuple[str, str]:
         found: list[SpatialElement] = []
-        found_element_type: str = ""
-        found_element_name: str = ""
+        found_element_type: list[str] = []
+        found_element_name: list[str] = []
         for element_type in ["images", "labels", "points", "shapes"]:
             for element_name, element_value in getattr(self, element_type).items():
                 if id(element_value) == id(element):
                     found.append(element_value)
-                    found_element_type = element_type
-                    found_element_name = element_name
+                    found_element_type.append(element_type)
+                    found_element_name.append(element_name)
         if len(found) == 0:
             raise ValueError("Element not found in the SpatialData object.")
         elif len(found) > 1:
-            raise ValueError("Element found multiple times in the SpatialData object.")
-        return found_element_name, found_element_type
+            raise ValueError(
+                f"Element found multiple times in the SpatialData object. Found {len(found)} elements with names: {found_element_name}, and types: {found_element_type}"
+            )
+        assert len(found_element_name) == 1
+        assert len(found_element_type) == 1
+        return found_element_name[0], found_element_type[0]
 
     def contains_element(self, element: SpatialElement, raise_exception: bool = False) -> bool:
+        """
+        Check if the spatial element is contained in the SpatialData object.
+
+        Parameters
+        ----------
+        element
+            The spatial element to check
+        raise_exception
+            If True, raise an exception if the element is not found. If False, return False if the element is not found.
+
+        Returns
+        -------
+        True if the element is found; False otherwise (if raise_exception is False).
+
+        """
         try:
             self._locate_spatial_element(element)
             return True
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -290,12 +290,15 @@ class SpatialData:
             raise ValueError("Element found multiple times in the SpatialData object.")
         return found_element_name, found_element_type
 
-    def contains_element(self, element: SpatialElement) -> bool:
+    def contains_element(self, element: SpatialElement, raise_exception: bool = False) -> bool:
         try:
             self._locate_spatial_element(element)
             return True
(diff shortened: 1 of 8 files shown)


LucaMarconato:
Ready for review!

spatialdata/_core/_spatialdata.py
```diff
@@ -290,12 +290,15 @@ def _locate_spatial_element(self, element: SpatialElement) -> tuple[str, str]:
             raise ValueError("Element found multiple times in the SpatialData object.")
```
giovp, line 290: could this line print the element types and names?

spatialdata/_core/_spatialdata.py
```diff
@@ -290,12 +290,15 @@ def _locate_spatial_element(self, element: SpatialElement) -> tuple[str, str]:
...
         return found_element_name, found_element_type
 
-    def contains_element(self, element: SpatialElement) -> bool:
+    def contains_element(self, element: SpatialElement, raise_exception: bool = False) -> bool:
```
giovp, line 293: I would add a docstring here

spatialdata/_core/_spatialdata.py
```diff
@@ -861,6 +864,23 @@ def h(s: str) -> str:
...
+            elements_in_cs = []
+            for k, name, obj in gen:
+                transformations = get_transformation(obj, get_all=True)
+                assert isinstance(transformations, dict)
+                coordinate_systems = transformations.keys()
+                if cs in coordinate_systems:
+                    elements_in_cs.append(f"/{k}/{name}")
+            if len(elements_in_cs) > 0:
+                descr += f'    with elements: {", ".join(elements_in_cs)}\n'
+        ##
```
giovp, line 883:
```suggestion

```

spatialdata/_core/_transform_elements.py
```diff
@@ -102,31 +109,75 @@ def _transform_raster(
         else:
             assert n_spatial_dims == 3
             # raise NotImplementedError()
-    return transformed_dask
+    return transformed_dask, translation
+
+
+def _prepend_transformation(
```
giovp, line 115: imho this should have a docstring or more comments even if private

spatialdata/_core/transformations.py
```diff
@@ -197,10 +197,40 @@ def _get_n_spatial_dims(self, axes: tuple[str, ...]) -> int:
...
+            If True, each transformation of the transformed element will be prepended with the inverse of the current
+            transformation. In this way the data is transformed but the global positioning is maintained. An use case
+            changing the orientation/scale/etc of the data but keeping the alignment of the data within each coordinate
+            system. Please see notes for more details of how this parameter interact with xarray.DataArray for raster
+            data.
+
+        Returns
+        -------
+        SpatialElement
+            Transformed spatial element.
```
giovp, line 218: this will mace read the docs fail I think, it should be ``SpatialElement``: the transformed spatial element.

spatialdata/_core/transformations.py
```diff
@@ -197,10 +197,40 @@ def _get_n_spatial_dims(self, axes: tuple[str, ...]) -> int:
...
+        Parameters
+        ----------
+        element
+            Spatial element to transform.
+        maintain_positioning
+            If True, each transformation of the transformed element will be prepended with the inverse of the current
+            transformation. In this way the data is transformed but the global positioning is maintained. An use case
+            changing the orientation/scale/etc of the data but keeping the alignment of the data within each coordinate
+            system. Please see notes for more details of how this parameter interact with xarray.DataArray for raster
+            data.
```
giovp, line 213: can we come up with a better name of this? maybe `keep_alignment`?

giovp (changes requested):
minor things mostly around docstrings

spatialdata/_core/transformations.py
kevinyamauchi: or maybe something like `update_transformations`? It would be helpful to explain what happens when set to False as well.

kevinyamauchi (approved):
Overall, this looks good to me. I agree with @giovp's suggestions below. I think it's good to go once you have addressed them.

spatialdata/_core/_spatialdata.py
LucaMarconato: done, and added tests for the `_locate_spatial_element()` method.

spatialdata/_core/_spatialdata.py
LucaMarconato: Done

spatialdata/_core/_transform_elements.py
LucaMarconato: done

spatialdata/_core/transformations.py
LucaMarconato: changed

spatialdata/_core/transformations.py
LucaMarconato: Updated the docstring. Regarding the name @giovp I think that `keep_alignment` is slightly wrong because it's not a about the alignment of an element with respect to another but the positioning of the element in a coordinate system (this parameter may be set to true even with a spatialdata object with just one element). Anyway, I made the docstring cleared by replacing "global positioning" to "positioning in any coordinate system". I keep the name like this so we can merge, but feel free to open a issue about it or change it in a follow up PR if you find a better name.

LucaMarconato: @kevinyamauchi please see above regarding the name `maintain_positioning`. I think the name `update_transformation` is also not working because the transformation is updated in any case when translating a raster data (since we don't translate the xarray coordinates), or when rotating an image. Anyway, I created now this issue for changing the name in the future. https://github.com/scverse/spatialdata/issues/148

PR #143, opened 2023-02-17
Add transformations to polygons and points bounding box query

kevinyamauchi:
Hey @LucaMarconato . Here are the transforms added to the bounding box query for points and polygons. This gets the main infrastructure in. We can extend the tests for a few corner cases in a follow up PR.

A couple of notes:
- the queries now return None if no data matches the query. This way the user doesn't have to figure out the specific way to check in an element is empty for each of the different types of elements.
- I had to reset the index on the DaskDataFrame when subsetting the points. Otherwise, the tables get misaligned during the transform because the index doesn't start from 0 and the index of the transformed points that are re-inserted do start from 0. This feels a bit hacky, but I think it's okay for now.

I won't be able to work on this tomorrow, so please feel free to take the PR over and directly push to it!

### DIFF
--- a/spatialdata/_core/_transform_elements.py
+++ b/spatialdata/_core/_transform_elements.py
@@ -5,7 +5,6 @@ from functools import singledispatch
 from typing import TYPE_CHECKING, Any
 
 import dask.array as da
-import dask.dataframe as dd
 import dask_image.ndinterp
 import numpy as np
 from anndata import AnnData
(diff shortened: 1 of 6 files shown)


spatialdata/_core/_transform_elements.py
```diff
@@ -197,7 +198,7 @@ def _(data: DaskDataFrame, transformation: BaseTransformation) -> DaskDataFrame:
     axes = get_dims(data)
     arrays = []
     for ax in axes:
-        arrays.append(data[ax].to_dask_array())
+        arrays.append(data[ax].to_dask_array(lengths=True))
```
kevinyamauchi, line 201: Without this, I was having issues with the chunk size being unknown for subsequent operations on the transformed points.

spatialdata/_core/_transform_elements.py
```diff
@@ -206,7 +207,7 @@ def _(data: DaskDataFrame, transformation: BaseTransformation) -> DaskDataFrame:
         indices = xtransformed["dim"] == ax
         new_ax = xtransformed[:, indices].data.flatten()
         # mypy says that from_array is not a method of DaskDataFrame, but it is
-        transformed[ax] = da.from_array(np.array(new_ax))  # type: ignore[attr-defined]
+        transformed[ax] = dd.from_dask_array(da.from_array(np.array(new_ax)))  # type: ignore[attr-defined]
```
kevinyamauchi, line 210: I think one is supposed to add a Series, not an array here.

spatialdata/_core/_transform_elements.py
LucaMarconato: thanks for the info

spatialdata/_core/_transform_elements.py
LucaMarconato: I changed this code, now it is fully lazy. I still use a dask array and not a series. I think it works both way, but if there will be performance problems or some other problems I will keep this in mind and maybe check again.

LucaMarconato:
I have reviewed this code and made mainly two changes:
* the transform points function is now lazy and doesn't convert back and forth to numpy arrays
* I realized that I could simplify the bounding box query for polygons (not points) by using shapely + geopandas APIs.

LucaMarconato (approved):

kevinyamauchi:
Thanks @LucaMarconato ! I went ahead and merged to keep things moving forward.

PR #145, opened 2023-02-20
[WIP] torch DataSet + utils

kevinyamauchi:
This PR adds a torch DataSet with some addition utils. Initially, this implements a spot ROI dataset for replicating [this squidpy](https://squidpy.readthedocs.io/en/stable/external_tutorials/tutorial_tf.html) example. The DataSet is implemented such that it is compatible with monai and pytorch lightning, which gives us access to a ton of tooling (e.g., multi-GPU training, tensorboard logging, learning rate schedulers).

This PR requires #132, #143, and image bounding box query with transforms to be merged.


kevinyamauchi:
I have created an example notebook showing how the Dataset and transforms work for the SpotCropDataset. We can now generate tiles such that they are compatible with monai, torchvision, and pytorch lightning, so we can add tons of augmentations, dataloader cacheing, multiGPU, etc.. Once #132 lands, I can update the spot centroid fetching to use the polygons item.

LucaMarconato:
minor features of this PR:
- [x] overloading of `__get_item__()` and `__set_item__()` for `SpatialData`.

LucaMarconato:
I have pushed a code that has still open todos and bugs to fix, but it is usable. An example of usage is in [this script from the sandbox](https://github.com/giovp/spatialdata-sandbox/blob/main/examples/xenium_and_visium.py) (which run like this shows a bug. But if you use it from the coordinate system `global` and only with visium data (not querying the xenium image with the visium cirlces), it will work. So it should be good enough for the deep learning example.

spatialdata/_core/_spatialdata.py
```diff
@@ -1198,6 +1198,53 @@ def _gen_elements(self) -> Generator[tuple[str, str, SpatialElement], None, None
...
+        -------
+        The element.
+        """
+        found = []
+        for _, element_name, element in self._gen_elements():
+            if element_name == item:
+                found.append(element)
+        assert len(found) <= 1
+        if len(found) == 1:
+            return found[0]
```
giovp, line 1219:
```suggestion
        for _, element_name, element in self._gen_elements():
            if element_name == item:
                return element
```

giovp: we have check for uniqueness, no need to get list and check length.

spatialdata/_core/data_extent.py
```diff
@@ -0,0 +1,35 @@
+"""This file contains functions to compute the bounding box describing the extent of a spatial element,
```
giovp, line 1: why this is not in the bounding box related module? I would put it there.

spatialdata/_dl/datasets.py
```diff
@@ -0,0 +1,130 @@
+from typing import Callable, Optional
```
giovp, line 1: nitpick: can we call this module different than _dl but something like `_dataloader` (although also not great). Mostly asking cause this could also be used to simply create a dataloader to then e./g. compute features on tiles, similar to https://squidpy.readthedocs.io/en/latest/api/squidpy.im.calculate_image_features.html#squidpy.im.calculate_image_features

spatialdata/_dl/datasets.py
```diff
@@ -0,0 +1,130 @@
...
+
+import numpy as np
+from geopandas import GeoDataFrame
+from multiscale_spatial_image import MultiscaleSpatialImage
+from spatial_image import SpatialImage
+from torch.utils.data import Dataset
```
giovp, line 7: I would make torch as optional depedency, therefore I think in __Init__ of this module or where the ImageTilesDataset is import, something like this would be needed

```python
try:
    from spatialdata._dl.datasets import ImageTilesDataset
except ImportError as e:
    _error: str | None = str(e)
else:
    _error = None
```

spatialdata/_core/_rasterize.py
```diff
@@ -433,12 +463,11 @@ def _(
     if schema == Labels2DModel or schema == Labels3DModel:
         kwargs = {"prefilter": False, "order": 0}
     elif schema == Image2DModel or schema == Image3DModel:
+        # kwargs = {"prefilter": True}
```
giovp, line 466:
```suggestion
```

spatialdata/_core/_rasterize.py
```diff
@@ -395,25 +414,36 @@ def _(
...
     half_pixel_offset = Translation([0.5, 0.5, 0.5], axes=("z", "y", "x"))
     sequence = Sequence(
         [
-            half_pixel_offset.inverse(),
+            # half_pixel_offset.inverse(),
             scale,
             translation,
             corrected_affine.inverse(),
-            half_pixel_offset,
+            # half_pixel_offset,
```
giovp, line 442: does this need to be commented out or can be removed?

spatialdata/_core/_rasterize.py
```diff
@@ -395,25 +414,36 @@ def _(
         min_coordinate=min_coordinate,
         max_coordinate=max_coordinate,
         target_sizes=target_sizes,
-        corrected_affine=corrected_affine,
+        target_coordinate_system=target_coordinate_system,
+        # corrected_affine=corrected_affine,
```
giovp, line 418:
```suggestion
```

giovp (approved):
@kevinyamauchi @LucaMarconato minor (maybe nitpick) comments.

one major one regarding examples. They have to be in `spatialdata-notebooks`, not here. Everything that is not API should stay there. Ok to have python files and not notebooks (although better notebooks) but would still move.

giovp:
> Current todos:
> 
> * [x]  wrong tiling result (wrong queried data and wrong content) when tiling a cropped multiscale image (probably due to this: [2 bugs with spatial cropping (with multiscale rasters and when missing the table) #178](https://github.com/scverse/spatialdata/issues/178))
> * [ ]  extend functionality to get tiles from the raw space (not just from the target space)
>   
>   * [ ]  when querying tiles from the raw space, allow to specify either units, either pixels. Not both
>   * [ ]  when querying tiles from the target space, allow to specify also only units (and infer pixels), or only pixels. This is possible only when the data is not a multiscale, otherwise the pixels could not be determined
> * [ ]  make tests

@LucaMarconato I would not extend functionality in this PR. Think priority wise is to add some minimal test and merge right away. We need to do the module conversion of the repo to start testing intersphinx for notebooks and documentation (and also this change will impact io/napari/plot so there will be lot of work to be done there as well.

LucaMarconato:
Ok, I created an issue to keep track of that. I will make the tests and ask for review (btw I am working on the Xenium + Visium data atm, but I am going to work on this PR right after).

spatialdata/_core/_rasterize.py
```diff
@@ -327,6 +335,40 @@ def _get_xarray_data_to_rasterize(
...
 
+def _get_corrected_affine_matrix(
+    data: Union[SpatialImage, MultiscaleSpatialImage],
+    axes: tuple[str, ...],
+    min_coordinate: ArrayLike,
+    max_coordinate: ArrayLike,
+    target_coordinate_system: str,
+) -> tuple[Affine, tuple[str, ...]]:
+    """
+    TODO: docstring
```
kevinyamauchi, line 346: Please at least add a docstring stating what the function does before merging

spatialdata/_dl/datasets.py
```diff
@@ -0,0 +1,130 @@
...
+
+class ImageTilesDataset(Dataset):
+    def __init__(
+        self,
+        sdata: SpatialData,
+        regions_to_images: dict[str, str],
+        tile_dim_in_units: float,
+        tile_dim_in_pixels: int,
+        target_coordinate_system: str = "global",
+        transform: Optional[Callable[[SpatialData], dict[str, SpatialImage]]] = None,
```
kevinyamauchi, line 33: please add a docstring. I think the input parameters aren't clear (e.g., `tile_dim_in_units` vs. `tile_dim_in_pixels`)

kevinyamauchi (commented):
This looks good to me! I can't approve because I opened the PR. For me, the main things to do before merging:
- make `torch` import optional
- make sure functions have docstrings (at least a description of what the function does)

LucaMarconato:
> one major one regarding examples. They have to be in spatialdata-notebooks, not here. Everything that is not API should stay there. Ok to have python files and not notebooks (although better notebooks) but would still move.

ok I deleted the folder `examples` and moved this to the sandbox. I'll made these example not to show things to other users but to debug/visually test the spatial query, rasterization and tiler. I am not using notebooks because I use these for debugging, setting breakpoint etc.

spatialdata/_core/_spatialdata.py
LucaMarconato: as of now, the user can always do this:

```
sdata.images['my_element'] = image
sdata.labels['my_element'] = label
```

this would not raise any error, but would make the `sdata` object inconsistent. I agree to remove the check for uniqueness here as you suggested. But we should think how to avoid the user to reuse the same name across elements.

Possibility 1:
make `.images`, `.labels` etc private and make the user call
```
sdata['my_element'] = image
```
the correct element type will be inferred by `get_schema()` called on `image`.

Possibility 2:
we turn `.images`, `.labels` private and we replace them with classes that control the `__set_item__` and `__get_item__` behavior.c

I'll open a issue on this so we don't have to worry now.

spatialdata/_core/data_extent.py
LucaMarconato: this and other functions that could populate this file are not used for spatial queries so I would not put them in `_spatial_query.py`. The complexity of that file increase when implementing non-bounding box queries, so I would keep this code in another place.

spatialdata/_dl/datasets.py
LucaMarconato: renamed

spatialdata/_dl/datasets.py
LucaMarconato: would you add `torch` somewhere in `pyproject.toml` or it would be responsibility of the user to install it properly? I would go for the second.

spatialdata/_core/_rasterize.py
LucaMarconato: I would keep it commented until this is addressed: https://github.com/scverse/spatialdata/issues/165

which we should do with medium priority

spatialdata/_core/_spatialdata.py
LucaMarconato: Issue opened here: https://github.com/scverse/spatialdata/issues/186

spatialdata/_core/_rasterize.py
LucaMarconato: added the docstring explaining why we need this function.

giovp:
@LucaMarconato could you also quickly re add mypy in CI, seems like it's skipped atm, wasn't aware of that
https://github.com/scverse/spatialdata/blob/main/.pre-commit-config.yaml

LucaMarconato:
@giovp restored mypy in ci, we have some problems with the installation

.pre-commit-config.yaml
timtreis, line 1: Need so specify `numpy` but leave out the `==1.22.0`

LucaMarconato:
Gonna merge and increase the coverage in a next pr.

Issue #149, opened 2023-02-23
On disk versioning feedback


ivirshup:
@giovp 

Two main points of feedback on how versioning is done for the on-disk format. One minor one. 

## Version the elements separately

I think it makes a lot of sense to version the elements separately. For example, I can see wanting to change how points are encoded, but only that. I don't think we need to bump the version number for tables, shapes, images, just because points changed.

## Use something other than `@type`

I think `@type` is a weird option for reserved keyword since it's not how either anndata or OME specifies the type of an element. How we can do "metadata conventions" in a consistent way has been discussed around zarr quite a few times (e.g. [ZEP 4](https://github.com/zarr-developers/zeps/pull/28)). I'm not completley settled on what should be done, but can suggest some alternatives.

I'll try and figure out which option I like the most ASAP.

### Use anndata's `encoding_type`, `encoding_version`

We could just use anndata's `encoding_type`, `encoding_version`.

**Advantages**

* consistency with anndata
* Could eventually hook into the anndata IO dispatch system

**Main disadvantage**

The spatial data table is a superset of an AnnData. So if we're now using the same key, what's the type? If it's `anndata`, the attributes won't be part of the spec. If it's `spatialdata-table` or `ngff:regions_table` it's overwriting that it's also an `anndata`.

That said, the anndata library could just register `ngff:regions_table` to be read in as an AnnData. This library could even do that registration with:

```python
from anndata._io.specs import _REGISTRY, IOSpec
from anndata._io.specs.methods import read_anndata
import zarr

_REGISTRY.register_read(zarr.Group, IOSpec("ngff:region_table", "0.5"))(read_anndata)
```

But that's only if you DON'T want to do something custom at read time, e.g. anything with the other attributes.

### Use some separate field

This could be something like `spatialdata-encoding-type`. I personally would like this over `@type`, since `@type` is a kinda taken convention, which nothing else here is 

### Use some separate name-spaced field

I think long term it would make sense to namespace these things. For example:

```
.attrs = {
	"encoding": {
		"spatialdata": {"type": "spatialdata-table", "0.1"},
		"anndata": {"type": "anndata", "version": "0.1.0"},
		"ngff": {"type": "table", "version": "0.5"}
	}
	...
}
```

Where spatialdata knows that it's a superset of "anndata", and to check for `attrs["encoding"]["anndata"]`.

While this would eventually require coordination, one could start doing it here. Would probably require more thought as I think this will actually be an issue for arrow tables.

[...]

PR #154, opened 2023-02-24
cleanup format

giovp:
addressed some of the points described in #149 

- split format version per element
  - except for images and labels that are grouped together, following ome-zarr format

- removed `@type` and replaced it with `encoding-type`

- added a test for for shapes format

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,7 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+line_length = 120
 
 [Makefile]
 indent_style = tab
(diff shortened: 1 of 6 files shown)


LucaMarconato (approved):
LGTM

ivirshup (approved):
> it means that without spatialdata that anndata in the zarr store cannot be read, is that correct?

Yeah, unless the encoding type was just `anndata`, or we added the registration in anndata.

-------

Overall looks good to me.

giovp:
> > it means that without spatialdata that anndata in the zarr store cannot be read, is that correct?
> 
> Yeah, unless the encoding type was just `anndata`, or we added the registration in anndata.

makes sense, maybe worth it once tables are part of the ngff spec

PR #155, opened 2023-02-25
Rasterize/resample (only images and labels)

LucaMarconato:


LucaMarconato:
Ready for review! Please run this function https://github.com/scverse/spatialdata/blob/2479b8b4c0c157100754286db68784fa4186c7ec/tests/_core/test_spatial_query.py#L364 for a napari exploration of spatial queries and rasterization.

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,383 @@
...
+
+def _compute_target_dimensions(
+    spatial_axes: tuple[str, ...],
+    min_coordinate: ArrayLike,
+    max_coordinate: ArrayLike,
+    target_unit_to_pixels: Optional[float],
+    target_width: Optional[float],
+    target_height: Optional[float],
+    target_depth: Optional[float],
+) -> tuple[float, float, Optional[float]]:
```
kevinyamauchi, line 50: I think this needs a docstring

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,383 @@
...
+    target_coordinate_system: str,
+    target_unit_to_pixels: Optional[float] = None,
+    target_width: Optional[float] = None,
+    target_height: Optional[float] = None,
+    target_depth: Optional[float] = None,
+) -> SpatialData:
+    from spatialdata import SpatialData
+
+    new_images = {}
+    for element_type in ["points", "images", "labels", "shapes"]:
```
kevinyamauchi, line 147: do we want to iterate through all element types? Since `rasterize` isn't implemented for most element types, this will often throw an error.

I think it might be better to just do images and labels and them emit a warning if there are any points/shapes.

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,383 @@
...
+    data: SpatialImage,
+    axes: tuple[str, ...],
+    min_coordinate: ArrayLike,
+    max_coordinate: ArrayLike,
+    target_coordinate_system: str,
+    target_unit_to_pixels: Optional[float] = None,
+    target_width: Optional[float] = None,
+    target_height: Optional[float] = None,
+    target_depth: Optional[float] = None,
+) -> SpatialImage:
```
kevinyamauchi, line 178: I think we need a docstring. I'm not sure if it's better here or where the single dispatch function is defined above.

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,383 @@
...
+    target_spatial_axes = get_spatial_axes(target_axes)
+    assert len(target_spatial_axes) == len(min_coordinate)
+    assert len(target_spatial_axes) == len(max_coordinate)
+    corrected_affine = affine.to_affine(input_axes=axes, output_axes=target_spatial_axes)
+
+    # get xdata
+    if isinstance(data, SpatialImage):
+        xdata = data
+        pyramid_scale = None
+    elif isinstance(data, MultiscaleSpatialImage):
```
kevinyamauchi, line 223: I think this would be more clear to separate `SpatialImage` and `MultiscaleSpatialImage` into different single dispatch functions.

You can wrap the things in lines 260 through 334 in a separate function that is called by both the `SpatialImage` and `MultiscaleSpatialImage` single dispatch functions.

tests/_core/test_rasterize.py
```diff
@@ -0,0 +1,44 @@
...
+            if "z" not in dims and "target_depth" in kwargs:
+                continue
+            rasterize(
+                raster,
+                axes=spatial_dims,
+                min_coordinate=[0] * len(spatial_dims),
+                max_coordinate=[1] * len(spatial_dims),
+                target_coordinate_system="global",
+                **kwargs,
+            )
```
kevinyamauchi, line 29: It would be nice if something were done here to verify that the rasterization was done correctly.

kevinyamauchi (commented):
I think this is looking good @LucaMarconato ! I've left some initial comments. I will review again once we merge #151 and rebase this PR

spatialdata/_core/_rasterize.py
LucaMarconato: added

spatialdata/_core/_rasterize.py
LucaMarconato: I have added the docstring. I have placed it where the single dispatch function is defined.

spatialdata/_core/_rasterize.py
LucaMarconato: I thought about it and I could 
* either go as you said for one single dispatch for images and one for multiscales, and have two functions wrapping the code to be reused: one until line 260, and one from line 334.
* either wrap the lines after "# get xdata" in a separate function and keep the rest in a unique single dispatch function. 

I prefer the second because execpt for the "# get xdata" part, the implementation is identical for single scale and multiscale (also is much more different regarding images vs labels than `SpatialImage` vs `MultiscaleSpatialImage`). I'll implement this.

LucaMarconato: done.

tests/_core/test_rasterize.py
LucaMarconato: I'll implement this for a case without rotations since, as shown in the example, there is a problem with the precise positioning of the rasterized image, probably due to the fact that in some parts of the code the pixels coordinates refer to the (upper left) corner, while in others refer to the center of the pixels. This half pixel offset is actually large when an image is low res and covers the same area of a high res image.

LucaMarconato: done

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,512 @@
...
+                target_width is not None,
+                target_height is not None,
+                target_depth is not None,
+            ]
+        )
+        == 1
+    )
+    assert set(spatial_axes) == {"x", "y"} or set(spatial_axes) == {"x", "y", "z"}
+    if "z" not in spatial_axes:
+        assert target_depth is None
```
kevinyamauchi, line 104: I think these asserts should return an explanation (e.g., `assert x> 0, "x must be greater than 0`) so the user knows how to fix their issue.

```suggestion
    assert (
        np.sum(
            [
                target_unit_to_pixels is not None,
                target_width is not None,
                target_height is not None,
                target_depth is not None,
            ]
        )
        == 1
    ), "you must specify only one of: target_unit_to_pixels, target_width, target_height, target_depth" 
    assert set(spatial_axes) == {"x", "y"} or set(spatial_axes) == {"x", "y", "z"}
    if "z" not in spatial_axes:
        assert target_depth is None, "you cannot specify a target depth for 2D data"
```

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,512 @@
...
+
+    x_index = spatial_axes.index("x")
+    y_index = spatial_axes.index("y")
+    w_bb = max_coordinate[x_index] - min_coordinate[x_index]
+    h_bb = max_coordinate[y_index] - min_coordinate[y_index]
+    assert w_bb > 0
+    assert h_bb > 0
```
kevinyamauchi, line 111:
```suggestion
    assert w_bb > 0, "all max_coordinate values must be greater than all min_coordinate values"
    assert h_bb > 0, "all max_coordinate values must be greater than all min_coordinate values"
```

spatialdata/_core/_rasterize.py
```diff
@@ -0,0 +1,512 @@
...
+    w_to_h_bb = w_bb / h_bb
+
+    d_bb = None
+    d_to_h_bb = None
+    if "z" in spatial_axes:
+        z_index = spatial_axes.index("z")
+        d_bb = max_coordinate[z_index] - min_coordinate[z_index]
+        assert d_bb > 0
```
kevinyamauchi, line 119:
```suggestion
        assert d_bb > 0, "all max_coordinate values must be greater than all min_coordinate values"
```

kevinyamauchi (approved):
Thanks, @LucaMarconato ! Some minor comments below. I think this is good to go!

There are a lot of code branches here. I think we need to make sure they are tested to avoid accidental regressions in the future. I think we can merge now and add them later. I have made an issue () to track

PR #156, opened 2023-02-25
added some docstrings to spatial data

EliHei2:
Not even close to complete, but just making the PR for things to be able to move forward!

### DIFF
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -144,7 +144,7 @@ class SpatialData:
         self, name: str, image: Union[SpatialImage, MultiscaleSpatialImage], overwrite: bool = False
     ) -> None:
         """Adds an image to the SpatialData object
-        
+
         Parameters
         ----------
         name : str
@@ -173,7 +173,7 @@ class SpatialData:
         self, name: str, labels: Union[SpatialImage, MultiscaleSpatialImage], overwrite: bool = False
     ) -> None:
         """Adds labels to the SpatialData object
-        
+
         Parameters
         ----------
         name : str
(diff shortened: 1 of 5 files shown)


spatialdata/_core/_spatialdata.py
```diff
@@ -200,6 +244,20 @@ def is_backed(self) -> bool:
 
     # TODO: from a commennt from Giovanni: consolite somewhere in a future PR (luca: also _init_add_element could be cleaned)
     def _get_group_for_element(self, name: str, element_type: str) -> zarr.Group:
+        """Get the group for an elemnt, creates a new one if the element doesn't exist
+
+        Parameters
+        ----------
+        name : str
```
giovp, line 251: in general, parameters should not have types as they are already present in signature. so this (and everything else) should just be `name`

spatialdata/_core/_spatialdata.py
```diff
@@ -200,6 +244,20 @@ def is_backed(self) -> bool:
...
+            name of the element
+        element_type : str
+            type of the element. Should be in ["images", "labels", "points", "polygons", "shapes"].
+
+        Returns
+        -------
+        zarr.Group
+            either the exisitng sub-group or a new one
```
giovp, line 259: same here, this one should just be
`:func:`zarr.hierarchy.group` following the definition in e.g. [here](https://zarr.readthedocs.io/en/stable/api/hierarchy.html)

giovp (changes requested):
hi @EliHei2 ,

thanks for adding docstrings to spatialdata, I left couple of comments that should be implemented in the rest of docstrings, unfortunately we do it slightly different than what you suggested (numpy style). It's usually a good practice to first take a look at how docs are defined for already existing functions/classes and copy from there.

thanks again for starting this!

LucaMarconato:
One more thing, please remove the file `spatialdata/_io/.ipynb_checkpoints/read-checkpoint.py` from the pr.

Issue #158, opened 2023-02-27
type and storage of `instance_key` issues


giovp:
This is to keep track of issues that might arise on the relationship between `instance_key` and the indexes of regions.

### Motivation
The link between a table and it's region object is specified by the tuple `("region", "region_key")` saved in `table.uns["spatialdata_attrs"]`. The types are the following
- `region: Union[str, Sequence[str]]`. It can contain a single region id or a sequence of region ids
- `region_key: Optional[str]`. In case of a single region_id, then it can be None

To link the single region instance, e.g. the cell, an additional key is passed: 
- `instance_key: Sequence[str, float, int]`

### Problems
- We don't check for the consistency of this mapping in spatial data, e.g.:
  - are all regions present in the table, may/should/must be in the Spatialdata? we are not clear about this
  - is the type of `instance_key` matching the type of the index/region in the respective element? e.g. for labels,, it should just be `int32` or `int64`, whereas for shapes, it could be `str` or `numeric`.
- we are not clear nor consistent on the way we name the regions and elements in the `spatialdata_attrs`

another thing to consider is that the `instance_key` can be just he index in a `Shapes` element, but it can never be an index in `table.obs`, because we can only have `str` and numeric indexes in anndata. This is also something we should be clearer about.

LucaMarconato:
I wrongly thought that `region_key` must be `None` when `region` is a string. I will update `_concatenate_tables()` accordingly.

TODO:
- [ ] fix `_concatenate_table()`

LucaMarconato:
Yeah there was this code in `Table.validate()` (git blame on me 😅). I'll remove it.
```
elif isinstance(attr["region"], str):
    assert attr["region_key"] is None
```

giovp:
> I wrongly thought that `region_key` had to be `None` when `region` is a string. I will update `_concatenate_tables()` accordingly.
> 
> TODO:
> 
> * [x]  fix `_concatenate_table()`. Done here: [fixed schema for table; fixed _concatenate_tables() #159](https://github.com/scverse/spatialdata/pull/159)

but this is in fact true! see what I wrote above

> region_key: Optional[str]. In case of a single region_id, then it can be None

so no need to ficx

LucaMarconato:
> region_key: Optional[str]. In case of a single region_id, then it can be None

[...]

PR #159, opened 2023-02-28
fixed schema for table; fixed _concatenate_tables()

LucaMarconato:
Addressed this comment: https://github.com/scverse/spatialdata/issues/158#issuecomment-1447223143.

Ready for review and being merged.

### DIFF
--- a/spatialdata/_core/_spatialdata_ops.py
+++ b/spatialdata/_core/_spatialdata_ops.py
@@ -8,6 +8,7 @@ import numpy as np
 from anndata import AnnData
 
 from spatialdata._core.models import TableModel
+from spatialdata._logging import logger
 
 if TYPE_CHECKING:
     from spatialdata._core._spatialdata import SpatialData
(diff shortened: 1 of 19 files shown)


giovp:
@LucaMarconato I think the logic that you outlined in the comment is correct, I confirmed it here

https://github.com/scverse/spatialdata/issues/158#issuecomment-1447767542

spatialdata/_core/models.py
```diff
@@ -629,7 +629,6 @@ def validate(
                 if "instance_key" not in attr:
                     raise ValueError("`instance_key` not found in `adata.uns['spatialdata_attr']`.")
             elif isinstance(attr["region"], str):
-                assert attr["region_key"] is None
```
giovp, line 632: this should not be removed, it's correct

spatialdata/_core/models.py
```diff
@@ -656,10 +655,6 @@ def parse(
             instance_key = attr[cls.INSTANCE_KEY]
 
         if isinstance(region, str):
-            if region_key is not None:
```
giovp, line 659: this should not be removed, it's correct

tests/conftest.py
```diff
@@ -251,14 +251,16 @@ def _get_table(
...
-        region_key = region_key or "annotated_region"
         instance_key = instance_key or "instance_id"
     adata = AnnData(RNG.normal(size=(100, 10)), obs=pd.DataFrame(RNG.normal(size=(100, 3)), columns=["a", "b", "c"]))
     if instance_key is not None:
+        # from today pycharm refuses to execute this line if I run a test...
         adata.obs[instance_key] = np.arange(adata.n_obs)
+        # adata.obs.__setitem__(instance_key, np.arange(adata.n_obs))
     if isinstance(region, str):
-        return TableModel.parse(adata=adata, region=region, instance_key=instance_key)
+        return TableModel.parse(adata=adata, region=region, region_key=region_key, instance_key=instance_key)
```
giovp, line 261: this is also correct and should be readded

tests/conftest.py
```diff
@@ -251,14 +251,16 @@ def _get_table(
     instance_key: Optional[str] = None,
 ) -> AnnData:
     if region is not None:
-        region_key = region_key or "annotated_region"
         instance_key = instance_key or "instance_id"
     adata = AnnData(RNG.normal(size=(100, 10)), obs=pd.DataFrame(RNG.normal(size=(100, 3)), columns=["a", "b", "c"]))
     if instance_key is not None:
+        # from today pycharm refuses to execute this line if I run a test...
```
giovp, line 257: unrelated but there are a bunch of lines of code or comments with # pycharm can or cannot do something around the repo. I would consider to remove them cause they clutter the code and introduce lines/asserts 😅 I would put code quality and consistency with higher priority than whether pycharm works or not in any specific case

giovp (changes requested):

tests/conftest.py
LucaMarconato: yeah I agree; I'll stop using the `mypy` extensions (it's anyway super slow), which is the cause of most of the comments. This one above is actually false, I'll remove it.

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -359,33 +361,34 @@ def _concatenate_tables(tables: list[AnnData]) -> Optional[AnnData]:
...
-            if isinstance(region, list):
-                all_region_keys.append(region_key)
+            if region_key is not None:
+                try:
+                    table.obs[MERGED_TABLES_REGION_KEY] = table.obs[region_key]
+                except KeyError as e:
+                    logger.error(
+                        f"The table has a region_key ({region_key}), but the column with that name is not present in the table"
+                    )
+                    raise e
```
ivirshup, line 380: This should probably be an actual error, not from a logger

kevinyamauchi:
Based on [this comment](https://github.com/scverse/spatialdata/issues/158#issuecomment-1448498092), I think this PR should be updated with the following behavior:

- region_key should always be present, even if region is str
- instance_key type should be consistent across regions. 


I think the filter method can be a follow up PR so that we can merge this quickly and @giovp can get going on the demo datasets.

LucaMarconato:
This PR can be closed without merging as the required changes have been addressed by https://github.com/scverse/spatialdata/pull/164

PR #162, opened 2023-02-28
Feature/polygon aggregations

ivirshup:
Starting off polygon aggregations.

- [ ] Figure out why `"count"` isn't working exactly how I expect
- [ ] Test for circles
- [ ] Check if polygon tests are the same
- [ ] Handle coordinate spaces (just error?)

### DIFF
--- a/spatialdata/_core/aggregate.py
+++ b/spatialdata/_core/aggregate.py
@@ -1,0 +1,158 @@
+from __future__ import annotations
+
+import anndata as ad
+import dask.dataframe as ddf
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from scipy import sparse
+
+import spatialdata as sd
+from spatialdata._core.models import PointsModel, ShapesModel, get_schema
+
+
+def aggregate(
+    values: ddf.DataFrame | gpd.GeoDataFrame,
+    by: gpd.GeoDataFrame,
+    id_key: str,
+    *,
+    value_key: str | None = None,
+    agg_func: str = "mean",
+) -> ad.AnnData:
+    """
+    Aggregate values by given shapes
+
+    Parameters
+    ----------
+    values
+        Values to aggregate
+    by
+        Shapes to aggregate by
+    id_key
+        Key to group values by. This is the key in the shapes dataframe.
+    value_key
+        Key to aggregate values by. This is the key in the values dataframe.
+        If nothing is passed here, assumed to be 1.
+    agg_func
+        Aggregation function to apply over point values, e.g. "mean", "sum", "count".
+        Passed to pandas.DataFrame.groupby.agg.
+    """
+    # TODO: Check that values are in the same space
+    # Dispatch
+    by_type = get_schema(by)
+    values_type = get_schema(values)
+    if by_type is ShapesModel:
+        if values_type is PointsModel:
+            return _aggregate_points_by_shapes(values, by, id_key, value_key=value_key, agg_func=agg_func)
+        elif values_type is ShapesModel:
+            return _aggregate_shapes_by_shapes(values, by, id_key, value_key=value_key, agg_func=agg_func)
+        else:
+            raise NotImplementedError(f"Cannot aggregate {values_type} by {by_type}")
+    else:
+        raise NotImplementedError(f"Cannot aggregate {values_type} by {by_type}")
+
+
+def _aggregate_points_by_shapes(
+    points: ddf.DataFrame | pd.DataFrame,
+    shapes: gpd.GeoDataFrame,
+    id_key: str,
+    *,
+    value_key: str | None = None,
+    agg_func: str = "count",
+) -> ad.AnnData:
+    # Have to get dims on dask dataframe, can't get from pandas
+    dims = sd.get_dims(points)
+    if isinstance(points, ddf.DataFrame):
+        points = points.compute()
+    points = gpd.GeoDataFrame(points, geometry=gpd.points_from_xy(*[points[dim] for dim in dims]))
+    print(points.geometry, shapes.geometry)
+    return _aggregate(points, shapes, id_key, value_key, agg_func)
+
+
+def _aggregate_shapes_by_shapes(
+    values: gpd.GeoDataFrame,
+    by: gpd.GeoDataFrame,
+    id_key: str,
+    *,
+    value_key: str | None = None,
+    agg_func: str = "count",
+) -> ad.AnnData:
+    def circles_to_polygons(df: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
+        # We should only be buffering points, not polygons. Unfortunately this is an expensive check.
+        values_geotypes = list(values.geom_type.unique())
+        if values_geotypes == ["Point"]:
+            df = df.buffer(df[ShapesModel.RADIUS_KEY])
+        elif "Point" in values_geotypes:
+            raise TypeError("Geometry contained shapes and polygons.")
+        return df
+
+    values = circles_to_polygons(values)
+    by = circles_to_polygons(by)
+
+    return _aggregate(values, by, id_key, value_key, agg_func)
+
+
+def _aggregate(
+    value: gpd.GeoDataFrame,
+    by: gpd.GeoDataFrame,
+    id_key: str,
+    value_key: str | None = None,
+    agg_func: str = "count",
+) -> ad.AnnData:
+    """
+    Aggregate points by polygons.
+
+    Params
+    ------
+    value
+        GeoDataFrame of points to aggregate
+    by
+        GeoDataFrame of polygons
+    id_key
+        Column in values that indicate value type, e.g. probe id or organelle type
+    value_key
+        Column in values that indicate point value, e.g. probe intensity. If not provided, uses a fill value of 1.
+    agg_func
+        Aggregation function to apply over point values, e.g. "mean", "sum", "count".
+        Passed to pandas.DataFrame.groupby.agg.
+    """
+    assert pd.api.types.is_categorical_dtype(value[id_key]), f"{id_key} must be categorical"
+
+    if by.index.name is None:
+        by.index.name = "cell"
+    by_id_key = by.index.name
+
+    joined = by.sjoin(value)
+
+    if value_key is None:
+        point_values = np.broadcast_to(True, joined.shape[0])
+        value_key = "count"
+    else:
+        point_values = joined[value_key]
+
+    to_agg = pd.DataFrame(
+        {
+            by_id_key: joined.index,
+            id_key: joined[id_key].values,
+            value_key: point_values,
+        }
+    )
+
+    aggregated = to_agg.groupby([by_id_key, id_key]).agg(agg_func).reset_index()
+    obs_id_categorical = pd.Categorical(aggregated[by_id_key])
+
+    X = sparse.coo_matrix(
+        (
+            aggregated[value_key].values,
+            (obs_id_categorical.codes, aggregated[id_key].cat.codes),
+        ),
+        shape=(len(obs_id_categorical.categories), len(joined[id_key].cat.categories)),
+    ).tocsr()
+
+    adata = ad.AnnData(
+        X,
+        obs=pd.DataFrame(index=obs_id_categorical.categories),
+        var=pd.DataFrame(index=joined[id_key].cat.categories),
+    )
+
+    return adata
--- a/tests/_core/test_aggregations.py
+++ b/tests/_core/test_aggregations.py
@@ -1,0 +1,60 @@
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+import shapely
+
+from spatialdata._core.aggregate import aggregate
+from spatialdata._core.models import PointsModel, ShapesModel
+
+
+def test_aggregate_points_by_polygons():
+    points = PointsModel.parse(
+        pd.DataFrame(
+            {
+                "x": [1.2, 2.3, 4.1, 6.0, 6.1, 8.0, 9.0],
+                "y": [3.5, 4.8, 7.5, 4.0, 9.0, 5.5, 9.8],
+                "gene": list("aaabbbb"),
+            }
+        ),
+        coordinates={"x": "x", "y": "y"},
+        feature_key="gene",
+    )
+    shapes = ShapesModel.parse(
+        gpd.GeoDataFrame(
+            geometry=[
+                shapely.Polygon([(0.5, 7.0), (4.0, 2.0), (5.0, 8.0)]),
+                shapely.Polygon([(3.0, 8.0), (7.0, 2.0), (10.0, 6.0), (7.0, 10.0)]),
+            ],
+            index=["shape_0", "shape_1"],
+        )
+    )
+
+    result_adata = aggregate(points, shapes, "gene", agg_func="sum")
+    assert result_adata.obs_names.to_list() == ["shape_0", "shape_1"]
+    assert result_adata.var_names.to_list() == ["a", "b"]
+    np.testing.assert_equal(result_adata.X.A, np.array([[2, 0], [1, 3]]))
+
+
+def test_aggregate_polygons_by_polygons():
+    cellular = ShapesModel.parse(
+        gpd.GeoDataFrame(
+            geometry=[
+                shapely.Polygon([(0.5, 7.0), (4.0, 2.0), (5.0, 8.0)]),
+                shapely.Polygon([(3.0, 8.0), (7.0, 2.0), (10.0, 6.0), (7.0, 10.0)]),
+            ],
+            index=["shape_0", "shape_1"],
+        )
+    )
+    subcellular = ShapesModel.parse(
+        gpd.GeoDataFrame(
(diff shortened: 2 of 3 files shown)


spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,157 @@
...
+
+    joined = by.sjoin(value)
+
+    if value_key is None:
+        point_values = np.broadcast_to(True, joined.shape[0])
+        value_key = "count"
+    else:
+        point_values = joined[value_key]
+
+    to_agg = pd.DataFrame(
```
kevinyamauchi, line 132: I think it would be nice to factor out the part of this function that generates the `to_agg` DataFrame (i.e., lines 118-132). I think we will likely want this for aggregating tables that annotate the regions.

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,157 @@
...
+    Params
+    ------
+    value
+        GeoDataFrame of points to aggregate
+    by
+        GeoDataFrame of polygons
+    id_key
+        Column in values that indicate value type, e.g. probe id or organelle type
+    value_key
+        Column in values that indicate point value, e.g. probe intensity. If not provided, uses a fill value of 1.
```
kevinyamauchi, line 113: I find these docstrings a bit confusing. It might be helpful to say which table each key should belong to (e.g., `id_key` is the column in `value` that indicates the value type)

kevinyamauchi (commented):
I think this is looking good! It would be nice to be able to aggregate the tables annotating regions as well. 

For example, in the case of visium + xenium analysis, it would be nice to be able to aggregate the nuclei by the visium spots and then get the counts of the different classes of nuclei in each spot. That is aggregate the region table annotating the nuclei Shapes element by visium spots.

Do you think that will be possible using a similar approach? Perhaps you can use the `to_agg` table (see comment below) to aggregate features on a table?

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,157 @@
...
+        Aggregation function to apply over point values, e.g. "mean", "sum", "count".
+        Passed to pandas.DataFrame.groupby.agg.
+    """
+    assert pd.api.types.is_categorical_dtype(value[id_key]), f"{id_key} must be categorical"
+
+    if by.index.name is None:
+        by.index.name = "cell"
+    by_id_key = by.index.name
+
+    joined = by.sjoin(value)
```
kevinyamauchi, line 124: I _think_ this doesn't matter when the predicate is `intersects`, but I would have intuitively thought we want to do `value.sjoin(by)` so that each row is annotated by which `by` instance it is inside of. Is there any advantage to doing either order?

spatialdata/_core/aggregate.py
ivirshup: Can you give me some specific examples of use-cases here? I'm not sure I completley understand what you're asking for.

I'm wondering if this can be solved by joining the table here. Or if this isn't something that would be exposed in this PR, if it's something that can be factored out whenever this other feature is developed.

spatialdata/_core/aggregate.py
ivirshup: I've improved the docstrings, but mostly consolidated them to the `aggregate` function. Let me know what you think.

giovp:
> One thing that came up for this (and discussed a bit with @giovp) is whether shapes objects can carry annotations, or if they have to have an associated table. Right now, it seems like they can carry their own values, which fits how this function is currently implemented, but I wasn't sure if this was always meant to be the case.

just to clarify on this. Right now, in 1 case for parsing, and all cases for validation, we don't check whether the shapes geopandas dataframe as other columns other than `geometry` and maybe `radius`. Nevertheless, as per design doc, regions (shapes/polygons and labels) should not carry their annotation, which should be instead in the table.

Now, think the question re allowing shapes geopandas to have other column that the one spatialdata wants, is up to us: do we want to be strict, and thus drop anything else, or do we want to be permissive: there may be some other annotations but we don't validate neither use/assume throughout repo (yet IO handles it and no data is lost). Currently, the permissive option is implemented.

> > For example, in the case of visium + xenium analysis, it would be nice to be able to aggregate the nuclei by the visium spots and then get the counts of the different classes of nuclei in each spot. That is aggregate the region table annotating the nuclei Shapes element by visium spots.

> I think this makes sense. In this case you need the whole spatialdata object, though I would also like to support being able to do these aggregation on the elements by themselves I would suggest overloading this function, or having two functions.

I agree this is part of the P0 aggregation I think we discussed. In terms of API, agree that having the full spatialdata object is more desirable, but not super sure about this.

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+        Column in value dataframe to group values by.
+    value_key
+        Column in value dataframe to perform aggregation on.
+    agg_func
+        Aggregation functio to apply over grouped values. Passed to pandas.DataFrame.groupby.agg.
+    """
+    assert pd.api.types.is_categorical_dtype(value[id_key]), f"{id_key} must be categorical"
+
+    if by.index.name is None:
+        by.index.name = "cell"
```
giovp, line 128: not sure `"cell"` is a good option here. E.g. in the visium/xenium example they are not cells but spot. I'd rather not use this info at all, or use a generic "index" (and still maybe not hardcode it but allowing user to pass it).

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+        Geopandas dataframe to be aggregated. Must have a geometry column.
+    by
+        Geopandas dataframe to group values by. Must have a geometry column.
+    id_key
+        Column in value dataframe to group values by.
+    value_key
+        Column in value dataframe to perform aggregation on.
+    agg_func
+        Aggregation functio to apply over grouped values. Passed to pandas.DataFrame.groupby.agg.
+    """
```
giovp, line 124: Returns is missing

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+) -> ad.AnnData:
+    """
+    Inner function to aggregate geopandas objects.
+
+    See docstring for `aggregate` for semantics.
+
+    Parameters
+    ----------
+    value
+        Geopandas dataframe to be aggregated. Must have a geometry column.
```
giovp, line 115:
```suggestion
        :class:`geopandas.GeoDataFrame` to be aggregated. Must have a geometry column.
```

giovp: also below, this would render nicely in docs with intersphinx

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+    by
+        Geopandas dataframe to group values by. Must have a geometry column.
+    id_key
+        Column in value dataframe to group values by.
+    value_key
+        Column in value dataframe to perform aggregation on.
+    agg_func
+        Aggregation functio to apply over grouped values. Passed to pandas.DataFrame.groupby.agg.
```
giovp, line 123:
```suggestion
        Aggregation functio to apply over grouped values. Passed to :meth:`pandas.DataFrame.groupby.agg`.
```

giovp (changes requested):

giovp:
just minor things on docs

spatialdata/_core/aggregate.py
ivirshup: Could do `"index"`. Open to other names as well. Note that this only is used if the index doesn't already have a name, so I think that would cover being "user passed".

spatialdata/_core/aggregate.py
ivirshup: I'll accept it, but this isn't exported so won't be rendered. Plus the intersphinx should show up on the type annotation, right?

spatialdata/_core/aggregate.py
ivirshup: I'm actually going to ignore these, since they are changes to private docs that won't be rendered.

spatialdata/_core/aggregate.py
ivirshup: I'm actually going to ignore these, since they are changes to private docs that won't be rendered.

(As discussed in person)

ivirshup:
> Now, think the question re allowing shapes geopandas to have other column that the one spatialdata wants, is up to us: do we want to be strict, and thus drop anything else, or do we want to be permissive: there may be some other annotations but we don't validate neither use/assume throughout repo (yet IO handles it and no data is lost). Currently, the permissive option is implemented.

I think being strict, and going with the design doc, would be okay. One big benefit is "there is one way to do things".

However, this could make the API need some more thought. For `aggregate(sdata: SpatialData, values: Shape, by: Shape)` we're pretty strongly requiring that `values` is annotated by `sdata.table`. So does this signature make sense, since`values` is basically implicit?

I've put this on the agenda for tomorrow's meeting

LucaMarconato:
> But also the specific case given with nuclei and cells could be a problem. E.g. how do you have annotation on both cells and nuclei in one SpatialData object? I think you'd need two (which is fine, probably means we need the container class).

In the [xenium + visium notebook](https://github.com/scverse/spatialdata-notebooks/blob/main/notebooks/xenium_and_visium.md) ([rendered version here](https://drive.google.com/file/d/11sZz2zOoK4bqR8Dy32oAzFSh0gg40tws/view?usp=share_link)), I keep the data in two separate `SpatialData` objects. One for Xenium and one for Visium. The APIs then would need to be called with just one SpatialData object in the function call (we need the `sdata` object that contains the regions being aggregated, say the cells, while for the regions that define the aggergation area, the element is sufficient).

LucaMarconato:
> Currently, the permissive option is implemented.

I also think that the current implementation is good.

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+        For points, this could be probe intensity.
+    agg_func
+        Aggregation function to apply over point values, e.g. "mean", "sum", "count".
+        Passed to pandas.DataFrame.groupby.agg.
+
+    Returns
+    -------
+    AnnData of shape (by.shape[0], values[id_key].nunique())])
+    """
+    # TODO: Check that values are in the same space
```
LucaMarconato, line 46: Each element could be associated to multiple spaces. I would add `target_coordinate_system` parameter (like done for the query APIs), and then use the spatial alignment defined in that space. See for instance here: https://github.com/scverse/spatialdata/blob/eb99aebbdb4f7b7ee69d9663c5c996f5fa8f0440/spatialdata/_core/_spatial_query.py#L119

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+    See docstring for `aggregate` for semantics.
+
+    Parameters
+    ----------
+    value
+        Geopandas dataframe to be aggregated. Must have a geometry column.
+    by
+        Geopandas dataframe to group values by. Must have a geometry column.
+    id_key
+        Column in value dataframe to group values by.
```
LucaMarconato, line 119: Do you mean "column in by dataframe"?

ivirshup: No, this is from `values`. This is so you can do things like count transcripts (`values`) by gene (`values[id_key]`) by cell (`by`).

spatialdata/_core/aggregate.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+def aggregate(
+    values: ddf.DataFrame | gpd.GeoDataFrame,
+    by: gpd.GeoDataFrame,
+    id_key: str,
+    *,
+    value_key: str | None = None,
+    agg_func: str = "mean",
+) -> ad.AnnData:
```
LucaMarconato, line 21: It would be cool to set `region`, `region_key` and `instance_key` in the returned `adata`, but `region` and `region_key` require a `sdata` in the function call, so here only `instance_key` can be set.

spatialdata/_core/aggregate.py
LucaMarconato: Ah got it! In the case of points in the model we have a `FEATURE_KEY` that tells which is the preferred aggregation column (see discussion here https://github.com/scverse/spatialdata/pull/113#issuecomment-1419502673), so I would put `id_key` optional and use that by default in the case of points, if available.

LucaMarconato:
I have to read the code more carefully, but for the moment I describe some use cases, so that you can check if they are already covered.

Say that we have visium expression data (visium circles), and single cells spatial expression form an overlapping spatial region. Say that the cells have also a cell type column.

I describe some possibilities for querying the cells by the visium circles. Ideally the `aggregate` function could be used for all of these:
1. we count the cells inside the visium circles, no need for the table (so no need for the enclosing `sdata` object). We return an `anndata` object with a single column.
2. we count the fractions of cell types for each visium circle, we need to pass the table (so the `sdata` object for the cells) and we need to specify the (single) column for the cells that contain the cell types. We return an `anndata` object with many columns, one for each category of the cell-type column.
3. we sum/average the cell expression inside each visium circle, we need to pass the table (so the `sdata` object for the cells). There is no need to specify any column because all the columns are aggregated by default. We return an `anndata` object that has as many columns as the ones in the table for cell expression. `obsm` could be omitted by default, and could be aggergated for instance if the user passed `aggergate_obsm=True`.

From my understanding of the APIs, since `key_id` is not optional, I think that the case 3 is only partially covered. In case 3 I think that `key_id` should be `None`.

spatialdata/_core/aggregate.py
ivirshup: So, I was thinking this could return a `SpatialData` where the region is `by`.

spatialdata/_core/aggregate.py
ivirshup: I think we may need `target_coordinate_system` due to the current implementation of coordinate systems, but don't think this argument is actually needed, or at least shouldn't be required.

* I think the aggregation should occur in `value`'s intrinsic coordinate space. This is because this function broadly assumes there are more `value`s than `by`'s to group by.
* But right now we can't actually uniquely identify the intrinsic coordinate system for coordinate based values (shapes, points) https://github.com/scverse/spatialdata/issues/169

LucaMarconato: In the [Xenium + Visium example](https://drive.google.com/file/d/11sZz2zOoK4bqR8Dy32oAzFSh0gg40tws/view?usp=share_link), having the aggregation happening in the "intrinsic coordinate space" would be incorrect.

You can see this in the GIF after the text "The two images are now aligned in the aligned coordinate system."

LucaMarconato: Both Visium and Xenium data (which live in different spatialdata objects), map to the "global" coordinate system. But the data is not aligned in this coordinate system. The data is only aligned in the "aligned" coordinate system.

ivirshup: Yes, it would be incorrect if the intent was to have two different coordinate spaces, but they were both called `"global"`. But, also, that seems like an extremely error prone thing to do even if we're requiring explicit target spaces.

spatialdata/_core/aggregate.py
ivirshup: I think it's right that there are probably good implicit defaults here. I think this get's more complicated for other cases.

In essence, `id_key` is needed when we have a one-dimensional value to aggregate. But there cases where we have two dimensional values to aggregate. 

* Where `values` is an image, the values are channels so we'd likely the output to be (observations in by, channels in values) (we also have no need for `values_key` here)
* Where `values` is a Shapes or Labels SpatialData. Here, we could be aggregating `X`.

I think the values case can wait for now, but these are good things to think about. Especially for shapes how we aggregate values other than `X`.

spatialdata/_core/aggregate.py
LucaMarconato: I see, ok this is what happens: the `visium` and `cosmx` functions from the `spatialdata-io` library returns the data (that "lives in separate spaces") and gives the default name "global".

ivirshup: The `library_id` value could be reasonable defaults values for 10x coordinate space names, since we broadly assume these are unique anyways.

LucaMarconato: @giovp wdyt? Do you think it's possible to find a library id for each tech (not just visium and xenium)?

giovp: yeah it would be possible but only for xenium and visium afaik. I think it's more sensible to have a `default` cs that is called something like `"{element_name}_default"`.

@ivirshup regarding this
> But right now we can't actually uniquely identify the intrinsic coordinate system for coordinate based values (shapes, points)

also related to the github issue. I think this is not true, as the intrinsic is the one where we have `Identity`. I don't think there will ever be a case where there are 2 cs with both identity, but we could also just validate that if `Identity` is there, then there can be only one.

giovp (approved):
@ivirshup how's this looking? @EliHei2 did you manage to test this out in code? (sorry think I missed the zulip conversations.

to me it looks pretty well fleshed out and with tests, we could probably merge?

giovp:
closed by #196

PR #163, opened 2023-02-28
Use warnings

ivirshup:
Use warnings, not logger.warnings.

Basically "libraries warn, applications log", but also really good to be consistent about these things so we can catch them in a consistent way.


kevinyamauchi (approved):
I'm not sure if you're still converting logging - > warning, but I'm approving because the change makes sense to me.

ivirshup:
I think this was all

PR #164, opened 2023-03-01
update table model

giovp:
as discussed in #158 

- [x] introduce mandatory `region` `region_key` and `instance_key`
- [x] rewrite concatenate following implementation
- [x] support string index io for shapes
- [x] change type of radius to `Union[float, ArrayLike]`

changes
- [x] the empty table case is not supported anymore, we can revert to this behaviour later but for now don't see priority

things leaving out of this PR as discussed in #158 
-  validation of type of categorical in for instances and instances in labels (should be done in spatialdata)
reason:
- not high priority and postponing to later work to consolidate spatialdata validation (e.g. we need a way to consolidate relationships between elements).

### DIFF
--- a/spatialdata/_core/_spatialdata_ops.py
+++ b/spatialdata/_core/_spatialdata_ops.py
@@ -311,23 +311,21 @@ def _concatenate_tables(
         if instance_key is None:
             raise ValueError("`instance_key` must be specified if tables have different instance keys")
 
-    tables_l = []
-    regions_l = []
-    instance_keys_l = []
-    for k, i, regs, table in zip(region_keys, instance_keys, regions, tables):
-        if isinstance(regs, list):
-            for r in regs:
-                table_ = table[table.obs[k] == r]
-                tables_l.append(table_)
-                regions_l.append(r)
-                instance_keys_l.append(table_.obs[i].values.flatten())
-        else:
-            tables_l.append(table)
-            regions_l.append(regs)
-            instance_keys_l.append(table.obs[i].values.flatten())
+    from copy import copy  # Should probably go up at the top
 
-    merged_table = anndata.concat(tables_l, label=region_key, keys=regions_l, **kwargs)
-    merged_table.obs[instance_key] = np.concatenate(instance_keys_l, axis=0)
+    tables_l = []
+    for table_region_key, table_instance_key, table in zip(region_keys, instance_keys, tables):
+        rename_dict = {}
+        if table_region_key != region_key:
+            rename_dict[table_region_key] = region_key
+        if table_instance_key != instance_key:
+            rename_dict[table_instance_key] = instance_key
+        if len(rename_dict) > 0:
+            table = copy(table)  # Shallow copy
+            table.obs = table.obs.rename(columns=rename_dict, copy=False)
+        tables_l.append(table)
+
+    merged_table = ad.concat(tables_l, **kwargs)
     attrs = {
         TableModel.REGION_KEY: regions_l,
         TableModel.REGION_KEY_KEY: region_key,
--- a/spatialdata/_core/_spatialdata_ops.py
+++ b/spatialdata/_core/_spatialdata_ops.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional, Union
 
 import anndata
 import networkx as nx
 import numpy as np
-from anndata import AnnData
+import pandas as pd
 
 from spatialdata._core.models import TableModel
 
@@ -26,7 +26,6 @@ __all__ = [
     "get_transformation",
     "remove_transformation",
     "get_transformation_between_coordinate_systems",
-    "_concatenate_tables",
     "concatenate",
 ]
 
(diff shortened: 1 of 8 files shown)


giovp:
actually please don't merge as I forgot to add support for non-numeric indexes in geopandas io

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -291,154 +287,76 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+    **kwargs: Any,
+) -> SpatialData:
+    """
+    Concatenate a list of spatial data objects.
 
     Parameters
     ----------
     sdatas
         The spatial data objects to concatenate.
-    omit_table
```
LucaMarconato, line 422: I would keep this. Use case: we have a xenium and a visium dataset and we want to merge them keeping only the table of one of the two. To do so we can call `concatenate(..., omit_table=True)` and then do `sdata_concatenated.table = xenium.table`.

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -291,154 +287,76 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+        TableModel.INSTANCE_KEY: instance_key,
+    }
+    merged_table.uns[TableModel.ATTRS_KEY] = attrs
+
+    return TableModel().validate(merged_table)
+
+
+def concatenate(
+    sdatas: list[SpatialData],
+    region_key: str = "region_merged",
```
LucaMarconato, line 329: If all the `region_key` across the tables are identical, (`len(region_keys` == 1) I would use this value. So I would define `region_key` as `region_key: Optional[str] = None`.

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -291,154 +287,76 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+    instance_key: str = "instance_merged",
```
LucaMarconato, line 330: Same here for what just said for `region_key`.

spatialdata/_core/models.py
```diff
@@ -658,38 +667,35 @@ def parse(
...
+            region = region.tolist()
+        if region is None:
+            raise ValueError(f"`{cls.REGION_KEY}` must be provided.")
+        region_ = region if isinstance(region, list) else [region]
+        if TYPE_CHECKING:
+            assert isinstance(region_, list)
+        if not adata.obs[region_key].isin(region_).all():
+            raise ValueError(f"`adata.obs[{region_key}]` values do not match with `{cls.REGION_KEY}` values.")
+        if not is_categorical_dtype(adata.obs[region_key]):
+            logger.warning(f"Converting `{cls.REGION_KEY_KEY}: {region_key}` to categorical dtype.")
```
LucaMarconato, line 682: Isaac suggested to use `warn` instead of `logger.warning` across the repo: https://github.com/scverse/spatialdata/pull/163

spatialdata/_core/models.py
```diff
@@ -658,38 +667,35 @@ def parse(
...
         return adata
 
 
 # TODO: consider removing if we settle with geodataframe
-def _sparse_matrix_from_assignment(
+def _sparse_matrix_from_assignment(  # pragma: no cover
     n_obs: int, var_names: Union[list[str], ArrayLike], assignment: pd.Series
 ) -> csr_matrix:
     """Create a sparse matrix from an assignment array."""
+    from tqdm import tqdm
```
LucaMarconato, line 697: `tqdm` has been removed from the dependencies of the package. But here I would remove the function `_sparse_matrix_from_assignment()`. It's a legacy from when we where representing points with `AnnData`.

spatialdata/_core/_spatialdata_ops.py
giovp: what's the difference between manually adding the visium image and visium shapes to the xenium spatialdata?

LucaMarconato: A big one: if you add an image to the xenium data, the new layer is saved to disk. On the contrary, `concatenate()` creates a new in-memory `sdata` object that is not backed (but the elements are backed from their respective `sdata` object)

giovp: ok what about then simply passing one of the two `spatialdata` without the table? to me it just complicates the signature while the behavior is already impliclty implemetned

giovp:
@LucaMarconato once this is merged, can #159 be closed?

spatialdata/_core/_spatialdata_ops.py
giovp: > A big one: if you add an image to the xenium data, the new layer is saved to disk. On the contrary, concatenate() creates a new in-memory sdata object that is not backed (but the elements are backed from their respective sdata object)

this hidden behavior makes me also think we might want to be explicit about in memory/backed in concatenate

spatialdata/_io/write.py
```diff
@@ -267,9 +269,12 @@ def write_shapes(
     shapes_group.create_dataset(name="coords", data=coords)
     for i, o in enumerate(offsets):
         shapes_group.create_dataset(name=f"offset{i}", data=o)
-    # index cannot be string
-    # https://github.com/zarr-developers/zarr-python/issues/1090
-    shapes_group.create_dataset(name="Index", data=shapes.index.values)
+    if shapes.index.dtype.kind == "U" or shapes.index.dtype.kind == "O":
```
LucaMarconato, line 272: oh nice!

LucaMarconato:
I'll have a quick double check of the tests that I added in https://github.com/scverse/spatialdata/pull/159, but now mostly is covered here. Btw I reviewed this PR, if you it's finished for me you can merge.

giovp:
> I'll have a quick double check of the tests that I added in #159, but now mostly is covered here. Btw I reviewed this PR, if you it's finished for me you can merge.

I think it was mostly work on concatenate which has been re-added here.

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -291,154 +289,90 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
+                tables_l.append(table_)
+                regions_l.append(r)
+                instance_keys_l.append(table_.obs[i].values.flatten())
+        else:
+            tables_l.append(table)
+            regions_l.append(regs)
+            instance_keys_l.append(table.obs[i].values.flatten())
+
+    merged_table = anndata.concat(tables_l, label=region_key, keys=regions_l, **kwargs)
+    merged_table.obs[instance_key] = np.concatenate(instance_keys_l, axis=0)
```
ivirshup, line 330: Maybe something more like this?

```suggestion
    from copy import copy  # Should probably go up at the top

    tables_l = []
    for table_region_key, table_instance_key, table in zip(region_keys, instance_keys, tables):
        rename_dict = {}
        if table_region_key != region_key:
            rename_dict[table_region_key] = region_key
        if table_instance_key != instance_key:
            rename_dict[table_instance_key] = instance_key
        if len(rename_dict) > 0:
            table = copy(table)  # Shallow copy
            table.obs = table.obs.rename(columns=rename_dict, copy=False)
        tables_l.append(table)

    merged_table = ad.concat(tables_l, **kwargs)
```

* I found the current version took a little effort to figure out. If you want to keep it, it could use a comment explaining that it's making the column names consistent
* This should be more performant, as it doesn't require the subsetting, and the shallow copy should be close to free

The result is slightly different. The order of the observations in tables is preserved, and the result couldn't include the original region or instance keys. If you wanted the possibility of preserving all the original columns in an `outer` join, you could add copies of the instance and region key columns instead of renaming them.

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -291,154 +289,90 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
         return sdatas[0]
 
-    if not omit_table:
-        list_of_tables = [sdata.table for sdata in sdatas if sdata.table is not None]
-        merged_table = _concatenate_tables(list_of_tables)
-    else:
-        merged_table = None
+    merged_table = _concatenate_tables(
+        [sdata.table for sdata in sdatas if sdata.table is not None], region_key, instance_key, **kwargs
+    )
```
ivirshup, line 375: Could the Exeptions bellow this be `KeyError`s or `ValueError`s?

Also these checks are quite cheap, can they happen before the potentially expensive table concatenation?

spatialdata/_core/_spatialdata_ops.py
```diff
@@ -291,154 +289,90 @@ def _describe_paths(paths: list[list[Union[int, str]]]) -> str:
...
-        The concatenated spatial data object.
+    The concatenated :class:`spatialdata.SpatialData` object.
     """
-    from spatialdata._core._spatialdata import SpatialData
+    from spatialdata import SpatialData
 
-    assert type(sdatas) == list
-    assert len(sdatas) > 0
+    assert type(sdatas) == list, "sdatas must be a list"
+    assert len(sdatas) > 0, "sdatas must be a non-empty list"
```
ivirshup, line 369: Can we remove the special casing for the "one table passed" case?

It's weird to me that it:

* A reference to the object that was passed in. So any modification of the result would modify the original, breaking the semantics of concatenation.
* The "renaming" properties of `region_key` and `instance_key` arguments are not respected.

Tests for this case would be great.

kevinyamauchi (approved):
LGTM! Thanks!

spatialdata/_core/_spatialdata_ops.py
ivirshup: I would remove this. I don't think the feature being described fits the definition of concatenation. Maybe this is more of a `"merge"`?

Also, is super easy for a user to do themselves in the meantime.

<details>
<summary> example </summary>

```python
from functools import reduce
from operator import or_

sdata_w_table = ...
other_sdatas: list[SpatialData]

sdatas = [sdata_w_table, *other_sdatas]

new_sdata = SpatialData(
    table=sdata_w_table.table,
    images=reduce(or_, (sdata.images for sdata in sdatas)),
    labels=reduce(or_, (sdata.labels for sdata in sdatas)),
    shapes=reduce(or_, (sdata.shapes for sdata in sdatas)),
    points=reduce(or_, (sdata.points for sdata in sdatas)),
)
```

</details>

spatialdata/_core/models.py
```diff
@@ -645,6 +635,24 @@ def parse(
         region_key: Optional[str] = None,
         instance_key: Optional[str] = None,
     ) -> AnnData:
+        """
+        Parse the :class:`anndata.AnnData` to be compatible with the model.
+
+        Parameters
+        ----------
+        adata
+            :class:`anndata.AnnData`.
```
ivirshup, line 644: Shouldn't the type of the argument class already be included from the type signature?

```suggestion
```

spatialdata/_core/models.py
```diff
@@ -658,63 +666,30 @@ def parse(
...
-                raise ValueError(f"`{cls.REGION_KEY}` must be of type `str` or `List`.")
+        if region_key is None:
+            raise ValueError(f"`{cls.REGION_KEY_KEY}` must be provided.")
+        if isinstance(region, np.ndarray):
+            region = region.tolist()
+        if region is None:
+            raise ValueError(f"`{cls.REGION_KEY}` must be provided.")
+        region_ = region if isinstance(region, list) else [region]
+        if TYPE_CHECKING:
+            assert isinstance(region_, list)
```
ivirshup, line 677: Is this necessary?

If mypy really can't figure out the previous line, could you change it to `region_: list = ...` instead?

```suggestion
```

tests/_core/test_spatialdata_operations.py
```diff
@@ -115,35 +115,28 @@ def test_concatenate_tables():
...
+    with pytest.raises(ValueError, match="`region_key` must be specified if tables have different region keys"):
+        assert len(_concatenate_tables([table4, table5, table6])) == len(table4) + len(table5) + len(table6)
+    assert len(_concatenate_tables([table4, table5, table6], region_key="region")) == len(table4) + len(table5) + len(
+        table6
     )
-
-    assert len(_concatenate_tables([table4, table5])) == len(table4) + len(table5)
-
-    with pytest.raises(RuntimeError):
-        _concatenate_tables([table4, table6])
```
ivirshup, line 146: Why was this removed?

tests/_core/test_spatialdata_operations.py
```diff
@@ -115,35 +115,28 @@ def test_concatenate_tables():
...
     table5 = _get_table(
-        region=["shapes/circles", "shapes/poly"], region_key="annotated_shape0", instance_key="instance_id"
+        region=["shapes/circles2", "shapes/poly2"], region_key="annotated_shape0", instance_key="instance_id"
     )
     table6 = _get_table(
-        region=["shapes/circles", "shapes/poly"], region_key="annotated_shape1", instance_key="instance_id"
+        region=["shapes/circles3", "shapes/poly3"], region_key="annotated_shape1", instance_key="instance_id"
+    )
+    with pytest.raises(ValueError, match="`region_key` must be specified if tables have different region keys"):
+        assert len(_concatenate_tables([table4, table5, table6])) == len(table4) + len(table5) + len(table6)
```
ivirshup, line 136: Why is there an assertion in here?

The assertion is never executed, so I'm not seeing the point in including it.

ivirshup (commented):
Mostly good, a few suggestions and minor things.

spatialdata/_core/_spatialdata_ops.py
LucaMarconato: You are right

tests/_core/test_spatialdata_operations.py
LucaMarconato: my mistake

LucaMarconato:
Answering to the comment re `omit_table`, https://github.com/scverse/spatialdata/pull/164#discussion_r1124456600. Maybe a `merge` would be better here. Not sure. I would then remove it for now and then we see how to provide this functionality. For the moment in the Xenium + Visium notebook we can either use the `reduce()` approach (or equivalent) or never merge the two `sdata` objects (we can do most/all the operations keeping them separate actually).

ivirshup:
@giovp, actually one other (maybe major) point.

What happens if (for any of the tables) instance key is `None`?

LucaMarconato:
> What happens if (for any of the tables) instance key is None?
I see that this is not possible for a table created with the parser (see here https://github.com/scverse/spatialdata/blob/80dc1024d397a06120f9bd5b27237b9772a85997/spatialdata/_core/models.py#LL685-L686C65), but a table with `instance_key = None` would pass validation. I would update the validation to catch this case and raise an exception.

spatialdata/_core/_spatialdata_ops.py
giovp: you mean this case?
```
if len(sdatas) == 1:
        return sdatas[0]
```

spatialdata/_core/models.py
giovp: ok, think it'd be nice to also implement a docstring processor likw we have in squidpy

tests/_core/test_spatialdata_operations.py
giovp: redundant with the new one

giovp:
thanks all for feedback! I accepted all suggestions and made couple of modifications
- explicitly check for regions in uns and obs, e.g. (spotted one bug in filter coordinate system @LucaMarconato but fixed it)

```python
set(adata.uns["spatialdata_attrs"]["regions"] ) == set(adata.obs.region_key.unqiue())
# True
```

- explicitly check for all spatialdata_attrs to be valid (e.g. value of `instance_key` cannot be None).
- accepted code suggestion by @ivirshup but added additional check on unique region names that suggested version missed, e.g.:
```python
sdata0 = SpatialData(labels={"sample1":..}, table}
sdata1 = SpatialData(labels={"sample1":..}, table}
concatenate(sdata0, sdata1)
# now throws error, as there is no way to understand to which obs refers which region.
```
the latter is something that ideally we will change, but it'd require keeping track of region names and rename them in the resulting spatialdata object. If tests pass I'll go on and merge

Issue #167, opened 2023-03-03
Query operations should subset the table


LucaMarconato:
we should add a parameter `query(filter_table: bool)` to subset also the table when doing a query operation.

PR #170, opened 2023-03-06
added filter_table param to bounding box query

LucaMarconato:
Small PR closing https://github.com/scverse/spatialdata/issues/167 as asked by @EliHei2:

when performing a bounding box query on a spatialdata object, it is now possible to filter the table to contain only the rows that are annotating any of the elements in the resulting queried object.

### DIFF
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -1133,7 +1133,7 @@ class QueryManager:
         target_coordinate_system
             The coordinate system the bounding box is defined in.
         filter_table
-            If True, the table is filtered to only contain rows that are annotating individual regions contained in the bounding box.
+            If True, the table is filtered to only contain rows that are annotating regions contained within the bounding box.
         Returns
         -------
         The SpatialData object containing the requested data.
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -1150,11 +1150,12 @@ class QueryManager:
             filter_table=filter_table,
         )
 
-    def __call__(self, request: BaseSpatialRequest) -> SpatialData:
+    def __call__(self, request: BaseSpatialRequest, **kwargs) -> SpatialData:  # type: ignore[no-untyped-def]
         from spatialdata._core._spatial_query import BoundingBoxRequest
 
         if isinstance(request, BoundingBoxRequest):
-            # TODO: filter table is on by default, shall we put filter_table inside the request?
-            return self.bounding_box(**request.to_dict(), filter_table=True)
+            # TODO: request doesn't contain filter_table. If the user doesn't specify this in kwargs, it will be set
+            #  to it's default value. This could be a bit unintuitive and we may want to change make things more explicit.
+            return self.bounding_box(**request.to_dict(), **kwargs)
         else:
             raise TypeError("unknown request type")
--- a/spatialdata/_core/_spatialdata.py
+++ b/spatialdata/_core/_spatialdata.py
@@ -1118,14 +1118,22 @@ class QueryManager:
         min_coordinate: ArrayLike,
         max_coordinate: ArrayLike,
         target_coordinate_system: str,
+        filter_table: bool = True,
     ) -> SpatialData:
         """Perform a bounding box query on the SpatialData object.
 
         Parameters
         ----------
-        request
-            The bounding box request.
-
+        axes
+            The axes the bounding box is defined in.
+        min_coordinate
+            The minimum coordinates of the bounding box.
+        max_coordinate
+            The maximum coordinates of the bounding box.
+        target_coordinate_system
+            The coordinate system the bounding box is defined in.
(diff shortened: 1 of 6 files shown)


spatialdata/_core/_spatialdata.py
```diff
@@ -1118,14 +1118,22 @@ def bounding_box(
...
+        axes
+            The axes the bounding box is defined in.
+        min_coordinate
+            The minimum coordinates of the bounding box.
+        max_coordinate
+            The maximum coordinates of the bounding box.
+        target_coordinate_system
+            The coordinate system the bounding box is defined in.
+        filter_table
+            If True, the table is filtered to only contain rows that are annotating individual regions contained in the bounding box.
```
kevinyamauchi, line 1136:
```suggestion
            If True, the table is filtered to only contain rows that are annotating regions contained within the bounding box.
```

spatialdata/_core/_spatialdata.py
```diff
@@ -1139,12 +1147,14 @@ def bounding_box(
...
+            filter_table=filter_table,
         )
 
     def __call__(self, request: BaseSpatialRequest) -> SpatialData:
         from spatialdata._core._spatial_query import BoundingBoxRequest
 
         if isinstance(request, BoundingBoxRequest):
-            return self.bounding_box(**request.to_dict())
+            # TODO: filter table is on by default, shall we put filter_table inside the request?
+            return self.bounding_box(**request.to_dict(), filter_table=True)
```
kevinyamauchi, line 1158: I think it makes sense to add `filter_table` to the request

kevinyamauchi (approved):
LGTM! Thanks!

spatialdata/_core/_spatialdata.py
LucaMarconato: Ah forgot to specify: `filter_table` would be used only when the queried object is a `sdata`, and it would be ignored when we query an element. This is why I didn't add it in the request class. Shall I still add it?

tests/_core/test_spatial_query.py
```diff
@@ -348,3 +352,33 @@ def test_bounding_box_spatial_data(full_sdata):
         d = get_transformation(element, get_all=True)
         new_d = {k.replace("global", "cropped"): v for k, v in d.items()}
         set_transformation(element, new_d, set_all=True)
+
+
+def test_bounding_box_filter_table():
+    ##
```
giovp, line 358:
```suggestion
```

spatialdata/_core/_spatialdata.py
kevinyamauchi: Ah yeah, good point. Let's not add `filter_table` to the request. Perhaps we can make `filter_table` a kwarg on `__call__`? Alternatively, we can just punt and deal with it later. I'm fine either way!

tests/_core/test_models.py
```diff
@@ -53,7 +53,7 @@
 RNG = default_rng()
 
 # should be set to False for pre-commit and CI; useful to set to True for are fixing/debugging tests
-SHORT_TESTS = False
+SHORT_TESTS = True
```
giovp, line 56: wonder if this is gonna produce inconsistencies across CI runs when it's left on. I think an option would be to add custom marks https://docs.pytest.org/en/7.1.x/example/markers.html

e.g. add `@pytest.mark.fast()` as decorator and then use `pytest -v -m fast`

tests/_core/test_spatial_query.py
```diff
@@ -348,3 +352,33 @@ def test_bounding_box_spatial_data(full_sdata):
...
+    queried1 = sdata.query.bounding_box(
+        axes=("y", "x"),
+        min_coordinate=np.array([15, 15]),
+        max_coordinate=np.array([25, 25]),
+        filter_table=False,
+        target_coordinate_system="global",
+    )
+    assert len(queried0.table) == 1
+    assert len(queried1.table) == 3
+    ##
```
giovp, line 384:
```suggestion

```

giovp (approved):
really nice @LucaMarconato !

tests/_core/test_spatial_query.py
kevinyamauchi: I think we need to add a pre-commit check just for @LucaMarconato  😉

spatialdata/_core/_spatialdata.py
LucaMarconato: Ok thanks, putting a `kwargs` in `__call__` leaving a `TODO`

tests/_core/test_models.py
LucaMarconato: Well spotted! I have decorated some tests with `@pytest.mark.ci_only`, and now locally, when debugging, I run with `pytest -v -m "not ci_only"`

tests/_core/test_spatial_query.py
LucaMarconato: I though the same 🙈

Issue #144, opened 2023-02-18
Document the CLI in `__main__.py`


LucaMarconato:
Add documentation for the CLI usage of `spatialdata`.

Issue #61, opened 2022-12-08
Refactoring project structure


LucaMarconato:
Some refactoring of the code structure that are low priority:
- [x] switch to a `src/` project structure
- [x] split `read.py` and `write.py` into different files (maybe one per element), so that reader and writer for the same element are in the same file
- [x] rename `_spatialdata.py` into `spatialdata.py`

Issue #7, opened 2022-07-11
Function to transform all spatial data


kevinyamauchi:
@tischi proposed that we have a `SpatialData.transform()` method that returns a new object where all of the spatial coordinates have been transformed. We could potentially support this if `obsm` arrays can be annotated with a coordinate space.

See @tischi 's comment for more details: https://forum.image.sc/t/proposal-for-ome-ngff-table-specification/68908/20

LucaMarconato:
This will be available in https://github.com/scverse/spatialdata/pull/189

LucaMarconato:
actually, at least for the moment, I prefer not to have `.transform()` as a method of `SpatialData`. The user can still transform the whole `SpatialData` object with

```
from spatialdata import transform
transform(sdata, transformation)
```

the reason to keep this separate is that there is a parallel with the other operations on spatialdata objects, like

```
from spatialdata import rasterize, aggregate, bounding_box_query
```

and that in this way the syntax is consistent for elements, i.e. for elements, like images, you can't call 
```
my_image.transform()
```
but you have to do
```
transform(my_image, transformation)
```

PR #189, opened 2023-03-15
Module hierarchy refactoring

LucaMarconato:
closes https://github.com/scverse/spatialdata/issues/103
closes https://github.com/scverse/spatialdata/issues/61
closes https://github.com/scverse/spatialdata/issues/7
closes https://github.com/scverse/spatialdata/issues/144


LucaMarconato:
Closes also https://github.com/scverse/spatialdata/pull/189 and https://github.com/scverse/spatialdata/issues/61

LucaMarconato:
I am going to create a module called `transformations` instead of the originally planned `transform`. The reason is that we will have a transform function, which will transform an element/spatialdata object and I don't want to create confusion with names. `transformations` is a module for the transformation classes, `transform` is a function that applies a transformation.

LucaMarconato:
Minor change:
- [x] In `transform()` (the function to transform elements or sdata objects), I changed the default of `maintain_positioning` from `False` to `True`, after feedback from @EliHei2.

LucaMarconato:
Finished 🥵 Before merging I need to make sure that `spatialdata-io`, `spatialdata-sandbox`, `spatialdata-notebooks` and `napari-spatialdata` work.

giovp:
so have checked it out locally, will report as comments here (github is super slow at rendering it)
- remove `elements_utils`, that stuff can go in either transformations/operations or global utils.
- remove the `_constants` module, apparently it's never used
- I would remove `__main__.py` I doubt there is a single use case of that from users, but no strong feelings.

rest looks really good, thanks @LucaMarconato !

giovp (approved):

LucaMarconato:
@giovp thanks for the review. 

* I have split`elements_utils` into `spatialdata.models._utils` and `spatialdata._utils`. 
* I have removed the `_constants` module.
* I kept the CLI (`__main__.py`), but I have added a docstring (so closing https://github.com/scverse/spatialdata/issues/144). I use the CLI quite a lot for quick look at the data and I think this would be a great way to interact with the tool for people not familiar with Python. Also I plan to expand the CLI functions, for instance to quickly get the extent of the data, to show the DAG of coordinate transformations and to quickly extract a downscaled version of the image (like a png) from the large zarr images. I opened an issue about these new features: https://github.com/scverse/spatialdata/issues/190

PR #194, opened 2023-03-21
transforms fixes

LucaMarconato:
fixes to bugs found when running the xenium + visium notebook: 
- [x] corrected radius when transforming shapes
- [x] corrected transform_to_coordinate_system()

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -541,16 +541,23 @@ class SpatialData:
         The transformed element.
         """
         from spatialdata import transform
+        from spatialdata.transformations import Identity
         from spatialdata.transformations.operations import (
             get_transformation_between_coordinate_systems,
+            remove_transformation,
+            set_transformation,
         )
 
         t = get_transformation_between_coordinate_systems(self, element, target_coordinate_system)
         transformed = transform(element, t, maintain_positioning=False)
+        remove_transformation(transformed, remove_all=True)
+        set_transformation(transformed, Identity(), target_coordinate_system)
+
         return transformed
 
     def transform_to_coordinate_system(
-        self, target_coordinate_system: str, filter_by_coordinate_system: bool = True
+        self,
+        target_coordinate_system: str,
     ) -> SpatialData:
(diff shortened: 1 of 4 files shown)


src/spatialdata/_core/spatialdata.py
```diff
@@ -541,16 +541,23 @@ def transform_element_to_coordinate_system(
...
+            remove_transformation,
+            set_transformation,
         )
 
         t = get_transformation_between_coordinate_systems(self, element, target_coordinate_system)
         transformed = transform(element, t, maintain_positioning=False)
+        remove_transformation(transformed, remove_all=True)
+        set_transformation(transformed, Identity(), target_coordinate_system)
+
         return transformed
```
giovp, line 556: why does it set it and returns it? also, should not create transformed just for return but return directly `transform(element, t, maintain_positioning=False)`

giovp (approved):

src/spatialdata/_core/spatialdata.py
LucaMarconato: The transform function preserves all the mapping to coordinate systems. Here we explicitly remove them. Two comments. After the transformation refactoring I plan to move this function, and `transform_to_coordinate_system()` somewhere else, or maybe replace them with `transform()`. So I keep it like this, but next week we should revisit this.

src/spatialdata/_core/spatialdata.py
```diff
@@ -559,19 +566,12 @@ def transform_to_coordinate_system(
...
 
         Returns
         -------
         The transformed SpatialData.
         """
-        if filter_by_coordinate_system:
-            sdata = self.filter_by_coordinate_system(target_coordinate_system, filter_table=False)
-        else:
-            sdata = self
+        sdata = self.filter_by_coordinate_system(target_coordinate_system, filter_table=False)
```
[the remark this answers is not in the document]
LucaMarconato: No, the default is `True`

LucaMarconato:
Also restored `contributing.md` into `docs`. It got deleted somehow a while ago.

PR #196, opened 2023-03-27
aggregation

giovp:
this PR does the following
- [x] finishes up shapes aggregation from @ivirshup (originally in #162 
- [x] implement aggregation of images by labels
- [x] add an `aggregate` method to spatialdata to return spatialdata and not anndata (as the functional approach does instead)
- [x] add tests for aggregation of points/shapes by shapes
- [x] add tests for aggregation of images by labels
- [x] add test for the `SpatialData().aggregate` method


giovp:
I need to write tests for the `spatialdata.aggregate` method but a first round of review would be appreciated.

I also added a dataset so that's convenient to have dummy spatialdata for notebooks.
it looks kind of like this
image

I think we can put the raccoon also there.

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+    agg_func: str | list[str] = "count",
+) -> ad.AnnData:
+    # Have to get dims on dask dataframe, can't get from pandas
+    dims = get_axes_names(points)
+    # Default value for id_key
+    if id_key is None:
+        id_key = points.attrs[PointsModel.ATTRS_KEY][PointsModel.FEATURE_KEY]
+
+    if isinstance(points, ddf.DataFrame):
+        points = points.compute()
```
kevinyamauchi, line 111: I think we should add a note to the docstring to say this will bring lazy points into memory. That could be an unpleasant surprise for a user working with large, lazy data.

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+        Values to aggregate.
+    by
+        Regions to aggregate by.
+    id_key
+        Key to group observations in `values` by. E.g. this could be transcript id for points.
+        Defaults to `FEATURE_KEY` for points, required for shapes.
+    value_key
+        Key to aggregate values by. This is the key in the values object.
+        If nothing is passed here, assumed to be a column of ones.
+        For points, this could be probe intensity.
```
kevinyamauchi, line 56: I don't understand this example.

kevinyamauchi (approved):
Thanks @giovp ! This looks good to me. It looks like most of the changes are renaming, so I didn't review those closely. I would say we should merge optimistically so we can test the aggregations in our vignettes.

src/spatialdata/_core/operations/aggregate.py
giovp: basically it's the column in poijnts that you want to aggregate by, so if it's genes, it would return the gene counts matrix with `agg_func = "counts"` whereas for a float, `agg_func = "mean"` could work./

src/spatialdata/_core/spatialdata.py
```diff
@@ -210,6 +210,84 @@ def from_elements_dict(elements_dict: dict[str, Union[SpatialElement, AnnData]])
...
+            Aggregation function to apply over point values, e.g. "mean", "sum", "count".
+            Passed to :func:`pandas.DataFrame.groupby.agg` or from :func:`xrspatial.zonal_stats`
+            according to the type of `values`.
+        target_coordinate_system
+            Coordinate system to transform to before aggregating.
+        region_name
+            Name of the region `by`.
+        region_key
+            Key to use for the region column in the aggregated table.
+        instance_key
```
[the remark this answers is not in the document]
LucaMarconato: Also, as above, why can't this just be fetched by the table?

src/spatialdata/_core/spatialdata.py
```diff
@@ -210,6 +210,84 @@ def from_elements_dict(elements_dict: dict[str, Union[SpatialElement, AnnData]])
...
+            For points, this could be probe intensity.
+        agg_func
+            Aggregation function to apply over point values, e.g. "mean", "sum", "count".
+            Passed to :func:`pandas.DataFrame.groupby.agg` or from :func:`xrspatial.zonal_stats`
+            according to the type of `values`.
+        target_coordinate_system
+            Coordinate system to transform to before aggregating.
+        region_name
+            Name of the region `by`.
+        region_key
```
LucaMarconato, line 245: Why do we need to pass this? Can't we read it from the table?

src/spatialdata/_core/spatialdata.py
```diff
@@ -210,6 +210,84 @@ def from_elements_dict(elements_dict: dict[str, Union[SpatialElement, AnnData]])
     def query(self) -> QueryManager:
         return self._query
 
+    def aggregate(
+        self,
+        values: DaskDataFrame | GeoDataFrame | SpatialImage | MultiscaleSpatialImage,
+        by: str,
```
LucaMarconato, line 216: I would either make values be a string, or make by be a `GeoDataFrame | Image2DModel | Image3DModel | Labels2DModel | Labels3DModel`. I prefer the second. The string can be then derived by calling `sdata._locate_spatial_element()`, as it is done for instance in the function `transform_to_coordinate_system()`.

tests/core/operations/test_aggregations.py
```diff
@@ -0,0 +1,128 @@
...
+    with pytest.raises(ValueError):
+        aggregate(subcellular, cellular, agg_func="mean")
+
+
+def test_aggregate_circles_by_polygons() -> None:
+    # Basically the same as above, but not buffering explicitly
+    cellular = ShapesModel.parse(
+        gpd.GeoDataFrame(
+            geometry=[
+                shapely.Polygon([(0.5, 7.0), (4.0, 2.0), (5.0, 8.0)]),
```
LucaMarconato, line 81: We should test also multipolygons, I have created a issue here so we can do this later: https://github.com/scverse/spatialdata/issues/213

LucaMarconato (commented):
Great work! I added a few comments on code changes that I would make before merging. I also opened some issue to improve later.

LucaMarconato:
The main comment in the review, that I would address before merging, is that I find the arguments `id_key` and `value_key` (and their relationship with `value`, `by`, `region_key` and `instance_key` in the `sdata.aggregate` call) confusing. I added comments on the code on how specifically I consider them confusing and how to maybe address that.

I have then added extra comments on things that we can improve later. I have created dedicated GitHub issues so that we can proceed with the merge. Here is an overview:
- [ ] overview of all the remaining cases not covered by the current APIs: https://github.com/scverse/spatialdata/issues/216. In particular: points by labels https://github.com/scverse/spatialdata/issues/209.
- [ ] multiscale case: https://github.com/scverse/spatialdata/issues/215
- [ ] multipolygons case: https://github.com/scverse/spatialdata/issues/213
- [ ] special case: https://github.com/scverse/spatialdata/issues/211
- [ ] points and lazy loading: https://github.com/scverse/spatialdata/issues/210

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+    for i, c in enumerate(values.coords["c"].values):
+        out = zonal_stats(by, values[i, ...], stats_funcs=agg_func, **kwargs).compute()
+        out.columns = [f"channel_{c}_{col}" if col != "zone" else col for col in out.columns]
+        out = out.loc[out["zone"] != 0].copy()
+        zones: ArrayLike = out["zone"].values
+        outs.append(out.drop(columns=["zone"]))  # remove the 0 (background)
+    df = pd.concat(outs, axis=1)
+
+    X = sparse.csr_matrix(df.values)
+
```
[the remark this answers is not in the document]
giovp: but wait, now by default in parser we don't accept name anymore, hence it's always image

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+    index = kwargs.get("zone_ids", None)  # `zone_ids` allows the user to select specific labels to aggregate by
+    if index is None:
+        index = np.array(da.array.unique(by.data))
```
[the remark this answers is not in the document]
giovp: above what?

src/spatialdata/_core/operations/aggregate.py
giovp: done

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -0,0 +1,261 @@
...
+        value_key = "count"
+    else:
+        point_values = joined[value_key]
+
+    to_agg = pd.DataFrame(
+        {
+            by_id_key: joined.index,
+            id_key: joined[id_key].values,
+            value_key: point_values,
+        }
```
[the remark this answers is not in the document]
giovp: I think this could be a good default for the index.name?>

src/spatialdata/_core/spatialdata.py
giovp: not really no, cause it represent the region of the new aggregated table. So it must be a new one.

src/spatialdata/_core/operations/aggregate.py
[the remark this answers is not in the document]
LucaMarconato: I am not 100% sure, but I think that we said that in the code we make the assumption that there is only one variable, but not about the name, and we use something like `next(iter(my_multiscale['scale0']))` to access this variable. At least this is what I am doing, I am never assuming to know the name.

src/spatialdata/_core/operations/aggregate.py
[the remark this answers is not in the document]
LucaMarconato: I think the comment got misplaced, it was regarding the use of `'image`. The first comment was for `by = by[key]["image"]`, the second was for `values = values[key]["image"`

src/spatialdata/_core/operations/aggregate.py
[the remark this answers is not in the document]
LucaMarconato: I would `instance` or `region`. If we aggregate by ROIs or Visium circles it could get confusing.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Ok I think I got it.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Got it! I'll make the docstring more explicit about this.

src/spatialdata/_core/spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: Now I got it, I'll try to make the docstring more explicit.

src/spatialdata/_core/operations/aggregate.py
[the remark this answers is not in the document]
LucaMarconato: I made this change in the code in the latest commit.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Done.

src/spatialdata/_core/spatialdata.py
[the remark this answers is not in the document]
LucaMarconato: Done.

LucaMarconato:
> @giovp yesterday I realized that in the Xenium + Visium code we have the following use case.
> 
> We want to aggregate points, by the Visium clonality, i.e. each Visium circle has a categorical column that tells which clone the circle belongs too. There are 5 classes in this column, so the output would be a 5 x n table, where n is the number of genes the points can be.
> 
> I am not sure it is covered already by the PR. Is it?

Checking the code better I think that this is not supported but this behavior could be enabled around `id_key`. Currently `id_key` does the same but for columns present in `values`, the idea would to make it look for columns both in `values` and `by`. What do you think? We can also open an issue, merge and do this later.

giovp:
> > @giovp yesterday I realized that in the Xenium + Visium code we have the following use case.
> > We want to aggregate points, by the Visium clonality, i.e. each Visium circle has a categorical column that tells which clone the circle belongs too. There are 5 classes in this column, so the output would be a 5 x n table, where n is the number of genes the points can be.
> > I am not sure it is covered already by the PR. Is it?
> 
> Checking the code better I think that this is not supported but this behavior could be enabled around `id_key`. Currently `id_key` does the same but for columns present in `values`, the idea would to make it look for columns both in `values` and `by`. What do you think? We can also open an issue, merge and do this later.

uu ok I think I get it now, basically aggregation per category, think this would be a tad more complicated than that, let's do it in another PR later on.

giovp:
> One test related to `AnnData` fails, I see that this [is being also discussed in Zulip](https://scverse.zulipchat.com/#narrow/stream/316218-repo-management/topic/pandas.202.2E0.20.E2.86.92.20relase.20anndata.200.2E9.3F/near/348444096).

new release fixed it.

PR #200, opened 2023-03-31
revert omero

giovp:
unfortunately, it looks like omero rewrites the name metadata if present: https://github.com/ome/ome-zarr-py/pull/261

this would require significant rewrite of IO from our side which I wouldn't do now, so using `channel_metadata` for noe.


giovp:
it's super minor thing and doesn't really fix omero so I'll merge anyway

Issue #201, opened 2023-04-02
Bug with omero metadata


LucaMarconato:
@giovp in this code

```python
# if image, read channels metadata
if raster_type == "image" and channels_metadata is not None:
    channels: list[Any] = fmt.channels_from_metadata(channels_metadata)
axes = [i["name"] for i in node.metadata["axes"]]
if len(datasets) > 1:
    multiscale_image = {}
    for i, d in enumerate(datasets):
        data = node.load(Multiscales).array(resolution=d, version=fmt.version)
        multiscale_image[f"scale{i}"] = DataArray(
            data,
            name=name,
            dims=axes,
            coords={"c": channels} if raster_type == "image" else {},
        )
    msi = MultiscaleSpatialImage.from_dict(multiscale_image)
    _set_transformations(msi, transformations)
    return compute_coordinates(msi)
data = node.load(Multiscales).array(resolution=datasets[0], version=fmt.version)
si = SpatialImage(
    data,
    name=name,
    dims=axes,
    coords={"c": channels} if raster_type == "image" else {},
)
```

The first `if` sometimes is not executed. In those cases the variable `channels` is not defined and I get this error:

```
/Users/macbook/miniconda3/envs/ome/bin/python /Users/macbook/embl/projects/basel/spatialdata-sandbox/examples/xenium_and_visium.py 
/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/geopandas/_compat.py:123: UserWarning: The Shapely GEOS version (3.11.1-CAPI-1.17.1) is incompatible with the GEOS version PyGEOS was compiled with (3.10.4-CAPI-1.16.2). Conversions between both will be slow.
  warnings.warn(
/Users/macbook/embl/projects/basel/spatialdata/src/spatialdata/__init__.py:9: UserWarning: Geopandas was set to use PyGEOS, changing to shapely 2.0 with:

	geopandas.options.use_pygeos = True

If you intended to use PyGEOS, set the option to False.
  _check_geopandas_using_shapely()
Traceback (most recent call last):
  File "/Users/macbook/embl/projects/basel/spatialdata-sandbox/examples/xenium_and_visium.py", line 21, in <module>
    xenium_sdata = sd.read_zarr(XENIUM_SDATA_PATH)
  File "/Users/macbook/embl/projects/basel/spatialdata/src/spatialdata/_io/io_zarr.py", line 36, in read_zarr
    images[k] = _read_multiscale(f_elem_store, raster_type="image")
[...]

PR #202, opened 2023-04-02
fix for bug with missing channels

LucaMarconato:
quick fix

closes https://github.com/scverse/spatialdata/issues/201

### DIFF
--- a/src/spatialdata/_io/io_raster.py
+++ b/src/spatialdata/_io/io_raster.py
@@ -69,8 +69,9 @@ def _read_multiscale(
     transformations = _get_transformations_from_ngff_dict(encoded_ngff_transformations)
     name = os.path.basename(node.metadata["name"])
     # if image, read channels metadata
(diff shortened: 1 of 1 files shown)


src/spatialdata/_io/io_raster.py
```diff
@@ -80,7 +81,7 @@ def _read_multiscale(
                 data,
                 name=name,
                 dims=axes,
-                coords={"c": channels} if raster_type == "image" else {},
+                coords={"c": channels} if channels is not None else {},
```
giovp, line 84: thanks! when is this prolbem arising?

giovp (approved):

src/spatialdata/_io/io_raster.py
LucaMarconato: I think I triggered this by trying to open files saved with the old omero metadata.

PR #204, opened 2023-04-03
tiny bugfix

LucaMarconato:
- fix rounding bug in rasterize
- add torch transform argument


LucaMarconato:
I merge because I have a problem with the data ci is running now and keeps checking out the main branch otherwise.

PR #206, opened 2023-04-03
Docs/improvements

LucaMarconato:
In this PR (ready to merge):
- I fixed all the warnings that the docs generation was making;
- I improved the aesthetics of the docs (asked by Oli).

More on the aesthetics:
- I added an icon on the top left of the docs
- I changed the theme from `sphinx_book_theme` to `sphinx_rtd_theme`. I think that the latter is graphically more pleasant (see also screenshots).

Not in this PR:
- I checked how much work it would be to make a gallery of icons for the vignette (like in squidpy). It's like a few hours of work, but I think we should do this with low priority (after all the notebooks are up). I created this issue to keep track of the progress https://github.com/scverse/spatialdata/issues/205.

PR #207, opened 2023-04-04
Tiles/fix

LucaMarconato:
https://github.com/scverse/spatialdata/pull/206 should be merged first so the diff in this PR becomes minimarl.


LucaMarconato:
Ready for review.

LucaMarconato:
@giovp @kevinyamauchi  this PR fixes some bugs in the tiling code, please review and merge when you have time.

src/spatialdata/dataloader/datasets.py
```diff
@@ -114,9 +117,18 @@ def __getitem__(self, idx: int) -> tuple[SpatialImage, str, int]:
...
-            centroid = np.atleast_2d(region.geometry.coords._coords[0])
+            shape = regions.geometry.iloc[0]
+            if isinstance(shape, Polygon):
+                centroid = np.atleast_2d(region.geometry.centroid.coords._coords[0])
+            elif isinstance(shape, MultiPolygon):
+                raise NotImplementedError("MultiPolygon not supported yet")
+            elif isinstance(shape, Point):
+                # the function coords.xy is just accessing _coords, and wrapping it with extra information,
+                # so we access it directly
+                centroid = np.atleast_2d(region.geometry.coords._coords[0])
```
kevinyamauchi, line 128: Do we have a test to cover this to catch if they change the private API?

kevinyamauchi (approved):
Hey @LucaMarconato ! This looks good to me. I made a suggestion for a test to be added. Once you've done that, please feel free to merge.

src/spatialdata/dataloader/datasets.py
LucaMarconato: Good point! Seems to be already broken. I am replacing this with public APIs and adding some tests (so far the code is not tested).

LucaMarconato: Turned out the private APIs were still working (and did something wrong with the tests). Anyway, I removed them and now we are using public APIs.

PR #236, opened 2023-05-02
add deploy to pypi

kevinyamauchi:
This PR adds a github actions just to deploy to pypi when a tag starting with `v` is pushed to main.


kevinyamauchi:
okay, I think is working @giovp . the read the docs is failing because the previous tag is not a valid pattern, so it can't semantically bump automatically.

giovp (approved):
think we don't need hjatch, last week at hackathon they piut together a bunch of github actions for this, linked them in comment

PR #271, opened 2023-05-19
Fix/aggregation

LucaMarconato:


tests/dataloader/test_datasets.py
```diff
@@ -0,0 +1,62 @@
+import contextlib
```
LucaMarconato, line 1: This file is entirely present in main, I just moved it from `tests/test_dataloader/test_datasets.py` to `tests/dataloader/test_datasets.py`

LucaMarconato:
Ready for review @giovp @kevinyamauchi

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -265,7 +267,8 @@ def _aggregate_shapes(
     ).tocsr()
     return ad.AnnData(
         X,
-        obs=pd.DataFrame(index=obs_id_categorical.categories),
+        # obs=pd.DataFrame(index=obs_id_categorical.categories),
+        obs=pd.DataFrame(index=pd.Categorical(by.index.tolist()).categories),
```
giovp, line 271: why changing this line, is modified in 259 already, would use that, logic is not clear

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -649,12 +655,12 @@ def _polygon_query(
     new_points = {}
     if points:
         for points_name, p in sdata.points.items():
-            points_gdf = points_dask_dataframe_to_geopandas(p)
+            points_gdf = points_dask_dataframe_to_geopandas(p, suppress_z_warning=True)
```
giovp, line 658: this are we sure we want to suppress warning and not just pass the x,y instead?

giovp (approved):

PR #276, opened 2023-05-22
Refactoring aggregation

LucaMarconato:
Removing `id_key` from aggregation to simplify the APIs.

### DIFF
--- a/src/spatialdata/_core/operations/aggregate.py
+++ b/src/spatialdata/_core/operations/aggregate.py
@@ -41,7 +41,7 @@ def aggregate(
     values: Optional[ddf.DataFrame | gpd.GeoDataFrame | SpatialImage | MultiscaleSpatialImage | str] = None,
     by: Optional[gpd.GeoDataFrame | SpatialImage | MultiscaleSpatialImage] = None,
     value_key: list[str] | str | None = None,
-    agg_func: str | list[str] = "mean",
+    agg_func: str | list[str] = "sum",
     target_coordinate_system: str = "global",
     fractions: bool = False,
     **kwargs: Any,
@@ -86,8 +86,9 @@ def aggregate(
              - when aggregating points this values must be left to False, as the points don't have area, thus
              otherwise a table of zeros will be obtained;
              - for categorical values count and sum are equivalent when fractions = False, but when fractions = True
-             count and sum are different: count would give unmeaningful results, while sum actually sums the values
-             of the intersecting regions, and should thus be used.
+             count and sum are different: count would give unmeaningful results and so it's not allowed, while sum
+             actually sums the values of the intersecting regions, and should thus be used.
+             - aggregating categorical values with agg_func = 'mean' is not allowed as it give unmeaningful results
 
     kwargs
         Additional keyword arguments to pass to :func:`xrspatial.zonal_stats`.
@@ -283,15 +284,27 @@ def _aggregate_shapes(
 
     categorical = pd.api.types.is_categorical_dtype(actual_values.iloc[:, 0])
 
+    # deal with edge cases
+    if fractions:
+        assert not (categorical and agg_func == "count"), (
+            "Aggregating categorical values using fractions=True and agg_func='count' will most likely give "
+            "unmeaningful results. Please consider using a different aggregation function, for instance "
+            "agg_func='sum' instead."
+        )
+        assert not isinstance(values.iloc[0].geometry, Point), (
+            "Fractions cannot be computed when values are points. " "Please use fractions=False."
+        )
+    assert not (categorical and agg_func == "mean"), (
+        "Incompatible choice: aggregating a categorical column with " "agg_func='mean'"
+    )
+
     ONES_COLUMN = "__ones_column"
     AREAS_COLUMN = "__areas_column"
-    JOINED_AREAS_COLUMN = "__joined_areas_column"
 
     e = f"Column names {ONES_COLUMN} and {AREAS_COLUMN} are reserved for internal use. Please rename your columns."
     assert ONES_COLUMN not in by.columns, e
     assert AREAS_COLUMN not in by.columns, e
     assert ONES_COLUMN not in actual_values.columns, e
-    assert AREAS_COLUMN not in actual_values.columns, e
 
     by[ONES_COLUMN] = 1
     by[AREAS_COLUMN] = by.geometry.area
@@ -325,7 +338,6 @@ def _aggregate_shapes(
 
         assert "__index" not in joined
         joined["__index"] = joined.index
-        joined[JOINED_AREAS_COLUMN] = joined.geometry.area
     else:
         overlayed = gpd.overlay(by, values, how="intersection")
 
@@ -342,24 +354,11 @@ def _aggregate_shapes(
             inplace=True,
         )
         overlayed["__index"] = overlayed["__index_left"]
-        overlayed[JOINED_AREAS_COLUMN] = overlayed.geometry.area
-
         joined = overlayed
 
+    fractions_of_values = None
     if fractions:
-        if categorical and agg_func == "count":
-            logger.warning(
-                "Aggregating categorical values using fractions=True and agg_func='count' will most likely give "
-                "unmeaningful results. Please consider using a differnt aggregation function, for instance "
-                "agg_func='sum' instead."
-            )
-        if isinstance(values.iloc[0].geometry, Point):
-            raise ValueError("Fractions cannot be computed when values are points. Please use fractions=False.")
-        # TODO::::::::::::::::::::::::::::::::::::::::::::::::::: most recent todoe
-        # TODO: this code works for numerical, but it's not good for categorical; maybe use a lambda finction  instead
-        # TODO: dissallow mean for categorical and explain this in the docstring
-        fractions_of_values = joined[JOINED_AREAS_COLUMN] / joined[AREAS_COLUMN + "_right"]
-        joined[value_key] = joined[value_key].to_numpy() * fractions_of_values.to_numpy().reshape(-1, 1)
+        fractions_of_values = joined.geometry.area / joined[AREAS_COLUMN + "_right"]
     ##
     # with pd.option_context(
     #     "display.max_rows",
@@ -379,20 +378,15 @@ def _aggregate_shapes(
         # give a different table as result of the aggregation, and we only support single tables
         assert len(value_key) == 1
         vk = value_key[0]
+        if fractions_of_values is not None:
+            joined[ONES_COLUMN + "_right"] = fractions_of_values
         aggregated = joined.groupby(["__index", vk])[ONES_COLUMN + "_right"].agg(agg_func).reset_index()
         aggregated_values = aggregated[ONES_COLUMN + "_right"].values
-        # joined.groupby([joined.index, vk])[[ONES_COLUMN + '_right', AREAS_COLUMN + '_right']].agg("sum")
     else:
-        ##
-        agg_func
-        joined
-        by
-        values
-        joined.iloc[0]
-        ##
+        if fractions_of_values is not None:
+            joined[value_key] = joined[value_key].to_numpy() * fractions_of_values.to_numpy().reshape(-1, 1)
         aggregated = joined.groupby(["__index"])[value_key].agg(agg_func).reset_index()
         aggregated_values = aggregated[value_key].values
-        # joined.groupby([joined.index, vk])[[ONES_COLUMN + '_right', AREAS_COLUMN + '_right']].agg("sum")
 
     # Here we prepare some variables to construct a sparse matrix in the coo format (edges + nodes)
     rows_categories = by.index.tolist()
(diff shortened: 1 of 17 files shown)


src/spatialdata/_core/operations/aggregate.py
```diff
@@ -221,53 +254,186 @@ def _aggregate_shapes(
...
+        )
+        assert not isinstance(values.iloc[0].geometry, Point), (
+            "Fractions cannot be computed when values are points. " "Please use fractions=False."
+        )
+    assert not (categorical and agg_func == "mean"), (
+        "Incompatible choice: aggregating a categorical column with " "agg_func='mean'"
     )
+
+    ONES_COLUMN = "__ones_column"
+    AREAS_COLUMN = "__areas_column"
```
kevinyamauchi, line 302: can you please add a comment explaining what these are for

kevinyamauchi (approved):
Hey @LucaMarconato ! Thanks for walking me through your PR today. As I mentioned, I don't have time to review in depth.  I given the tests and that you will be testing these functions in the Xenium-Visium vignette, I think we should just merge after you've take a pass to clean up a bit (remove commented code blocks, `##`, etc.)

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+if TYPE_CHECKING:
+    from spatialdata import SpatialData
+
 __all__ = ["aggregate"]
 
 
+def _parse_element(element: str | SpatialElement, sdata: SpatialData | None, str_for_exception: str) -> SpatialElement:
+    if not ((sdata is not None and isinstance(element, str)) ^ (not isinstance(element, str))):
+        raise ValueError(
+            f"To specify the {str_for_exception!r} spatial element element, please do one of the following: "
```
giovp, line 44:
```suggestion
            f"To specify the {str_for_exception!r} spatial element, please do one of the following: "
```

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+            f"- either pass a SpatialElement to the {str_for_exception!r} parameter (and keep "
```
giovp, line 45: spatialElement her is capitalized and up not

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+
+
 def aggregate(
-    values: ddf.DataFrame | gpd.GeoDataFrame | SpatialImage | MultiscaleSpatialImage,
-    by: gpd.GeoDataFrame | SpatialImage | MultiscaleSpatialImage,
-    id_key: str | None = None,
-    *,
-    value_key: str | None = None,
-    agg_func: str | list[str] = "mean",
+    values_sdata: SpatialData | None = None,
```
giovp, line 58: I don't understand how a function can have all default arguments, at least one should have a mandatory argument no?

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
     ----------
+    values_sdata
+        SpatialData object containing the values to aggregate: if None, values must be a SpatialElement; if not None,
+        values must be a string.
     values
-        Values to aggregate.
+        The values to aggregate: if values_sdata is None, must be a SpatialElement, otherwise must be a string
+        specifying the name of the SpatialElement in values_sdata
+    by_sdata
+        Regions to aggregate by: if None, by must be a SpatialElement; if not None, by must be a string.
```
giovp, line 83: please use `by` when referring to the argument otherwise it's the same as literal by

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+        If nothing is passed here, it defaults to the equivalent of a column of ones.
+        Defaults to `FEATURE_KEY` for points (if present).
     agg_func
         Aggregation function to apply over point values, e.g. "mean", "sum", "count".
         Passed to :func:`pandas.DataFrame.groupby.agg` or to :func:`xrspatial.zonal_stats`
         according to the type of `values`.
     target_coordinate_system
         Coordinate system to transform to before aggregating.
+    fractions
+        Adjusts for partial areas overlap between regions in values and by.
```
giovp, line 106: also here, `values` and `by`

giovp: and everywhere else really

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+        More precisely: in the case in which a region in by partially overlaps with a region in values, this setting
+        specifies whether the value to aggregate should be considered as it is (fractions = False) or it is to be
+        multiplied by the following ratio: "area of the intersection between the two regions" / "area of the region in
+        values".
+
+        Additional details:
+
+             - default is fractions = False.
+             - when aggregating points this values must be left to False, as the points don't have area, thus
+             otherwise a table of zeros will be obtained;
```
giovp, line 116: remove thus

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+             - for categorical values count and sum are equivalent when fractions = False, but when fractions = True
```
giovp, line 117:
```suggestion
             - for categorical values, `count` and `sum` are equivalent when fractions = False, but when fractions = True
```

giovp: and also fix in rest

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+             count and sum are different: count would give unmeaningful results and so it's not allowed, while sum
```
giovp, line 118:
```suggestion
             count and sum are different: count would give not meaningful results and so it's not allowed, while sum
```

giovp: also in the other please fix

src/spatialdata/_core/operations/aggregate.py
```diff
@@ -1,148 +1,256 @@
...
+    deepcopy
+        Whether to deepcopy the shapes in the returned SpatialData object. If the shapes are large (e.g. large
+        multiscale labels), you may consider disabling the deepcopy to use a lazy Dask representation.
     kwargs
         Additional keyword arguments to pass to :func:`xrspatial.zonal_stats`.
 
     Returns
     -------
-    AnnData of shape (by.shape[0], values[id_key].nunique())])
+    Returns a SpatialData object with the "by" shapes as SpatialElement and a table with the aggregated values
```
giovp, line 134: please use `by`

src/spatialdata/_core/query/relational_query.py
```diff
@@ -48,8 +99,186 @@ def _filter_table_by_elements(table: AnnData, elements_dict: dict[str, dict[str,
...
+        SpatialElement object; either element or (sdata, element_name) must be provided
+    sdata
+        SpatialData object; either element or (sdata, element_name) must be provided
+    element_name
+        Name of the element; either element or (sdata, element_name) must be provided
+
+    Returns
+    -------
+    DataFrame or ndarray with the values requested.
+
```
giovp, line 230: remove empty line,

src/spatialdata/_core/query/relational_query.py
```diff
@@ -48,8 +99,186 @@ def _filter_table_by_elements(table: AnnData, elements_dict: dict[str, dict[str,
...
+                if value_key in table.obs.columns:
+                    value = table.obs[value_key]
+                    is_categorical = pd.api.types.is_categorical_dtype(value)
+                    origins.append(_ValueOrigin(origin="obs", is_categorical=is_categorical, value_key=value_key))
+                # check if the value_key is in the var
+                elif value_key in table.var_names:
+                    origins.append(_ValueOrigin(origin="var", is_categorical=False, value_key=value_key))
+
+    # adding from the channel names of images
+    # TODO
```
giovp, line 203: remove

src/spatialdata/_core/spatialdata.py
```diff
@@ -213,91 +213,51 @@ def query(self) -> QueryManager:
...
-        region_key
-            Name that will be given to the new region column in the returned aggregated table.
-        instance_key
-            Name that will be given to the new instance id column in the returned aggregated table.
-        kwargs
-            Additional keyword arguments to pass to :func:`xrspatial.zonal_stats`.
+        Notes
+        -----
+        This function calls :func:`spatialdata.aggregate` with the convenience that values and by can be string
+        without having to specify the values_sdata and by_sdata, which in that case will be replaced by `self`.
```
giovp, line 235: by reading this docstring, why is values_sdata and by_sdata still there?

tests/conftest.py
```diff
@@ -392,29 +409,47 @@ def _make_sdata_for_testing_querying_and_aggretation() -> SpatialData:
...
+        _ = spatialdata_plot
+
+        ax = plt.gca()
+        sdata.pl.render_shapes(element="values_polygons", color="categorical_in_obs").pl.show(ax=ax)
+        sdata.pl.render_shapes(element="values_circles", color="categorical_in_obs").pl.show(ax=ax)
+        sdata.pl.render_shapes(element="by_polygons", na_color=(1.0, 0.7, 0.7, 0.3)).pl.show(ax=ax)
+        sdata.pl.render_shapes(element="by_circles", na_color=(1.0, 0.7, 0.7, 0.3)).pl.show(ax=ax)
+        sdata.pl.render_points(color="categorical_in_ddf", size=10.0, palette="tab10").pl.show(ax=ax)
+        plt.show()
     return sdata
```
giovp, line 450: I don't think we want to have data plotting to be a dependency of this, I would remove this

giovp (commented):
mostly comment on docstrings. I must sa y I still don't like the values and sdata_values logic, but I guess it's too late to discuss again API

giovp (approved):
mostly comment on docstrings. I must sa y I still don't like the values and sdata_values logic, but I guess it's too late to discuss again API

giovp:
@LucaMarconato #303 introduced the pyupgrade typing so that stuff like `Optional[str]` should be changed to `str | None`.

I ignore such change selectively in some files (e.g. transformation module) and the `UP` code (pyupgrade rules by ruff) is added as `unfixable` (ruff can fix everything but I would keep this as unfixable as it can be useful to reason about it, especially be contributors). 

It will probably throw a bunch of errors in the files you are working on, so an option is to remove `UP` from unficable, run ruff with `pre-commit run --all-files` and then re-add `UP` to unfixable. So it fixes the files you want to get fixed automatically, but it won't in the future. 

I apologize for the conflicts but I took a look and they are very minor.

on a separate note, there are a lot of `per-file-ignore` in the transformation module. Would keep an eye on that when the transformation is refactored

src/spatialdata/_core/operations/aggregate.py
LucaMarconato: true, `values` and `by` are always mandatory, I will change the order. I made them optional because I had put them after the optional values `values_sdata` and `by_sdata`

src/spatialdata/_core/operations/aggregate.py
LucaMarconato: Done, also for `values`

src/spatialdata/_core/operations/aggregate.py
LucaMarconato: done

src/spatialdata/_core/operations/aggregate.py
LucaMarconato: done

src/spatialdata/_core/query/relational_query.py
LucaMarconato: done

src/spatialdata/_core/spatialdata.py
LucaMarconato: @giovp because one may still want to aggregate values from a different object.

tests/conftest.py
LucaMarconato: agree, we don't to have it as a dependency. It was useful for developing the tests, but I will move it to the sandbox.

src/spatialdata/_core/spatialdata.py
giovp: yeah but this is the method right? by this logic there is no differentiation between what the function does and what the method does?

LucaMarconato: Exactly, there is no differentiation and the method is a convenience. I am in favor on this as we do it also for `bounding_box_query`, `polygon_query` and `transform` (well in this case it's called `transform_to_coordinate_system`, but after the refactoring will also be renamed). I will push soon the notebook, there I use mostly the method because since all the elements are in the `sdata` objects there is no need to pass `values_sdata` and `by_sdata`, and I can just pass `values` and `by` as strings.

LucaMarconato:
Thanks for the update! I will keep this in mind and gradually fix that!

LucaMarconato:
> mostly comment on docstrings. I must sa y I still don't like the values and sdata_values logic, but I guess it's too late to discuss again API

Thanks for the review, I now implemented the changes (only one open comment left). I actually think that the `value_sdata` and `by_sdata` usage is quite flexible and intuitive. I have updated the aggregation notebook so we can collect some feedback. Anyway, up to change/improve it later on!

PR #278, opened 2023-05-23
Support for consolidated remote zarr

### LINKED ISSUE #275: Reading from remote .zarr stores
Right now, reading from a remote .zarr URL fails:
```
In [7]: sdata = sd.SpatialData.read('https://s3.embl.de/spatialdata/spatialdata-sandbox/cosmx_io.zarr/')
---------------------------------------------------------------------------
PathNotFoundError                         Traceback (most recent call last)
Input In [7], in <cell line: 1>()
----> 1 sdata = sd.SpatialData.read('https://s3.embl.de/spatialdata/spatialdata-sandbox/cosmx_io.zarr/')

File ~/Documents/GitHub/spatialdata/src/spatialdata/_core/spatialdata.py:1183, in SpatialData.read(file_path)
   1179 @staticmethod
   1180 def read(file_path: str) -> SpatialData:
   1181     from spatialdata import read_zarr
-> 1183     return read_zarr(file_path)

File ~/Documents/GitHub/spatialdata/src/spatialdata/_io/io_zarr.py:22, in read_zarr(store)
     19 if isinstance(store, str):
     20     store = Path(store)
---> 22 f = zarr.open(store, mode="r")
     23 images = {}
     24 labels = {}

File /opt/homebrew/Caskroom/mambaforge/base/envs/spatialdata/lib/python3.10/site-packages/zarr/convenience.py:122, in open(store, mode, zarr_version, path, **kwargs)
    120     return open_group(_store, mode=mode, **kwargs)
    121 else:
--> 122     raise PathNotFoundError(path)

PathNotFoundError: nothing found at path ''
```

Ideally, this would just work the same as a local dataset or fail with informative errors. Some issues causing this:

- [ ] URL strings are parsed using os.path or Path, which does not work for: 
     - [ ] .exist() e.g. `images_store.exists() in io_zarr.py`
     - [ ] the remote store is not iterable for subelements as a local path is e.g. `for k in f in io_raster.py 
- [ ] URL strings are sometimes required to be str | Path:
     - [ ] e.g.` _read_multiscale in io_raster.py`
    

The library [urlpath](https://github.com/brandonschabell/urlpath) extends pathlib to also work with URLs, but ideally this is handled by the string parsing of the zarr library instead.

The list of subelements can be provided by consolidate_metadata, same as https://github.com/kevinyamauchi/ome-ngff-tables-prototype/pull/12

berombau:
Closes #275

- Uses `._store.path` and Path to get to subelements.
- Ignores hidden files

### DIFF
--- a/src/spatialdata/_io/io_zarr.py
+++ b/src/spatialdata/_io/io_zarr.py
@@ -1,5 +1,6 @@
 import logging
 from pathlib import Path
+from urlpath import URL
 from typing import Optional, Union
 
 import numpy as np
@@ -14,12 +15,15 @@ from spatialdata._io.io_raster import _read_multiscale
 from spatialdata._io.io_shapes import _read_shapes
 from spatialdata.models import TableModel
 
+from spatialdata._logging import logger
+
 
 def read_zarr(store: Union[str, Path, zarr.Group]) -> SpatialData:
-    if isinstance(store, str):
-        store = Path(store)
+    if isinstance(store, zarr.Group):
+        f = store
+    else:
+        f = zarr.open(store, mode="r")
 
-    f = zarr.open(store, mode="r")
     images = {}
     labels = {}
     points = {}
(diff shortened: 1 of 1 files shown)


berombau:
I will test it with napari-spatialdata for the consolidated version at https://dl01.irc.ugent.be/spatial/cosmx/data.zarr/, but someone with an S3 bucket should test it with `s3://`, maybe there is still a bug there (@LucaMarconato).

LucaMarconato:
Thanks @berombau this is amazing! 🚀

src/spatialdata/_io/io_zarr.py
```diff
@@ -12,62 +13,73 @@
...
-    if images_store.exists():
-        f = zarr.open(images_store, mode="r")
-        for k in f:
-            f_elem = f[k].name
-            f_elem_store = f"{images_store}{f_elem}"
-            images[k] = _read_multiscale(f_elem_store, raster_type="image")
+    if "images" in f:
+        group = f["images"]
+        count = 0
+        for k in group:
```
kevinyamauchi, line 33: Minor nitpick: I would prefer if the iterator variable had a more clear name (e.g., `subgroup_name` instead of `k`).

kevinyamauchi (commented):
Hey @berombau ! This is super nice. I left a couple of very minor comments below. I think you can wait until @LucaMarconato does a full review to address them. I just wanted to write them down while I was having a look.

Thanks!

berombau:
fyi `python -m spatialdata peek https://dl01.irc.ugent.be/spatial/cosmx/data.zarr/` takes a minute, downloading and serving locally is faster with `python -m http.server -b 0.0.0.0` and `python -m spatialdata peek http://0.0.0.0:8000/cosmx/data.zarr`

giovp (approved):
thanks a lot for this @berombau ! Approve in principle but this doesn't work for me,
` python -m spatialdata peek https://dl01.irc.ugent.be/spatial/cosmx/data.zarr`

`Error: .zarr storage not found at https://dl01.irc.ugent.be/spatial/cosmx/data.zarr. Please specify a valid OME-NGFF spatial data (.zarr) file. Example "python -m spatialdata peek data.zarr"`

although I can click on the link and take a look at the HTML. Do you happen of any idea why? I will try from another network just in case tomorrow

berombau:
I committed a change to the CLI a forgot to push. This should work now and is also only 10 seconds instead of minutes: `python -m spatialdata peek https://dl01.irc.ugent.be/spatial/mibitof/data.zarr`.

LucaMarconato:
@berombau thanks for the PR (and sorry for the long silence on this). I tried it with your two urls you gave and the functionality is super cool!
Some comments below.
- I added docstrings for the `selection` parameters in the CLI and in the `SpatialData.read` and `read_zarr` functions.
- I tried to use this on the mibitof dataset, after I reuploaded it in S3 with the consolidated metadata but it didn’t work (it showed an empty dataset without any error). The reason was a bug in `zarr` https://github.com/zarr-developers/zarr-python/issues/1121. In short `zmetadata` is written to disk instead of `.zmetadata` when the file separator is `/`, so when reading it fails. I believe that you didn't encounter the problem because the files present in the url that you used are "discoverable", while in a S3 storage one "can't do `ls`".
- I noticed that in the mibitof dataset what takes most of the time is loading the (small) table. There will have room to improve this in the future.
-

LucaMarconato:
Reported the bug in the `anndata` repo: https://github.com/scverse/anndata/issues/1056.

LucaMarconato (approved):

src/spatialdata/__main__.py
```diff
@@ -3,39 +3,46 @@
...
-
 import click
 
 
-@click.command(help="Peek inside the SpatialData .zarr file")
+@click.command(help="Peek inside the SpatialData .zarr dataset")
 @click.argument("path", default=False, type=str)
-def peek(path: str) -> None:
+@click.argument("selection", type=str, nargs=-1)
+def peek(path: str, selection: tuple[str]) -> None:
```
giovp, line 14: `selection` could be `Literal`?

src/spatialdata/__main__.py
```diff
@@ -3,39 +3,46 @@
...
+        # so we just catch the ValueError and print a help message
+        print(e)  # noqa: T201
         print(  # noqa: T201
             f"Error: .zarr storage not found at {path}. Please specify a valid OME-NGFF spatial data (.zarr) file. "
-            "Example "
-            '"python -m '
-            'spatialdata peek data.zarr"'
+            "Examples "
+            '"python -m spatialdata peek data.zarr"'
+            '"python -m spatialdata peek https://remote/.../data.zarr labels table"'
```
giovp, line 44: nice clarification

src/spatialdata/_core/spatialdata.py
```diff
@@ -1238,7 +1259,7 @@ def h(s: str) -> str:
                                 assert isinstance(t, tuple)
                                 assert len(t) == 1
                                 parquet_file = t[0]
-                                table = read_table(parquet_file)
+                                table = read_parquet(parquet_file)
```
giovp, line 1262: nice catch

src/spatialdata/_io/io_points.py
```diff
@@ -29,8 +30,10 @@ def _read_points(
     assert isinstance(store, (str, Path))
     f = zarr.open(store, mode="r")
 
-    path = Path(f._store.path) / f.path / "points.parquet"
-    table = read_parquet(path)
+    path = os.path.join(f._store.path, f.path, "points.parquet")
+    # cache on remote file needed for parquet reader to work
+    # TODO: allow reading in the metadata without caching all the data
+    table = read_parquet("simplecache::" + path if "http" in path else path)
```
giovp, line 36: googled this cause i didn't know what it should do and found this option https://github.com/catalyst-cooperative/pudl/issues/1496#issuecomment-1085033950 and https://github.com/intake/intake-parquet/issues/18#issuecomment-1092217888 , maybe better?

giovp (approved):
minor nitpick on type and a suggestion that might not be the exact same 😅

src/spatialdata/__main__.py
LucaMarconato: Good idea! Done.

src/spatialdata/_io/io_points.py
LucaMarconato: I think we would need to benchmark that. I'd do that in a follow up PR. @berombau WDYT?

LucaMarconato:
> Reported the bug in the anndata repo: https://github.com/scverse/anndata/issues/1056.

Fixed. 

Thanks for the review Giovanni, merging now.

PR #282, opened 2023-06-01
Fix blobs

### LINKED ISSUE #281: Validation error blobs dataset on windows
With the latest spatialdata version on main I get a validation error when trying to load the blobs dataset (writing tests in napari-spatialdata). This happens on a windows machine, did not check others yet. 

MWE
```
sdata=blobs()
```
Raises `AssertionError`

afbeelding

melonora:
This PR is a fix for the problem raised in #281 and thus would close #281. Specifically, the integer dtype has been specified to be `np.int64` for the coordinates of points. 
In addition, for the layer visibility PR it is required to have a test dataset which has an additional coordinate space. The `BlobsDataset has been adjusted to allow for an extra coordinate space with an `Identity` transform.

### DIFF
--- a/src/spatialdata/datasets.py
+++ b/src/spatialdata/datasets.py
@@ -1,6 +1,5 @@
 """SpatialData datasets."""
-from abc import ABC
-from typing import Any, Optional, TypeVar, Union
+from typing import Any, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -24,8 +23,6 @@ from spatialdata.models import (
     TableModel,
 )
 from spatialdata.transformations import Identity
-
-T = TypeVar("T", bound=ABC)
 
 __all__ = ["blobs", "raccoon"]
 
--- a/src/spatialdata/datasets.py
+++ b/src/spatialdata/datasets.py
@@ -45,6 +45,9 @@ def blobs(
     n_shapes
         Number of max shapes to generate.
         At most, as if overlapping they will be discarded
+    extra_coord_space
+            Extra coordinate space on top of the standard global coordinate space. Will have only identity transform.
+
 
     Returns
     -------
@@ -98,7 +101,7 @@ class BlobsDataset:
         n_shapes
             Number of max shapes to generate.
             At most, as if overlapping they will be discarded
-        extra_space
+        extra_coord_space
             Extra coordinate space on top of the standard global coordinate space. Will have only identity transform.
         """
         self.length = length
--- a/src/spatialdata/datasets.py
+++ b/src/spatialdata/datasets.py
@@ -31,7 +31,7 @@ __all__ = ["blobs", "raccoon"]
 
 
 def blobs(
-    length: int = 512, n_points: int = 200, n_shapes: int = 5, extra_coord_space: Optional[str] = None
+    length: int = 512, n_points: int = 200, n_shapes: int = 5, extra_coord_system: Optional[str] = None
 ) -> SpatialData:
     """
     Blobs dataset.
@@ -45,7 +45,7 @@ def blobs(
     n_shapes
         Number of max shapes to generate.
         At most, as if overlapping they will be discarded
-    extra_coord_space
+    extra_coord_system
             Extra coordinate space on top of the standard global coordinate space. Will have only identity transform.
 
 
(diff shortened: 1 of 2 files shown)


melonora:
I had issues with mypy and the type hints for transforms. To circumvent this I type hint transforms as `dict[str, Any]` for now. Type hinting as either `Identity` or `BaseTransformation` does not work as `Identity` is not recognized as a `BaseTransformation`.

melonora:
@giovp @LucaMarconato Small PR, if you could have a look please.

LucaMarconato:
Looks good to me! One comment, in napari one would need a test that tries changing the coordinate system in the presence of an element that not only has 2 coordinate systems (as in this PR), but also the transformation is different for each system.

An option would be to add an affine transformation instead of an identity, but this could also be done in-place in the test after getting `blobs` as a test fixture (which maybe would keep things cleaner). Please choose what you think it's more appropriate.

LucaMarconato (approved):

melonora:
> Thanks for the PR @melonora, I will now look at the code changes.
> 
> Quick comment re the type hints, this should work: `your_dict: dict[str, type[BaseTransformation]] = {}`, could you try please?

This is indeed what I thought would have to work as well so tried that first, but got this:
afbeelding

melonora:
> Looks good to me! One comment, in napari one would need a test that tries changing the coordinate system in the presence of an element that not only has 2 coordinate systems (as in this PR), but also the transformation is different for each system.
> 
> An option would be to add an affine transformation instead of an identity, but this could also be done in-place in the test after getting `blobs` as a test fixture (which maybe would keep things cleaner). Please choose what you think it's more appropriate.

I think in place in a test would be cleaner. In particular because allows to keep it clean if not more than an Identy transform is needed. As you said it can always be extended with specific transforms in a test if required.

src/spatialdata/datasets.py
```diff
@@ -75,7 +86,9 @@ def raccoon(
 class BlobsDataset:
     """Blobs dataset."""
 
-    def __init__(self, length: int = 512, n_points: int = 200, n_shapes: int = 5) -> None:
+    def __init__(
+        self, length: int = 512, n_points: int = 200, n_shapes: int = 5, extra_coord_space: Optional[str] = None
```
giovp, line 90: I would call it `coord_system` and not space

src/spatialdata/datasets.py
```diff
@@ -171,21 +196,27 @@ def _generate_blobs(self, length: int = 512, seed: Optional[int] = None) -> Arra
...
             },
         )
-        return PointsModel.parse(arr, annotation=annotation, feature_key="genes", instance_key="instance_id")
+        return PointsModel.parse(
+            arr, transformations=transforms, annotation=annotation, feature_key="genes", instance_key="instance_id"
+        )
 
-    def _circles_blobs(self, length: int = 512, n_shapes: int = 5) -> GeoDataFrame:
+    def _circles_blobs(
+        self, transforms: Optional[dict[str, Any]] = None, length: int = 512, n_shapes: int = 5
```
giovp, line 218: why not call it `transformations` instead of transforms, as it is also called int he models? If you change it here, please change it everywhere

giovp (approved):
minor things on naming variables.

the `np.int_ -> np.int64` ok for me to rename, but thought it's pretty standard you to use underscore for full precision

src/spatialdata/datasets.py
melonora: resolved with [bef3a35](https://github.com/scverse/spatialdata/pull/282/commits/bef3a353b97f515e5561d8672c88b55252b7a675)

src/spatialdata/datasets.py
melonora: resolved with [bef3a35](https://github.com/scverse/spatialdata/pull/282/commits/bef3a353b97f515e5561d8672c88b55252b7a675)

melonora:
> minor things on naming variables.
> 
> the `np.int_ -> np.int64` ok for me to rename, but thought it's pretty standard you to use underscore for full precision

yes, the problem is that it can evaluate to a np.int32 type. The other option would be to add np.int32 to the allowed dtypes.

melonora:
@giovp @LucaMarconato For fixing the type hint, the code has to be changed throughout transforms. It basically affects every transform. Perhaps lets merge this PR for now if you both agree and then I can open the layer visibility PR with tests.

PR #283, opened 2023-06-03
Tranformation copy

melonora:
Currently it is the case that a user creates a transformation dictionary which is used while parsing multiple data sources into spatialdata elements. When on top of that setting an additional transformation later for one specific element, all the other elements have the transformation set as well. As suggested by @LucaMarconato, this PR fixes this issue by creating a dictionary copy in the parse function if `transformations` is not `None`. Also, a test has been added.


LucaMarconato:
Perfect thanks a lot!

PR #307, opened 2023-06-22
specify n_channels

melonora:
This is for writing tests in napari-spatialdata where for RGB it is required to specify the number of channels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ spatialdata-sandbox
 
 # version file
 _version.py
+
+# prevent submodule notebooks getting out of sync
+docs/tutorials/notebooks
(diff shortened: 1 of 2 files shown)


melonora:
@LucaMarconato @kevinyamauchi Could one of you review this please and then release? The test for RGB in napari-spatialdata depends on it.

kevinyamauchi (approved):
LGTM, thanks!

@giovp , can you please take a look? I don't understand the notebook submodule diff. Why is that showing up here? Is it just saying that the notebooks repo was updated separately and those changes will be committed here now?

giovp:
should be good but to be extra sure @melonora could you fetch and pull from the submodule and push again with the latest version? then you shouldn't have those diffs. I think we need to add that folder to gitignore to prevent this.

kevinyamauchi:
Thanks for the update, @melonora ! The diff now looks good. I will merge now.

PR #309, opened 2023-06-26
Fix c coords

### LINKED ISSUE #295: `transform` converts string indicies to numbers
I am trying to fix an issue in which @berombau wants to visualise indiviudal channels of an image by their string-encoded index, discussio here: https://github.com/scverse/spatialdata-plot/issues/92

I am unable to fix this on the side of `spatialdata_plot`, but was able to trace it back to `spatialdata`. It seems that inside of the `transform` function, all potential string indicies are just converted to numbers, as far as I could see that through `range(len(x))`.

### Steps to reproduce:
Before transform
```
import spatialdata as sd
import numpy as np

sdata = sd.SpatialData()
sdata.add_image(
    'img',
    image = sd.models.Image2DModel.parse(
        data = np.random.rand(4, 100, 100),
        dims = ['c', 'y', 'x'],
        c_coords = ['c1', 'c2', 'c3', 'c4']
    )
)
print(sdata.images["img"])
```
image

After transform
```
from spatialdata.transformations import get_transformation
from spatialdata import transform

t = get_transformation(sdata.images["img"])
sdata.images["img"] = transform(sdata.images["img"], t)
print(sdata.images["img"])
```
image

berombau:
Closes #295 

- pass c_coords also to newly transformed SpatialImages
- add c_coords to new blobs dataset for debugging
- add tests for blobs dataset with both numerical and string c_coords

Now spatialdata_plot works as expected.
```python
import spatialdata_plot

sdata = blobs(c_coords=['a', 'b'])
sdata.pl.render_images('blobs_image', channel=['a', 'b'], palette=['red', 'blue']).pl.show()
```

### DIFF
--- a/src/spatialdata/_core/operations/transform.py
+++ b/src/spatialdata/_core/operations/transform.py
@@ -242,8 +242,9 @@ def _(data: SpatialImage, transformation: BaseTransformation, maintain_positioni
     transformed_dask, raster_translation = _transform_raster(
         data=data.data, axes=axes, transformation=transformation, **kwargs
     )
+    c_coords = data.indexes["c"].values if "c" in data.indexes else None
     # mypy thinks that schema could be ShapesModel, PointsModel, ...
-    transformed_data = schema.parse(transformed_dask, dims=axes)  # type: ignore[call-arg,arg-type]
+    transformed_data = schema.parse(transformed_dask, dims=axes, c_coords=c_coords)  # type: ignore[call-arg,arg-type]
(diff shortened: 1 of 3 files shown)


giovp (approved):
great catch @berombau thank you! approve on my side, maybe @LucaMarconato can take a quick look?

LucaMarconato:
Thanks @berombau, looks great! I just made a very tiny modification (added a log when both `c_coords` and `n_channels` are specified in `blobs()`. Merging.

PR #316, opened 2023-07-11
Feature/graph loader

LucaMarconato:


LucaMarconato:
The previous commits are all from Giovanni's PR, the new code starts from here.

LucaMarconato:
Actually the code doesn't the one from the latest Giovanni's PR, so closing and will make a new one.

PR #318, opened 2023-07-11
get_extent() refactoring

LucaMarconato:
Starting a PR to rework and move the `get_extent()` function from @timtreis from `spatialdata-plot` to `spatialdata`. This PR is needed for some tiles generation code that @EliHei2 is working on.


LucaMarconato:
Actually it's completed and ready for review.

src/spatialdata/_core/data_extent.py
```diff
@@ -362,6 +330,7 @@ def _compute_extent_in_coordinate_system(
     df = pd.DataFrame(corners.data, columns=corners.axis.data.tolist())
     points = PointsModel.parse(df, coordinates={k: k for k in axes})
     transformed_corners = transform(points, transformation).compute()
-    min_coordinates = transformed_corners.min().to_numpy()
-    max_coordinates = transformed_corners.max().to_numpy()
+    # Make sure min and max values are in the same order as axes
+    min_coordinates = transformed_corners.min()[list(axes)].to_numpy()
```
Sonja-Stockhaus, line 334: This should fix the issue https://github.com/scverse/spatialdata/issues/335#issue-1842914360

LucaMarconato:
@Sonja-Stockhaus thanks for the refactoring! Please let me know when you finish with the additions and the PR is ready for review. You can tag me here or write me in Zulip.

src/spatialdata/_core/data_extent.py
```diff
@@ -157,7 +188,15 @@ def _(e: SpatialData, coordinate_system: str = "global") -> BoundingBoxDescripti
 
 
 @get_extent.register
-def _(e: GeoDataFrame, coordinate_system: str = "global") -> BoundingBoxDescription:
+def _(
+    e: GeoDataFrame,
+    coordinate_system: str = "global",
+    has_images: bool = True,
```
LucaMarconato, line 196: A suggestion, if possible I would remove these arguments from the single dispatch implementation of shapes, points and raster data, and I would leave it in the `SpatialData` implementation. I think that if you add a **kwags to the base declaration (the one for `SpatialData | SpatialElement`) and remove the arguments where not needed the syntax should be correct, please give it a try.

Sonja-Stockhaus: Sure, thanks for the feedback!

src/spatialdata/_core/data_extent.py
```diff
@@ -1,36 +1,325 @@
-"""Functions to compute the bounding box describing the extent of a SpatialElement or region."""
```
giovp, line 1: can we keep a description line here?

src/spatialdata/_core/data_extent.py
```diff
@@ -1,36 +1,325 @@
...
-    """Get the coordinates for the corners of the bounding box of that encompasses a circle element.
 
-    This computes the bounding box around all circles in the element.
+def _get_extent_of_circles(circles: GeoDataFrame) -> BoundingBoxDescription:
+    """
+    Compute the extent (bounding box) of a set of circles.
+
+    Parameters
+    ----------
+    circles
```
giovp, line 36: description missing

src/spatialdata/_core/data_extent.py
```diff
@@ -1,36 +1,325 @@
...
+    )
+    bounds["minx"] -= circles["radius"]
+    bounds["miny"] -= circles["radius"]
+    bounds["maxx"] += circles["radius"]
+    bounds["maxy"] += circles["radius"]
+
+    extent = {}
+    for ax in axes:
+        extent[ax] = (bounds[f"min{ax}"].min(), bounds[f"max{ax}"].max())
+    return extent
```
giovp, line 62:
```suggestion
    return {ax:(bounds[f"min{ax}"].min(), bounds[f"max{ax}"].max()) for ax in axes}
```

src/spatialdata/_core/data_extent.py
```diff
@@ -1,36 +1,325 @@
...
+    The bounding box description.
+    """
+    assert isinstance(shapes.geometry.iloc[0], (Polygon, MultiPolygon))
+    axes = get_axes_names(shapes)
+    bounds = shapes["geometry"].bounds
+    # NOTE: this implies the order x, y (which is probably correct?)
+    extent = {}
+    for ax in axes:
+        extent[ax] = (bounds[f"min{ax}"].min(), bounds[f"max{ax}"].max())
+    return extent
```
giovp, line 86: same as before

src/spatialdata/_core/data_extent.py
giovp: nitpick

src/spatialdata/_core/data_extent.py
```diff
@@ -1,36 +1,325 @@
...
+
+
+def _get_extent_of_data_array(e: DataArray, coordinate_system: str) -> BoundingBoxDescription:
+    # lightweight conversion to SpatialImage just to fix the type of the single-dispatch
```
giovp, line 90: we really need to figure this out 😅

src/spatialdata/_core/data_extent.py
```diff
@@ -1,36 +1,325 @@
...
+
+    Returns
+    -------
+    The bounding box description.
+    """
+    _check_element_has_coordinate_system(element=e, coordinate_system=coordinate_system)
+    # remove potentially empty geometries
+    e_temp = e[e["geometry"].apply(lambda geom: not geom.is_empty)]
+
+    # separate points from (multi-)polygons
```
LucaMarconato, line 209: This is not needed because we assume that the GeoDataFrame doesn't mix Point with Polygon/MultiPolygon types. I removed it to simplify the code.

src/spatialdata/models/_utils.py
```diff
@@ -167,9 +167,9 @@ def _(e: MultiscaleSpatialImage) -> tuple[str, ...]:
 
 @get_axes_names.register(GeoDataFrame)
 def _(e: GeoDataFrame) -> tuple[str, ...]:
-    all_dims = (X, Y, Z)
+    all_dims = (Z, Y, X)
```
LucaMarconato, line 170: @timtreis @Sonja-Stockhaus  I restored x, y, z since this is what also we do for points and since other parts of the framework (like napari) expect this. But we can consider changing it in another PR, what motivated the change here?

timtreis: Honestly don't remember, has been weeks. Either consistency with sth else or a bug

src/spatialdata/_core/data_extent.py
LucaMarconato: In the PR I kept the removal of empty points, but I have removed the split of shapes and points. It can be easily restored inside the function `_get_extent_of_shapes()` if you want. We don't expect to need it now, but it would be still ok to have, just to make the code more future proof.

LucaMarconato:
As you can see, the Python 3.9 tests fail in [`a0e26ba` (#318)](https://github.com/scverse/spatialdata/pull/318/commits/a0e26ba4f73c0d782a0b6ad7e727bf4e7ddd96eb), but they succeed in [`184dd82` (#318)](https://github.com/scverse/spatialdata/pull/318/commits/184dd824a831ce902943a601f33bc582bc6c16b6). Not sure why it happens, but as a workaround @timtreis suggested to replace the `|` with `Union`, and to silence Ruff.

PR #321, opened 2023-07-13
fix docs

LucaMarconato:
- Added installation instructions for M1/M2
- Added explicit links to the docs and to the repos in the docs landing page (feedback from @ilia-kats)

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -47,10 +47,7 @@ mamba create -n my_env python==3.10 -y
 conda activate my_env
 
 mamba install -c conda-forge napari -y
-pip install spatialdata-io
-pip install spatialdata-plot
-pip install napari-spatialdata
-pip install spatialdata
+pip install spatialdata spatialdata-io spatialdata-plot napari-spatialdata
 ```
 
 ## Development version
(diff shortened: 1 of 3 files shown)


docs/installation.md
```diff
@@ -46,9 +64,19 @@ Alternative you can clone the repository and do an editable install with:
 pip install -e .
 ```
 
-This is the reccommended way to install the package in case in which you want to contribute to the code. To update the package you can use `git pull`.
+This is the reccommended way to install the package in case in which you want to contribute to the code. In this case, to subsequently update the package you can use `git pull`.
```
melonora, line 67: In that case, wouldn't we prefer people to create a fork first?

MobiTobi:
```suggestion
An editable installation is the best option if you want to contribute to the development of spatialdata.
Check out our [contributing guide](https://spatialdata.scverse.org/en/latest/contributing.html).
```

explaining forking goes beyond the scope of installation instructions.
But we can use the opportunity to funnel people to the development guide.

melonora: Yes in any case for Numfocus we do require a contributors guide.

LucaMarconato: Thanks for the comments. It's hidden but it's there: https://github.com/scverse/spatialdata/blob/main/docs/contributing.md.

I will modify installation.md to quickly mentione forks.

Issue #314, opened 2023-06-29
Dev feature: decompose affine transformation into its component


LucaMarconato:
Convenience function needed by @timtreis for `spatialdata-plot` to fix https://github.com/scverse/spatialdata-plot/issues/103.

PR #327, opened 2023-07-25
decompose affine into simpler transformations

LucaMarconato:
Helper function for @timtreis as asked in https://github.com/scverse/spatialdata/issues/314.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,10 @@ If you found a bug, please use the [issue tracker][issue-tracker].
 [L Marconato*, G Palla*, KA Yamauchi*, I Virshup*, E Heidari, T Treis, M Toth, R Shrestha, H Vöhringer, W Huber, M Gerstung, J Moore, FJ Theis, O Stegle, bioRxiv, 2023](https://www.biorxiv.org/content/10.1101/2023.05.05.539647v1). \* = equal contribution
 
 ## Sponsor
+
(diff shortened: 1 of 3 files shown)


LucaMarconato:
Please check the docstring of the `_decompose_transformation()` for a detailed explanation. Please let me know if you need more use cases or a different type of representation (like combining "shear" and "rotation" into a unique affine matrix, or forcing "shear" having diagonal values all 1; things like these).

LucaMarconato:
I added some tests, but at each run the function is also doing some additional lightweight checks before returning, so if there are bugs we catch them.

LucaMarconato:
Please also have a look at the much simpler `_compose_affine_from_linear_and_translation()`. For the new transformations refactoring I will need the new function of this PR because I want to separate operations on the coordinates and operations on the raw data that needs to be committed. But for plotting this PR may be overkill (or I may need to make a simpler function).

LucaMarconato:
@timtreis I made some changes, I added the parameter `simple_decomposition` to `_decompose_transformation()` and I believe that this is what you need (not `_decompose_affine_into_linear_and_translation()). Thought about it and the original decomposition (`simple_decomposition=False`) is probably not useful for matplotlib plotting.

LucaMarconato:
Other PR have more priorities for code review. Gonna merge since the PR is thoroughly tested and since I am going to test it manually again when refactoring the transformations using this code. CC @kevinyamauchi @giovp

Issue #293, opened 2023-06-09
Refactoring the io and in-memory lazy vs non-lazy representation


LucaMarconato:
Linked issues:

Lazy vs non-lazy
https://github.com/scverse/spatialdata/issues/243

LucaMarconato:
@lopollar reported that saving extra columns in shapes layers is not working at the moment https://github.com/scverse/spatialdata/issues/311. I think it was working but we didn't have a test for that.

Please @giovp when working on this PR also add a test for this.

LucaMarconato:
Refactoring the geometry argument of the shapes model: https://github.com/scverse/spatialdata/issues/315

giovp:
question regarding incremental, plan is to remove entirely the write/read steps when `__setitem__` is called, unless `overwrite=True`, in that case a call to `write` follows and hence makes sense to reload data lazy?

LucaMarconato:
I'd maybe remove `add_image()`, etc and leave only `__setitem__`. In this way there is no option to pass `overwrite=True`. I would then add a method to resave to disk a specifc element. So the workflow would be:
```python
sdata.images['my_image'] = im
# or even the following, that is already implemented and uses single dispatch to forward the right `__setitem__`
sdata['my_image'] = im
sdata.write('my_data.zarr')
# example of in-place operation (can be both lazy or in-memory)
sdata['my_image'] = sdata['my_image'] + 1
sdata.write(element='my_image')
```
We could add a parameter `reload=True` to `write()` that does the lazy re-reading, and let the user disable it for specific cases, like when the data is all full in-memory and the user wants to keep it there.

LucaMarconato:
I'd then have something like this to load an element in-memory.
```python
# all is read lazy by default
sdata = read_zarr('my_data.zarr')
sdata.load_in_memory(element='my_image')
```

LucaMarconato:
We should also consider adding the option to keep data lazy but making it persistent (`.persist()`).

PR #329, opened 2023-07-30
fix incremental update

giovp:
starting to work on #293

this PR only partially address the issue, in particular it implements the following:
- enforce uniqueness of names of elements, by introducing the `Elements` class
- remove lazy vs. in-memory logic for adding elements.


giovp:
heads up, had to introduce a new module 😬  called "elements" (happy to change name).
Thing is, the `__setitem__` method should be consistent whether a new element is added directly to spatialdata, or to the elements dict.
e.g.
```python
sdata = SpatiaData(images={"myimage":image})
sdata["myimage"] = image # raise error
sdata.images["myimage"] = image # did not raise error
```

I must say I'm not 100% sure this is expected behaviour but at least it is consistent, wdyt? @LucaMarconato

giovp:
I've started with option 1 from zulip but now the following example doesn't throw warning
```python
sdata = SpatiaData(images={"myimage":image}, labels={"mylabel":label})
sdata["myimage"] = newimage
>>> UserWarning
sdata.labels["myimage"] = newlabel # does not throw warning, but it should..
```
need to think of best option

LucaMarconato:
@giovp still draft or ready for review?

giovp:
still draft but very close!

giovp:
@LucaMarconato @kevinyamauchi this is ready for review !

melonora (commented):
Just some minor questions / comments. Overall looks good to me and in fact a lot cleaner!

melonora:
Also upon testing with napari spatialdata, the userwarning is given every time you click on a coordinate space. I will look into it what causes this behaviour.

melonora:
The problem originates from `_core/spatialdata.py`, particularly `filter_by_coordinate_system` returning the filtered `SpatialData` object.

src/spatialdata/_core/_elements.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Elements(OrderedDict[str, Any]):
+    _shared_keys: set[str] = set()
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        if key in self._shared_keys:
+            warn(f"Key `{key}` already exists.", UserWarning, stacklevel=2)
+        super().__setitem__(key, value)
+        self._shared_keys.add(key)
```
melonora, line 33:
```suggestion
class Elements(OrderedDict[str, Any]):

    def __init__(self):
        super().__init__()
        self._shared_keys = set()

    def __setitem__(self, key: str, value: Any) -> None:
        if key in self._shared_keys:
            warn(f"Key `{key}` already exists.", UserWarning, stacklevel=2)
        super().__setitem__(key, value)
        self._shared_keys.add(key)
```
This makes `_shared_keys` an instance attribute instead of a class attribute which ensures that we do not spam warnings when returning a filtered `SpatialData` object.

melonora (changes requested):

giovp:
this should be ready to give it another go @melonora , thanks again for the review!

> just for others after some small QA with Giovanni, in case that the scope of this PR is to also allow for partial writes to disk, this should still be implemented as I believe this PR does not do this yet. In case it is implemented it is currently not covered by the tests yet.

I totally agree but I feel it would become a rather big PR, I think what has been done here is the following:
- enforce uniqueness of names of elements
- remove lazy vs. in-memory logic for adding elements. 

In a separate PR I would then add options to load and save elements only. wdyt? also @LucaMarconato ?

giovp:
I'm afraid there is no easy way to enforce uniqueness of elements names across spatialdata, unless additional complexity is added at the main spatialdata class itself. I think we'd need to essentially subclass a `Mapping` with the attributes bein mapping as well, and when `__setitem__` is called, then it would check across elements and raise a warning.

A much simpler option would be to enforce uniqueness only within element type, such as something like this would not raise a warning
```python
sdata = SpatialData(images={"temp":im})
sdata.labels["temp"] = lab
```

but something like this would:

```python
sdata = SpatialData(images={"temp":im})
sdata.images["temp"] = im
```
THe uniqueness of the element name is therefore true at the spatialdata class level but _implicitly_. On disk instead, it will be resolved by relative paths to the elements, as it is now.

```
data.zarr
  -- images/A
  -- labels/A
```

wdyt? @melonora @LucaMarconato . I think this might be the easy solution and it kind of reflects already the expected behaviour.

melonora:
> this should be ready to give it another go @melonora , thanks again for the review!
> 
> > just for others after some small QA with Giovanni, in case that the scope of this PR is to also allow for partial writes to disk, this should still be implemented as I believe this PR does not do this yet. In case it is implemented it is currently not covered by the tests yet.
> 
> I totally agree but I feel it would become a rather big PR, I think what has been done here is the following:
> 
> * enforce uniqueness of names of elements
> * remove lazy vs. in-memory logic for adding elements.
> 
> In a separate PR I would then add options to load and save elements only. wdyt? also @LucaMarconato ?

Yeah, I agree. I just wasn't certain what the original scope of the PR was supposed to be. Better to keep it small for easier reviewing.

melonora:
> I'm afraid there is no easy way to enforce uniqueness of elements names across spatialdata, unless additional complexity is added at the main spatialdata class itself. I think we'd need to essentially subclass a `Mapping` with the attributes bein mapping as well, and when `__setitem__` is called, then it would check across elements and raise a warning.
> 
> A much simpler option would be to enforce uniqueness only within element type, such as something like this would not raise a warning
> 
> ```python
> sdata = SpatialData(images={"temp":im})
> sdata.labels["temp"] = lab
> ```
> 
> but something like this would:
> 
> ```python
> sdata = SpatialData(images={"temp":im})
> sdata.images["temp"] = im
> ```
> 
> THe uniqueness of the element name is therefore true at the spatialdata class level but _implicitly_. On disk instead, it will be resolved by relative paths to the elements, as it is now.
> 
> ```
> data.zarr
>   -- images/A
>   -- labels/A
> ```
> 
> wdyt? @melonora @LucaMarconato . I think this might be the easy solution and it kind of reflects already the expected behaviour.

I would be ok with this option, but that does mean that acces like this should be disallowed as it could lead to conflicts.:
```
sdata["my_array"]
```

giovp:
> I would be ok with this option, but that does mean that acces like this should be disallowed as it could lead to conflicts.:

> sdata["my_array"]

I agree!

giovp:
I quickly wanted to fix this and found a straightforward solution. 
Please check this minimal behaviour

```python
from spatialdata.models import Image2DModel, Labels2DModel
import numpy as np

im = Image2DModel.parse(np.random.normal(size=(3,50,50)))
lab = Labels2DModel.parse(np.random.normal(size=(50,50)))
sdata = SpatialData()
sdata.images._shared_keys
>>> set()
```
if the spatialdata is empty, there are not shared keys across elements

```python
sdata = SpatialData(images={"temp":im})#, labels={"mylabel":lab})
sdata.labels._shared_keys
>>> {"temp"}
```
now the key "temp" is stored in the shared_keys attribute for all elements.

now, in the case where the user set an items with key `"temp"`
- if "temp" is a key in the element type the user is trying to set, return warning and exectue -> previous element is lost
- if "temp" is a key in an element type that is different from the one the user is trying to set -> keyerror, fails

```python
sdata = SpatialData(images={"temp":im})#, labels={"mylabel":lab})
sdata.images["temp"] = im
>>> warning
sdata.labels["temp"] = lab
>>> keyerror, fails
```

I think this is intuitive behaviour, and it was also the result of the discussion we had on imagesc zulip some time ago, but happy to discuss more.

src/spatialdata/_core/_elements.py
```diff
@@ -0,0 +1,104 @@
...
+    ShapesModel,
+    get_axes_names,
+    get_model,
+)
+
+
+class Elements(UserDict[str, Any]):
+    def __init__(self, shared_keys: set[str | None]) -> None:
+        self._shared_keys = shared_keys
+        super().__init__()
```
melonora, line 30:
```suggestion
class Elements(UserDict[str, Any]):
    _shared_keys = set
    def __init__(self, shared_keys: set[str | None]) -> None:
        super().__init__()
```
Would there be a downside to making `_shared_keys` a class attribute here?

melonora (changes requested):
I would like to know if there are any downsides to actually using a class attribute here for `_shared_keys`. Also more from the perspective of clarity, would a name like `_existing_keys` be more clear? The keys are not actually shared, in fact they should not be shared as they should be unique.

src/spatialdata/_core/_elements.py
```diff
@@ -0,0 +1,104 @@
...
+    Labels2DModel,
+    Labels3DModel,
+    PointsModel,
+    ShapesModel,
+    get_axes_names,
+    get_model,
+)
+
+
+class Elements(UserDict[str, Any]):
```
kevinyamauchi, line 27: Nitpick, but why a `UserDict` here instead of a regular dict?

kevinyamauchi (commented):
LGTM. I like the additional `Elements` class. This could be useful for shimming other actions and validators in the future. Thanks!

kevinyamauchi (approved):
Oops, last review was meant to be an approval

src/spatialdata/_core/_elements.py
LucaMarconato: Looks good to me.

src/spatialdata/_core/_elements.py
LucaMarconato: The guidelines seem to be that one subclasses `dict` to extend the interface and one subclasses `UserDict` to override methods, like `__setitem__()` https://stackoverflow.com/a/7148602/3343783. So I would keep `UserDict`.

LucaMarconato:
> I would like to know if there are any downsides to actually using a class attribute here for _shared_keys. Also more from the perspective of clarity, would a name like _existing_keys be more clear? The keys are not actually shared, in fact they should not be shared as they should be unique.

I agree, I would rename to `_existing_keys`, wdyt @giovp?

src/spatialdata/_core/_elements.py
LucaMarconato: The suggestion from Wouter can't be applied because the code is outdated, but I also thing that `_shared_keys` (or `_existing_keys` if it gets renamed) should be an instance method and not a class method, because this could lead to wrong behaviors when multiple `SpatialData` objects are being handled at the same time.

src/spatialdata/_io/io_raster.py
```diff
@@ -67,7 +66,8 @@ def _read_multiscale(
     # and for instance in the xenium example
     encoded_ngff_transformations = multiscales[0]["coordinateTransformations"]
     transformations = _get_transformations_from_ngff_dict(encoded_ngff_transformations)
-    name = os.path.basename(node.metadata["name"])
+    # TODO: what to do with name? For now remove?
+    # name = os.path.basename(node.metadata["name"])
```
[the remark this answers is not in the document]
LucaMarconato: I agree with @giovp in removing this, using a custom name has two downsides:
1) it makes the access to tensor inside the xarray object more verbose and non uniform, for instance one needs to write something like `sdata['my_image_name']['scale0']['my_image_name']`, instead of the standardized `sdata['my_image_name']['scale0']['image']`.
2) if the user assignes the image to a different name, like with `sdata['my_image2'] = sdata['my_image_name']`, the internal name will still renamin the old one, which is confusion.

@giovp for the sake of brevity I would actually call the name `im` instead of `image`; also in the case of labels I think `im` sounds better than `image`.

tests/conftest.py
```diff
@@ -138,7 +138,10 @@ def _get_images() -> dict[str, Union[SpatialImage, MultiscaleSpatialImage]]:
     dims_3d = ("z", "y", "x", "c")
     out["image2d"] = Image2DModel.parse(RNG.normal(size=(3, 64, 64)), dims=dims_2d, c_coords=["r", "g", "b"])
     out["image2d_multiscale"] = Image2DModel.parse(
-        RNG.normal(size=(3, 64, 64)), scale_factors=[2, 2], dims=dims_2d, c_coords=["r", "g", "b"]
+        RNG.normal(size=(3, 64, 64)),
```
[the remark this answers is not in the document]
LucaMarconato: Here rgba would not test extra things. There are other places in the codebase where this would make a difference though: in napari-spatialdata and spatialdata-plot. I suggest to keep it the test as it is now and to test the rgba when we will merge the relative PRs from Andreas (this one https://github.com/scverse/napari-spatialdata/pull/153 and the ones that are linked in the PR).

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -42,14 +39,14 @@ def test_element_names_unique():
...
 
-    with pytest.raises(ValueError):
-        sdata.add_image(name="points", image=image)
-    with pytest.raises(ValueError):
-        sdata.add_points(name="image", points=points)
-    with pytest.raises(ValueError):
-        sdata.add_shapes(name="image", shapes=shapes)
-    with pytest.raises(ValueError):
-        sdata.add_labels(name="image", labels=labels)
+    with pytest.warns(UserWarning):
```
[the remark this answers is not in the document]
LucaMarconato: Yes, please opena an issue for this; we should consider adding a Windows CI at some point in the next months.

src/spatialdata/_core/_elements.py
LucaMarconato: Ah, the latest code already uses instance attributes, so all good here.

src/spatialdata/_core/spatialdata.py
```diff
@@ -396,13 +304,8 @@ def _get_group_for_element(self, name: str, element_type: str) -> zarr.Group:
...
     def _init_add_element(self, name: str, element_type: str, overwrite: bool) -> zarr.Group:
-        if self.path is None:
-            # in the future we can relax this, but this ensures that we don't have objects that are partially backed
-            # and partially in memory
-            raise RuntimeError(
-                "The data is not backed by a Zarr storage. In order to add new elements after "
-                "initializing a SpatialData object you need to call SpatialData.write() first"
-            )
+        # if self.path is None:
+        #     warn("Rasterizing points and shapes is not supported yet. Skipping.", UserWarning, stacklevel=2)
```
LucaMarconato, line 308: This comment seems misplaced, I'd remove.

src/spatialdata/_core/spatialdata.py
```diff
@@ -263,115 +256,30 @@ def aggregate(
...
-        if name in self._points and not overwrite:
-            raise KeyError(f"Points {name} already exists in the dataset.")
-        Point_s.validate(points)
-        self._points[name] = points
-
     def is_backed(self) -> bool:
         """Check if the data is backed by a Zarr storage or if it is in-memory."""
         return self.path is not None
 
+    @property
```
LucaMarconato, line 267: <todo add comment>

src/spatialdata/_core/spatialdata.py
```diff
@@ -682,292 +585,6 @@ def transform_to_coordinate_system(
             elements[element_type][element_name] = transformed
         return SpatialData(**elements, table=sdata.table)
 
-    def add_image(
```
LucaMarconato, line 685: Let's keep in mind that this PR removed the commented code here, since it will probably be useful for making the "lazy copy" approach, or similar, work in the future.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1044,8 +661,7 @@ def write(
                     # reload the image from the Zarr storage so that now the element is lazy loaded,
                     # and most importantly, from the correct storage
                     element_path = Path(self.path) / "images" / name
-                    image = _read_multiscale(element_path, raster_type="image")
-                    self._add_image_in_memory(name=name, image=image, overwrite=True)
+                    _read_multiscale(element_path, raster_type="image")
```
LucaMarconato, line 664: I don't get it, why do you do `_read_multiscale(element_path, raster_type="image")` without assigning it to anything? I think it should be `sdata[name] = _read_multiscale(element_path, raster_type="image")`. Same comment for labels, shapes and points.

LucaMarconato: Ah ok I get it, the old code was for reloading lazily those data that could be lazy loaded. Here we should remove those lines of code.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1216,25 +829,49 @@ def read(file_path: str, selection: tuple[str] | None = None) -> SpatialData:
         return read_zarr(file_path, selection=selection)
 
     @property
-    def images(self) -> dict[str, SpatialImage | MultiscaleSpatialImage]:
+    def images(self) -> Images:
         """Return images as a Dict of name to image data."""
         return self._images
 
+    @images.setter
```
LucaMarconato, line 836: This is equivalent to a `dict.extend()`, while if I called a `sdata.images = new_images` I would expect a complete replacemnet of the dict. I would change this to reflect this expected behavior. Same for labels, points and shapes.

src/spatialdata/_core/spatialdata.py
LucaMarconato: But for the moment I would remove the _read_multiscale() lines and the _read_points() line.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1023,7 +640,7 @@ def write(
         #     self.path = str(file_path)
         # else:
         #     self.path = str(tmp_zarr_file)
-        self.path = str(file_path)
+        self.path = Path(file_path)
```
LucaMarconato, line 643: I have a comment on this. See GitHub discussion.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Mmm, actually maybe it's not so easy, I have a comment on this. See GitHub discussion.

src/spatialdata/_core/spatialdata.py
```diff
@@ -263,115 +256,30 @@ def aggregate(
...
     def is_backed(self) -> bool:
         """Check if the data is backed by a Zarr storage or if it is in-memory."""
         return self.path is not None
 
+    @property
+    def path(self) -> Path | None:
+        """Path to the Zarr storage."""
+        return self._path
+
+    @path.setter
```
LucaMarconato, line 272: I have a comment on this setter, not sure if we should disallow it. See GitHub discussion. (I added a similar "See GitHub discussion" comment in other places, they all refer to the same discussion.)

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -25,31 +22,58 @@
...
+    with pytest.raises(KeyError):
+        sdata.images["points"] = image
+    with pytest.raises(KeyError):
+        sdata.images["shapes"] = image
+    with pytest.raises(KeyError):
+        sdata.labels["points"] = labels
+    with pytest.raises(KeyError):
+        sdata.points["shapes"] = points
+    with pytest.raises(KeyError):
+        sdata.shapes["labels"] = shapes
```
LucaMarconato, line 64: I think some cases are missing, but probably no need to add them all; I would still rearrange it a bit, more maybe, not it isn't immediate which cases are tested and which not.

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -25,31 +22,58 @@
...
+    assert sdata["labels"].shape == labels.shape
+    assert len(sdata["points"]) == len(points)
+    assert sdata["shapes"].shape == shapes.shape
+
+    with pytest.raises(KeyError):
+        sdata["labels"] = image
+    with pytest.warns(UserWarning):
+        sdata["points"] = points
+
+    sdata["image2"] = image
```
LucaMarconato, line 77: maybe add comment

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -25,31 +22,58 @@
...
+
+    assert sdata["image"].shape == image.shape
+    assert sdata["labels"].shape == labels.shape
+    assert len(sdata["points"]) == len(points)
+    assert sdata["shapes"].shape == shapes.shape
+
+    with pytest.raises(KeyError):
+        sdata["labels"] = image
+    with pytest.warns(UserWarning):
+        sdata["points"] = points
```
LucaMarconato, line 74: same comment as above, is it ok that not all the combinations are tested? If yes please add it as a comment.

tests/core/operations/test_spatialdata_operations.py
LucaMarconato: Or if it is ok that not all the combinations are tested I would write this in a comment.

tests/io/test_readwrite.py
```diff
@@ -275,45 +276,6 @@ def test_overwrite_onto_non_zarr_file(self, full_sdata):
...
-            full_sdata.add_shapes("new_shapes", full_sdata.shapes["new_shapes"], overwrite=True)
-
-            # commenting out as it is failing
-            # f2 = os.path.join(tmpdir, "data2.zarr")
-            # sdata2 = SpatialData(table=full_sdata.table.copy())
-            # sdata2.write(f2)
-            # del full_sdata.table
-            # full_sdata.table = sdata2.table
-            # full_sdata.write(f2, overwrite=True)
-
```
LucaMarconato, line 316: Let's remember that this code was removed here because we will need to add similar tests when working with incremental IO or with lazy copy.

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -282,17 +306,15 @@ def test_get_item(points):
...
+def test_set_item(full_sdata: SpatialData) -> None:
     for name in ["image2d", "labels2d", "points_0", "circles", "poly"]:
         full_sdata[name + "_again"] = full_sdata[name]
-        with pytest.raises(KeyError):
+        with pytest.warns(UserWarning):
+            full_sdata[name] = full_sdata[name]
+    for name in ["image2d", "labels2d", "points_0"]:
+        # trying to overwrite the file used for backing (only for images, labels and points)
+        with pytest.warns(UserWarning):
             full_sdata[name] = full_sdata[name]
```
LucaMarconato, line 317: to be removed since we already test this above and we don't risk overwriteing lazy data now with the __setitem__ call.

LucaMarconato (approved):

src/spatialdata/_core/spatialdata.py
LucaMarconato: Yes, I think this setter should not be public.

src/spatialdata/_core/spatialdata.py
LucaMarconato: I think we shuold replace this with `self._path = Path(file_path)`

giovp:
thanks @LucaMarconato for the comment, I agree on the plan. 
- I have fixed the setter behavior for images/labels/points/shapes and added tests
- I have removed the path setter and removed the re-reading after write of the elements.

PR #342, opened 2023-08-24
Fix missing c_coords argument in blobs multiscales

aeisenbarth:
This change is needed for tests in https://github.com/scverse/napari-spatialdata/pull/153 to pass.

(You could seriously consider a monorepository that keeps related code and discussion together, but still release code as separate packages.)


LucaMarconato:
Thanks a lot for the contribution, looks good to me. We never thought about a mono-repository actually; not sure about changing now, it would be quite some work, but we will think about it as it would definitely improve the discussions. 

Also, we will certainly put effort to improve the documentation since it's scattered around the repos, and maybe put it all under one link as it is dispersive at the moment.

PR #344, opened 2023-08-29
Allow force-overwriting existing files (non-backing)

aeisenbarth:
When I want to create and force-write SpatialData to some location, not caring whether there exist older files, I would intuitively use the `overwrite=True` argument. A use case is when developing/retrying a script that is supposed to be idempotent, repeatable no matter whether it has been run previously.

However, the code assumes from existing files that the SpatialData is backed (although `path` is None) and refuses to overwrite the supposed backing file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-## [0.0.15] - tbd
+## [0.1.0] - tbd
+
+#### Minor
+
+-   improved usability and robustness of sdata.write() when overwrite=True @aeisenbarth
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning][].
 
 ## [0.1.0] - tbd
 
-#### Minor
+### Minor
 
 -   improved usability and robustness of sdata.write() when overwrite=True @aeisenbarth
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-## [0.1.0] - tbd
+## [0.0.x] - tbd
 
 ### Minor
 
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -571,6 +571,7 @@ class SpatialData:
         consolidate_metadata: bool = True,
     ) -> None:
         from spatialdata._io import write_image, write_labels, write_points, write_shapes, write_table
+        from spatialdata._io._utils import get_dask_backing_files
 
         """Write the SpatialData object to Zarr."""
         if isinstance(file_path, str):
@@ -583,6 +584,7 @@ class SpatialData:
         # old code to support overwriting the backing file
         # target_path = None
         # tmp_zarr_file = None
+
         if os.path.exists(file_path):
             if parse_url(file_path, mode="r") is None:
                 raise ValueError(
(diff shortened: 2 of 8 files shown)


giovp:
thanks @aeisenbarth for this PR! I think it's related to #329 , will take a look over weekend!

tests/io/test_readwrite.py
```diff
@@ -238,6 +238,18 @@ def test_replace_transformation_on_disk_non_raster(self, shapes, points):
...
+        with tempfile.TemporaryDirectory() as tmpdir:
+            f = os.path.join(tmpdir, "data.zarr")
+            old_data = SpatialData()
+            old_data.write(f)
+            # Since not backed, no risk of overwriting backing data. Should not raise.
+            try:
+                full_sdata.write(f, overwrite=True)
+            except Exception as e:
+                assert "The file path specified is the same as the one used for backing" not in str(e)
+                raise  # Let other unexpected exceptions bubble up
```
melonora, line 251:
```suggestion
            full_sdata.write(f, overwrite=True)
```
What is the reason for the `try` `except` construct here? It is expected that this would not throw an error and if it does the test should fail. This would work already this way without the `try` `except` no? I understand that in this case you specifically want to test overwriting without backed data and only fail if this does not work, but I am just not certain whether it is worth it to write this block for it instead of just `full_sdata.write(f, overwrite=True)` in which case it could throw all errors which would be useful information in any case.

melonora (commented):
I do not entirely agree with the test, but for the rest, it looks good to me:)

tests/io/test_readwrite.py
aeisenbarth: I agree that the try/except is redundant. I used it to be more explicit about which exception of the previous behavior should not occur anymore by design (versus other unexpected exceptions).

LucaMarconato:
@aeisenbarth it took me a while to get back to this, it's a delicate part of the code so I wanted to examine it with care and I couldn't find the time before. 

The use case that you described is important and the logic that you implemented using the function `self.is_backed()` is valid, but unfortunately as it is the code can lead to some data loss.

This is due to the fact that `self.is_backed()` only checks if the data was written to disk using `write()` (this is when self.path` is set to a non-`None` value), but it is possible for the user to construct an in-memory object, thus non-backed, by combining together backed files. In such case, if the user tries to overwrite a backed file it would lead to a data loss.

Anyway, this has a simple solution, which is to modify the `is_backed()` function to check if the in-memory object depends on some storage files by looking at the computational graph instead than looking at `self.path`. I will make a patch in this PR.

LucaMarconato:
> Anyway, this has a simple solution, which is to modify the is_backed() function to check if the in-memory object depends on some storage files by looking at the computational graph instead than looking at self.path. I will make a patch in this PR.

Actually, instead of modifying `SpatialData.is_backed()` I have improved a function that already existed and made it public; I have now named it `get_dask_backing_files()`. With this new function I have added the support for the case that you fixed in this PR by avoiding the risk of data loss for the case I introduced above.

LucaMarconato:
Ready for review, I kindly ask you to have a look.

Also CC-ing who previously reviewed @giovp @melonora

CHANGELOG.md
```diff
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-## [0.0.15] - tbd
+## [0.1.0] - tbd
```
giovp, line 11: I wouldn't release 0.1.0 just yet, there are too many changes around storage, data models and transformations that needs to happen, wdyt? @melonora @LucaMarconato @kevinyamauchi

src/spatialdata/_core/spatialdata.py
```diff
@@ -583,21 +584,30 @@ def write(
...
+                    "The file path specified is the same as the one used for backing. "
+                    "Overwriting the backing file is not supported to prevent accidental data loss."
+                    "We are discussing how to support this use case in the future, if you would like us to "
+                    "support it please leave a comment on https://github.com/scverse/spatialdata/pull/138"
+                )
+            if any(Path(fp).resolve().is_relative_to(file_path.resolve()) for fp in get_dask_backing_files(self)):
+                raise ValueError(
+                    "The file path specified is a parent directory of one or more files used for backing for one or "
+                    "more elements in the SpatialData object. You can either load every element of the SpatialData "
+                    "object in memory, either save to a different file path."
```
giovp, line 608:
```suggestion
                    "object in memory, or save the current spatialdata to a different path."
```

giovp (approved):
looks good to me, this will probably need to be reworked again when we implement incremental update.

CHANGELOG.md
LucaMarconato: Maybe for multiple tables let's make 0.1.0, but we could also wait for a all the big PRs being merged. I changed for the moment.

LucaMarconato:
Thanks for the review! Fixing the tests and merging!

Issue #347, opened 2023-08-31
Exceptions when adding labels to backed SpatialData


aeisenbarth:
When saving labels to a SpatialData that has already been written to disk (with or without `overwrite=True`), there are multiple exceptions.

### Example code

```python
import numpy as np
import spatialdata

sdata = spatialdata.SpatialData()
sdata.write("/tmp/spatialdata.zarr")
labels = spatialdata.models.Labels2DModel.parse(np.ones((2, 2)))
sdata.add_labels(name="my_label", labels=labels, overwrite=True)
```

## 1. UnboundLocalError: local variable 'elem_group' referenced before assignment

A variable is accessed  (L415) which is only conditionally declared. It would be safer either to assign a default value, or to access it only in the same conditional branch.

I don't understand the reasoning of the comment, but to me it seems a fix would be to remove the condition above L411.

https://github.com/scverse/spatialdata/blob/4dd525002f12381858251df9cd304a7108a21886/src/spatialdata/_core/spatialdata.py#L409-L416

```
Traceback (most recent call last):
  File "<ipython-input-6-6d9e217edb59>", line 1, in <module>
    sdata.add_labels(name="my_label", labels=labels, overwrite=True)
  File "…/spatialdata/_core/spatialdata.py", line 779, in add_labels
    elem_group = self._init_add_element(name=name, element_type="labels", overwrite=overwrite)
  File "…/spatialdata/_core/spatialdata.py", line 415, in _init_add_element
    if name in elem_group:
UnboundLocalError: local variable 'elem_group' referenced before assignment
```

## 2. AssertionError on type of dims

For backed labels, `xdata` is not a `DataArray` (where `dims` is `("y", "x")`) but a `DataTree` where `dims` is `Frozen({"y": 1633, "x": 1290})`.

https://github.com/scverse/spatialdata/blob/4dd525002f12381858251df9cd304a7108a21886/src/spatialdata/models/_utils.py#L151-L153

This works with:
```python
xdata = next(iter(e["scale0"].ds.values()))
```
or converting the dictionary to tuple:
```python
        if isinstance(dims_data, Mapping):
            dims_data = tuple(dims_data.keys())
```

## 3. DataTree has no attribute `shape`

https://github.com/scverse/spatialdata/blob/4dd525002f12381858251df9cd304a7108a21886/src/spatialdata/transformations/_utils.py#L118-L119

Also here:
```python
        xdata = next(iter(e["scale0"].ds.values()))
```

[...]

PR #348, opened 2023-08-31
[WIP] Resolve exceptions when adding labels to backed SpatialData

aeisenbarth:
This PR contains suggested changes for avoiding errors when adding labels to backed SpatialData (https://github.com/scverse/spatialdata/issues/347). With these changes, I can get my use case to run, however I am not sure about the consequences or whether these are the correct way to solve this. Please run the test case in the issue to inspect the errors.


aeisenbarth:
I think it's not that simple. Instead of modifying the locations in these commits (and causing further problems), the root of the problem needs to be fixed. I don't have enough understanding of the xarray DataTrees/DataArrays to know how their nesting should be.

There seems to be an inconsistency between the MultiscaleSpatialImage created in memory and one backed on disk. In a backed image the DataTrees are not correctly set up.

- Some places in the code expect a DataArray but receive a DataTree.
- `Rasterschema.validate(MultiscaleSpatialImage)` expects all scale levels to have a transform, but the backed image has only the transform in `scale0`.

LucaMarconato:
Thanks a lot for the PR, I will look into this.

aeisenbarth:
This issue is still occurring.

I updated the fix and removed the other commits unrelated to the original issue.

The following test case reproduces "local variable 'elem_group' referenced before assignment" on the current main branch (ef0a2dc7).
```python
import numpy as np
import pytest

from spatialdata.models import Labels2DModel


@pytest.mark.parametrize("overwrite", [False, True])
def test_add_labels(overwrite, tmp_path):
    sdata_path = tmp_path / "sdata.zarr"
    sdata = SpatialData()
    sdata.write(sdata_path)
    labels = Labels2DModel.parse(np.ones((10, 10)), dims=("y", "x"), scale_factors=(2, 2))
    # sdata.zarr/labels does not exist.
    # When using overwrite=True, it is assumed it exists.
    sdata.add_labels(name="labels1", labels=labels, overwrite=overwrite)

```
I marked the PR ready for review.

aeisenbarth:
Can someone review this?

`local variable 'elem_group' referenced before assignment` indicates invalid Python code.

I still cannot use an official spatialdata release in my dependencies because `0.0.15` has this.

aeisenbarth:
Thanks! The test was just meant to reproduce the issue.

This issue is now obsolete. The current main branch removed support for `overwrite` (added elements are not anymore written them to disk, so no overwrite). That means this merge request fixes a now orphaned method `_init_add_element` and has no effect.

It would have been good to have a `0.0.16` bugfix release, before major works. For myself, I have a workaround for the broken `0.0.15`, in case anyone needs:
```python
def workaround_overwrite_labels_issue348(sdata: "SpatialData"):
    # Workaround for https://github.com/scverse/spatialdata/pull/348
    # which causes "UnboundLocalError: local variable 'elem_group' referenced before assignment"
    # when adding elements to backed SpatialData with overwrite=True.
    import zarr
    from ome_zarr.io import parse_url

    if sdata.path is not None:
        store = parse_url(sdata.path, mode="r+").store
        root = zarr.group(store=store)
        if "labels" not in root:
            root.create_group(name="labels")

# To be called before:
workaround_overwrite_labels_issue348(sdata)
sdata.add_labels(name=labels_name_added, labels=labels, overwrite=True)
```

I think the issue can be closed.

LucaMarconato:
Thank you for your comment and apologies for not having merged before. We will now reiterate on incremental IO to enable stable APIs.

PR #367, opened 2023-10-08
fix bug with get_values and labels

LucaMarconato:
This PR is minor and unlocks other PRs, I'll merge it straight away.

### 1. Fix bug with `get_values()` for labels

`get_values()` was failing when the internal function `_filter_table_by_elements()` was called with `match_rows=True`, because the aggregated table from blobs doesn't contain the background in the table (for how we implemented the aggregate function), while the labels has it.

To fix, I simply discard the background index when matching a labels element to the table.

### 2. Minor: fix an assert
Fix an assert that was wrongly written and refactored into the function `_get_element()`

### DIFF
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -93,6 +93,9 @@ def _filter_table_by_elements(
                     xdata = next(iter(v))
                     # can be slow
                     instances = da.unique(xdata.data).compute()
+                if 0 in instances:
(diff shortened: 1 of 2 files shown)


LucaMarconato:
I have modified the way we deal with the background. Now values that are present in an element and not in the table (so a labels element with a background value 0 is one of the cases), do not trigger an error when `_filter_table_by_elements()` is called with `match_rows=True`.

This more general approach is implemented here: https://github.com/scverse/spatialdata/pull/376

PR #369, opened 2023-10-08
Sorting labels indices in get_values()

LucaMarconato:
For labels, get_values() returns a table whose row are sorted by ascending labels value.


LucaMarconato:
The failing tests are due to this: https://github.com/scverse/anndata/issues/1170
Merging because the problem is already in main anyway.

PR #371, opened 2023-10-08
added pre-release test

LucaMarconato:
@ivirshup can you review please?


LucaMarconato:
Let's see if the jobs succeeds, but it seems to be working 😊

ivirshup:
Looks kinda like https://github.com/scverse/spatialdata/pull/370 😉

LucaMarconato:
Nice 😁 Gonna merge yours, but with macOS. In this way if something breaks we can test it locally, and also in this way we test both Python 3.9 and 3.10 with macOS (one in normal mode and one in pre-release mode).

PR #372, opened 2023-10-09
unpin typing extensions

giovp:
spatialdata pins typing extensions for some reasons that I don't remember, but it often conflicts with other packages in the same env. Would like to merge this asap


LucaMarconato:
The change was done here: https://github.com/scverse/spatialdata/pull/279, I think it was to be able to install napari on M1 chips.

I have just tried the installation of every repo, including napari, works on my M1 Mac. So good to go!

LucaMarconato (approved):

Issue #378, opened 2023-10-20
spatialdata.transform MultiscaleSpatialImage raises KeyError


aeisenbarth:
The [implementation of `transform`](https://github.com/scverse/spatialdata/blob/cf1c1f74c9aaa1f99eca3ebc363cc990e381fb7d/src/spatialdata/_core/operations/transform.py#L262C9-L262C10) for `MultiscaleSpatialImage` is not covered by tests. This is a problem because (Multiscale)SpatialImage are structurally and implementationwise significantly different.

For scale1 and greater, the key is supposed to have a dictionary of a transformation matrix from the downscaling operation:
```
            scale = _get_scale(xdata.attrs["transform"])
```
But the key does not exist and I also cannot see it anywhere else in the DataTree.

See this test case in https://github.com/scverse/spatialdata/pull/379:
https://github.com/scverse/spatialdata/blob/7690b49c595b8e996696d5ab4357d0332d42f851/tests/transformations/test_transformations.py#L870-L876

```python
__________ test_keep_numerical_coordinates_c[blobs_multiscale_image] ___________

image_name = 'blobs_multiscale_image'

    @pytest.mark.parametrize("image_name", ["blobs_multiscale_image"])
    def test_keep_numerical_coordinates_c(image_name):
        c_coords = range(3)
        sdata = blobs(n_channels=len(c_coords))
        t = get_transformation(sdata.images[image_name])
>       t_blobs = transform(sdata.images[image_name], t)

test_transformations.py:876: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
…/functools.py:888: in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

data = DataTree('None', parent=None)
├── DataTree('scale0')
│       Dimensions:  (c: 3, y: 512, x: 512)
│       Coordinates:
...0.0
        Data variables:
            image    (c, y, x) float64 dask.array<chunksize=(3, 128, 128), meta=np.ndarray>
transformation = Sequence 
    Scale (y, x)
        [1. 1.]
    Identity 
maintain_positioning = False

    @transform.register(MultiscaleSpatialImage)
    def _(
[...]

PR #379, opened 2023-10-20
Fix channel names in transform for MultiscaleSpatialImage

aeisenbarth:
When transforming a MultiscaleSpatialImage, a KeyError is raised. See https://github.com/scverse/spatialdata/issues/378

So far, I've modified the test case to cover both SpatialImage and MultiscaleSpatialImage and reproduce the problem. I have no proposed fix yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@ and this project adheres to [Semantic Versioning][].
 -   Added `get_element_instances()` (replaces `_get_unique_label_values_as_index()`) #582
 -   Added `get_element_annotators()`, retrieving the tables that annotate a particular SpatialElement #595
 
+### Fixed
+
+-   Preserve channel names of multi-scale images in `transform` (#379)
+
 ## [0.1.2] - 2024-03-30
 
 ### Minor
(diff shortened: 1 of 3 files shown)


giovp:
I think this can be closed right?

aeisenbarth:
I rebased onto main and the KeyError does not occur anymore.

However, the second test case fails because it gets only integer channel names instead of string channel names. It seems the string values get lost by `transform` if it transforms a multi-channel image.

aeisenbarth:
Yes, the changes look good!

melonora (approved):
Thanks it looks good to me. Minor comment, we don't have to do `.images[name_element]` anymore. I usually prefer it as short as possible, but no blocker:)

giovp:
> we don't have to do .images[name_element] anymore

true but I think people still use it (I do) ? also is just for the tests

giovp:
@aeisenbarth could you write a very small note for release? then we can merge, thank you!

melonora:
> > we don't have to do .images[name_element] anymore
> 
> true but I think people still use it (I do) ? also is just for the tests

yeah like I said for me personal preference, but non-blocking:)

melonora:
Thanks Andreas!

PR #383, opened 2023-10-24
contrain spatial_image to avoid xarray typing error

melonora:
Spatial image was updated to a new version, but this currently leads to typing errors, for example in napari-spatialdata. This PR temporarily constrains the spatial-image libraries to avoid this.


giovp (approved):

giovp:
thank you @melonora LGTM!

PR #389, opened 2023-10-28
Fixed bug with pandas+anndata+categories

LucaMarconato:
Fixed a bug that appeared today/a few days ago, probably due to a new release in pandas. The bug is described here: https://github.com/scverse/anndata/issues/1210


LucaMarconato:
I also modified the `ImageTilesDataset` class so that it returns also the geometries used to derive the tiles.

No code review needed, it's very small code.

PR #392, opened 2023-10-30
Added api for renaming coordinate systems

LucaMarconato:
# TLDR;
- `SpatialData.rename_coordinate_systems()`
- No need for review since we want to prioritize bigger PRs
- Tagging you just to make you aware this function exists: @giovp @kevinyamauchi @timtreis @Sonja-Stockhaus @melonora 

# Use case
Simple but convenient API, I wrote so that I can rename the coordinate systems of Visium samples read with `visium()` before concatenating into a unique object, with one coordinate system per sample.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ ci:
     skip: []
 repos:
     - repo: https://github.com/psf/black
-      rev: 23.10.0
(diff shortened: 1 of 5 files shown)


LucaMarconato:
The test fails for unrelated reasons, gonna fix them.

PR #405, opened 2023-11-23
initial tests multi_table design

melonora:
This is a draft PR to allow for discussion regarding the multiple table design for SpatialData. Currently it includes tests with pseudocode. This is work in progress which will be polished.


melonora:
Merging pseudo code. Will be no problem since we are using a feature branch. The branch will be used to work on multiple table implementation during the scverse hackathon.

PR #409, opened 2023-11-28
Fix query of 2D/3D data with 2D/3D bounding box

LucaMarconato:
I improved the 3D support for bounding box queries to cover the following cases:
- points: data can be 2D/3D, bounding box can be 2D/3D
- images and labels: data can be 2D/3D, bounding box can be 2D/3D

I added tests to all the above, and in doing this I refactored the old tests for raster data to use `pytest.mark.parameterize()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ spatialdata-sandbox
 
 # version file
 _version.py
+
+# other
+node_modules/
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -341,16 +341,34 @@ def _(
         raise ValueError(error_message)
 
     if set(axes) != set(output_axes_without_c):
-        if set(axes).issubset(output_axes_without_c):
-            # TODO: hereeeeeeeeeeeeeee need to to something with input_axes_without_c and output_axes_without_c
-            pass
-        else:
-            error_message = (
-                f"Invalid case. The bounding box axes are {axes},"
-                f"the spatial axes in {target_coordinate_system} are"
-                f"{output_axes_without_c}"
-            )
-            raise ValueError(error_message)
+        axes_only_in_bb = set(axes) - set(output_axes_without_c)
+        axes_only_in_output = set(output_axes_without_c) - set(axes)
+
+        # let's remove from the bounding box whose axes that are not in the output axes
+        indices_to_remove_from_bb = [axes.index(ax) for ax in axes_only_in_bb]
+        axes = tuple([ax for ax in axes if ax not in axes_only_in_bb])
+        min_coordinate = np.delete(min_coordinate, indices_to_remove_from_bb)
+        max_coordinate = np.delete(max_coordinate, indices_to_remove_from_bb)
+
+        # if there are axes in the output axes that are not in the bounding box, we need to add them to the bounding box with a huge range
+        for ax in axes_only_in_output:
+            axes = axes + (ax,)
+            M = np.finfo(np.float32).max - 1
+            min_coordinate = np.append(min_coordinate, -M)
+            max_coordinate = np.append(max_coordinate, M)
+
+        #     if case == 1 or case == 5:
+        #         # in these cases the data is 2D or 3D and the bounding box is specified on a different number of axes (1, 2 or 3)
+        #         axes
+        #     else:
+        #         raise NotImplementedError('Cases 2, 3, 4 are not implemented yet.')
+        # else:
+        #     error_message = (
+        #         f"Invalid case. The bounding box axes are {axes},"
+        #         f"the spatial axes in {target_coordinate_system} are"
+        #         f"{output_axes_without_c}"
+        #     )
+        #     raise ValueError(error_message)
 
     spatial_transform = Affine(m_without_c, input_axes=input_axes_without_c, output_axes=output_axes_without_c)
     spatial_transform_bb_axes = Affine(
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -284,7 +284,7 @@ def _get_axes_of_tranformation(
     element: SpatialElement, target_coordinate_system: str
 ) -> tuple[ArrayLike, tuple[str, ...], tuple[str, ...]]:
     """
-    Get the transformation (ignoring c) from the element's intrinsic coordinate system to the query coordinate space.
+    Get the transformation, and transformation's axes (ignoring `c`), from the element's intrinsic coordinate system to the query coordinate space.
 
     Note that the axes which specify the query shape are not necessarily the same as the axes that are output of the
     transformation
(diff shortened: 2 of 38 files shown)


src/spatialdata/_core/query/spatial_query.py
```diff
@@ -269,6 +280,69 @@ def _(
...
+        min_coordinate = np.delete(min_coordinate, indices_to_remove_from_bb)
+        max_coordinate = np.delete(max_coordinate, indices_to_remove_from_bb)
+
+        # if there are axes in the output axes that are not in the bounding box, we need to add them to the bounding box
+        # with a huge range
+        for ax in axes_only_in_output:
+            axes = axes + (ax,)
+            M = np.finfo(np.float32).max - 1
+            min_coordinate = np.append(min_coordinate, -M)
+            max_coordinate = np.append(max_coordinate, M)
```
kevinyamauchi, line 342: Is this the best approach? I'm fine if this is a temporary solution to move us forward, but it feels a bit odd. Maybe just add a comment suggesting what we could do in the future?

kevinyamauchi (commented):
I don't have a ton of time to review in depth right now, but I had a quick look. I think these are good bugs to address, but I don't think I fully understand the solution.

What is happening when there is a 2D bounding box and 3D data?

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: I am working on a follow up PR to fix problems with `polygon_query`, in that context I have found what I think is a better approach. In that cases one has a (2D) polygon to query 3D points. What I suggest is to restrict to the case in which the transformation is invertible and then compute the counter image of the polygon in the intrinsic coordinate system. This polygon will lie on a plane. The better approach says to project the 3D points on this plane and then perform a 2D query. 

Anyway, also in that case I haven't implemented the full approach, rather I check that the z coordinate is not being mixed with x and y coordinates, so that the plane is assured to be orthogonal to the z axis in the intrinsic coordinate system. Here is the code: https://github.com/scverse/spatialdata/blob/aeb2db027cb38494dc01e0f0547336b271a20d35/src/spatialdata/_core/query/spatial_query.py#L122

LucaMarconato: The approach used here with -M, M is a bit more general because it allows for the case in which z and x, y are mixed (but in practice probably this is useless and it just complicates things). Also the bounding_box query for points doesn't require the transformation to be invertible (again, in practice probably this is useless, and it's also slower). 

I propose to add all these things as comments, and then in a future follow up PR address these things and clear the code a bit. But for the moment I think that this fix and the fixes on polygon_queries have higher priority.

LucaMarconato:
> If somebody else has time to review and approves, please feel free to merge without me.

@giovp @melonora would you have some time?

.gitignore
```diff
@@ -45,3 +45,6 @@ spatialdata-sandbox
 
 # version file
 _version.py
+
+# other
+node_modules/
```
giovp, line 50: I saw it as well, what is this for? just out of curisotiy

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -269,6 +280,69 @@ def _(
...
+    """
+    Get the transformation (ignoring c) from the element's intrinsic coordinate system to the query coordinate space.
+
+    Note that the axes which specify the query shape are not necessarily the same as the axes that are output of the
+    transformation
+
+    Parameters
+    ----------
+    element
+    target_coordinate_system
```
giovp, line 295:
```suggestion
    element
    	SpatialData element to be transformed.
    target_coordinate_system
    	The target coordinate system for the transformation.
```

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -269,6 +280,69 @@ def _(
     return SpatialData(**new_elements, table=table)
 
 
+def _get_axes_of_tranformation(
+    element: SpatialElement, target_coordinate_system: str
+) -> tuple[ArrayLike, tuple[str, ...], tuple[str, ...]]:
+    """
+    Get the transformation (ignoring c) from the element's intrinsic coordinate system to the query coordinate space.
```
giovp, line 287:
```suggestion
    Get the transformation, and transformation's axes (ignoring `c`) from the element's intrinsic coordinate system to the query coordinate space.
```

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -269,6 +280,69 @@ def _(
...
+
+
+def _adjust_bounding_box_to_real_axes(
+    axes: tuple[str, ...],
+    min_coordinate: ArrayLike,
+    max_coordinate: ArrayLike,
+    output_axes_without_c: tuple[str, ...],
+) -> tuple[tuple[str, ...], ArrayLike, ArrayLike]:
+    """Adjust the bounding box to the real axes of the transformation."""
+    if set(axes) != set(output_axes_without_c):
```
giovp, line 326: are you sure isn't it better to check with `symmetric_difference`

giovp: although the order would also be the same I imagine, so maybe ok to keep

src/spatialdata/_core/query/spatial_query.py
giovp: > Anyway, also in that case I haven't implemented the full approach, rather I check that the z coordinate is not being mixed with x and y coordinates, so that the plane is assured to be orthogonal to the z axis in the intrinsic coordinate system.
> The approach used here with -M, M is a bit more general because it allows for the case in which z and x, y are mixed (but in practice probably this is useless and it just complicates things).

I indeed also worry that the solution is rather complicated, what does it mean that "z and x,y are mixed" ?

giovp (approved):
looks good @LucaMarconato ! I would really recc to leave a comment on the `_adjust_bounding_box_to_real_axes` because the behaviour is not very intuitive.

.gitignore
melonora: yeah I have in the multiple table PR as well

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: Apologies it was not clear from my message, I meant when z depends on x and y, for instance in a non-invertible shear transformation.

LucaMarconato: The new implementation is in the middle between a general implementation and making some reasonable assumptions. I would keep it like this at this stage to move things forward and then refine in a follow up PR to simplify the code and cut some edge cases.

.gitignore
LucaMarconato: not sure tbh, I think it came from the pre-commits

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: I don't use symmetric_difference here because I need to operate on those two sets differently

PR #410, opened 2023-11-29
Multi table

melonora:
This is the PR adding support for multiple tables in SpatialData. Tables still use `region`, `region_key` and `instance_key` in `table.uns['spatialdata_attrs']`. However, all of these need to be present or none of these are present. This supports the case of tables not annotating an element. Basic functionality is as shown in `test_multiple_table.py`.

Two open issues are what the interface should be for concatenating SpatialData objects. The current implementation assumes only 1 table per SpatialData object to be present with only one name 'table'. This is not the case anymore. I propose two options.
1. enforce that table names should be different and if not throw an error. This would be the same behaviour as currently happens for 'SpatialElements'.    
2. Provide argument to the user 'concatenate_table'. If set to 'True' tables if having the same name will be concatenated.
Personally I would vote for the latter option.

The other issue is regarding the dataloader still assuming 1 table. Question here is whether we still only want to support 1 table for each dataloader instance of whether we want multiple. My proposal would be to have an argument 'table_name' specifying which table should be in the dataloader. Question here is whether there would be a usecase for having more than 1 table in the dataloader.

Matching elements to table and vice versa still needs to be better supported, but given the size of the PR I would reserve this for the next PR. Also reserved for a follow up PR would be the filtering of specific tables if the user wishes to do so.

Note:
The function 'gen_elements' iterates also over tables. In order to avoid confusion with 'gen_elements_values' I have renamed the latter to 'gen_spatial_elements_values' as the latter is often assigned to a variable and does not allow for filtering out tables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning][].
 
 -   Implemented support in SpatialData for storing multiple tables. These tables
     can annotate a SpatialElement but not necessarily so.
+-   Increased in-memory vs on-disk control: changes performed in-memory (e.g. adding a new image) are not automatically performed on-disk.
 
 #### Minor
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
+
 ## [0.1.0] - tbd
 
 ### Added
@@ -24,7 +25,19 @@ and this project adheres to [Semantic Versioning][].
 -   Added public helper function check_target_region_column_symmetry in spatialdata.models to check whether annotation
     metadata in table.uns['spatialdata_attrs'] corresponds with respective columns in table.obs.
 
+## [0.0.x] - tbd
+
+### Minor
+
+-   improved usability and robustness of sdata.write() when overwrite=True @aeisenbarth
+
 ### Added
+
+### Fixed
+
+#### Minor
+
+-   refactored data loader for deep learning
 
 ## [0.0.14] - 2023-10-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,11 @@ and this project adheres to [Semantic Versioning][].
 
 ### Added
 
+-   added SpatialData.subset() API
+
 ### Fixed
+
+-   generalized queries to any combination of 2D/3D data and 2D/3D query region #409
 
 #### Minor
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning][].
 ### Added
 
 -   added SpatialData.subset() API
+-   added SpatialData.locate_element() API
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-
 ## [0.1.0] - tbd
 
 ### Added
@@ -24,6 +23,23 @@ and this project adheres to [Semantic Versioning][].
 -   Added public helper function get_table_keys in spatialdata.models to retrieve annotation information of a given table.
 -   Added public helper function check_target_region_column_symmetry in spatialdata.models to check whether annotation
     metadata in table.uns['spatialdata_attrs'] corresponds with respective columns in table.obs.
+-   Added function validate_table_in_spatialdata in SpatialData to validate the annotation target of a table being
+    present in the SpatialData object.
+-   Added function get_annotated_regions in SpatialData to get the regions annotated by a given table.
+-   Added function get_region_key_column in SpatialData to get the region_key column in table.obs.
+-   Added function get_instance_key_column in SpatialData to get the instance_key column in table.obs.
+-   Added function set_table_annotates_spatialelement in SpatialData to either set or change the annotation metadata of
+    a table in a given SpatialData object.
+-   Added tables property in SpatialData.
+-   Added tables setter in SpatialData.
+-   Added gen_spatial_elements generator in SpatialData to generate the SpatialElements in a given SpatialData object.
+-   Added gen_elements generator in SpatialData to generate elements of a SpatialData object including tables.
+
+### Changed
+
+#### Minor
+
+-   Changed the string representation of SpatialData to reflect the changes in regard to multiple tables.
 
 ## [0.0.x] - tbd
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,20 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-## [0.0.15] - tbd
+## [0.1.0] - tbd
+
+### Added
+
+#### Major
+
+-   Implemented support in SpatialData for storing multiple tables. These tables
+    can annotate a SpatialElement but not necessarily so.
+
+#### Minor
+
+-   Added public helper function get_table_keys in spatialdata.models to retrieve annotation information of a given table.
+-   Added public helper function check_target_region_column_symmetry in spatialdata.models to check whether annotation
+    metadata in table.uns['spatialdata_attrs'] corresponds with respective columns in table.obs.
 
 ### Added
 
--- a/src/spatialdata/_core/_utils.py
+++ b/src/spatialdata/_core/_utils.py
@@ -1,7 +1,4 @@
-from typing import TYPE_CHECKING
-
-if TYPE_CHECKING:
-    from spatialdata import SpatialData
+from spatialdata._core.spatialdata import SpatialData
 
 
 def find_common_keys(sdatas: list[SpatialData]) -> set[str]:
--- a/src/spatialdata/_core/_utils.py
+++ b/src/spatialdata/_core/_utils.py
@@ -1,0 +1,25 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from spatialdata import SpatialData
+
+
+def find_common_keys(sdatas: list[SpatialData]) -> set[str]:
+    """
+    Find table keys present in more than one SpatialData object.
+
+    Parameters
+    ----------
+    sdatas
+        A list of SpatialData objects.
+
+    Returns
+    -------
+    A set of common keys that are present in the tables of more than one SpatialData object.
+    """
+    common_keys = set(sdatas[0].tables.keys())
+
+    for sdata in sdatas[1:]:
+        common_keys.intersection_update(sdata.tables.keys())
+
+    return common_keys
--- a/src/spatialdata/_core/_utils.py
+++ b/src/spatialdata/_core/_utils.py
@@ -1,7 +1,7 @@
 from spatialdata._core.spatialdata import SpatialData
 
 
-def find_common_table_keys(sdatas: list[SpatialData]) -> set[str]:
+def _find_common_table_keys(sdatas: list[SpatialData]) -> set[str]:
     """
     Find table keys present in more than one SpatialData object.
 
--- a/src/spatialdata/_core/_utils.py
+++ b/src/spatialdata/_core/_utils.py
@@ -1,7 +1,7 @@
 from spatialdata._core.spatialdata import SpatialData
 
 
-def find_common_keys(sdatas: list[SpatialData]) -> set[str]:
+def find_common_table_keys(sdatas: list[SpatialData]) -> set[str]:
     """
     Find table keys present in more than one SpatialData object.
 
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+from collections import defaultdict
 from copy import copy  # Should probably go up at the top
 from itertools import chain
 from typing import Any
+from warnings import warn
 
 import numpy as np
 from anndata import AnnData
@@ -115,12 +117,26 @@ def concatenate(
     assert len(sdatas) > 0, "sdatas must be a non-empty list"
 
     if not concatenate_tables:
-        merged_tables = {**{k: v for sdata in sdatas for k, v in sdata.tables.items()}}
-        if len(merged_tables) != np.sum([len(sdata.tables) for sdata in sdatas]):
-            raise KeyError(
-                "Tables must have unique names across the SpatialData objects to concatenate unless"
-                " concatenate_tables is set to True."
+        key_counts: dict[str, int] = defaultdict(int)
+        for sdata in sdatas:
+            for k in sdata.tables:
+                key_counts[k] += 1
+
+        if any(value > 1 for value in key_counts.values()):
+            warn(
+                "Duplicate table names found. Tables will be added with integer suffix. Set concatenate_tables to True"
+                "if concatenation is wished for instead.",
+                UserWarning,
+                stacklevel=2,
             )
+        merged_tables = {}
+        count_dict: dict[str, int] = defaultdict(int)
+
+        for sdata in sdatas:
+            for k, v in sdata.tables.items():
+                new_key = f"{k}_{count_dict[k]}" if key_counts[k] > 1 else k
+                count_dict[k] += 1
+                merged_tables[new_key] = v
     else:
         common_keys = _find_common_table_keys(sdatas)
         merged_tables = {}
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -110,7 +110,6 @@ def concatenate(
     merged_shapes = {**{k: v for sdata in sdatas for k, v in sdata.shapes.items()}}
     if len(merged_shapes) != np.sum([len(sdata.shapes) for sdata in sdatas]):
         raise KeyError("Shapes must have unique names across the SpatialData objects to concatenate")
-    # TODO: finish
 
     assert isinstance(sdatas, list), "sdatas must be a list"
     assert len(sdatas) > 0, "sdatas must be a non-empty list"
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -116,7 +116,7 @@ def concatenate(
 
     if not concatenate_table:
         merged_tables = {**{k: v for sdata in sdatas for k, v in sdata.tables.items()}}
-        if len(merged_shapes) != np.sum([len(sdata.tables) for sdata in sdatas]):
+        if len(merged_tables) != np.sum([len(sdata.tables) for sdata in sdatas]):
             raise KeyError(
                 "Tables must have unique names across the SpatialData objects to concatenate unless"
                 "concatenate_table is set to True."
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -89,7 +89,7 @@ def concatenate(
         If all region_keys are the same, the `region_key` is used.
     instance_key
         The key to use for the instance column in the concatenated object.
-    concatenate_table
+    concatenate_tables
         Whether to merge the tables in case of having the same element name.
     kwargs
         See :func:`anndata.concat` for more details.
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -7,7 +7,7 @@ from typing import Any
 import numpy as np
 from anndata import AnnData
 
-from spatialdata._core._utils import find_common_table_keys
+from spatialdata._core._utils import _find_common_table_keys
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata.models import TableModel
 
@@ -122,7 +122,7 @@ def concatenate(
                 " concatenate_tables is set to True."
             )
     else:
-        common_keys = find_common_table_keys(sdatas)
+        common_keys = _find_common_table_keys(sdatas)
         merged_tables = {}
         for sdata in sdatas:
             for k, v in sdata.tables.items():
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -7,7 +7,7 @@ from typing import Any
 import numpy as np
 from anndata import AnnData
 
-from spatialdata._core._utils import find_common_keys
+from spatialdata._core._utils import find_common_table_keys
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata.models import TableModel
 
@@ -74,7 +74,7 @@ def concatenate(
     sdatas: list[SpatialData],
     region_key: str | None = None,
     instance_key: str | None = None,
-    concatenate_tables: bool = True,
+    concatenate_tables: bool = False,
     **kwargs: Any,
 ) -> SpatialData:
     """
@@ -120,10 +120,10 @@ def concatenate(
         if len(merged_tables) != np.sum([len(sdata.tables) for sdata in sdatas]):
             raise KeyError(
                 "Tables must have unique names across the SpatialData objects to concatenate unless"
-                "concatenate_tables is set to True."
+                " concatenate_tables is set to True."
             )
     else:
-        common_keys = find_common_keys(sdatas)
+        common_keys = find_common_table_keys(sdatas)
         merged_tables = {}
         for sdata in sdatas:
             for k, v in sdata.tables.items():
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -7,6 +7,7 @@ from typing import Any
 import numpy as np
 from anndata import AnnData
 
+from spatialdata._core._utils import find_common_keys
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata.models import TableModel
 
@@ -73,7 +74,7 @@ def concatenate(
     sdatas: list[SpatialData],
     region_key: str | None = None,
     instance_key: str | None = None,
-    concatenate_table: bool = True,
+    concatenate_tables: bool = True,
     **kwargs: Any,
 ) -> SpatialData:
     """
@@ -114,17 +115,22 @@ def concatenate(
     assert isinstance(sdatas, list), "sdatas must be a list"
     assert len(sdatas) > 0, "sdatas must be a non-empty list"
 
-    if not concatenate_table:
+    if not concatenate_tables:
         merged_tables = {**{k: v for sdata in sdatas for k, v in sdata.tables.items()}}
         if len(merged_tables) != np.sum([len(sdata.tables) for sdata in sdatas]):
             raise KeyError(
                 "Tables must have unique names across the SpatialData objects to concatenate unless"
-                "concatenate_table is set to True."
+                "concatenate_tables is set to True."
             )
     else:
-        merged_tables = _concatenate_tables(
-            [sdata.table for sdata in sdatas if sdata.table is not None], region_key, instance_key, **kwargs
-        )
+        common_keys = find_common_keys(sdatas)
+        merged_tables = {}
+        for sdata in sdatas:
+            for k, v in sdata.tables.items():
+                if k in common_keys and merged_tables.get(k) is not None:
+                    merged_tables[k] = _concatenate_tables([merged_tables[k], v], region_key, instance_key, **kwargs)
+                else:
+                    merged_tables[k] = v
 
     return SpatialData(
         images=merged_images,
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -8,6 +8,7 @@ from typing import Any, Callable
 import dask.array as da
 import numpy as np
 import pandas as pd
+from anndata import AnnData
 from dask.dataframe.core import DataFrame as DaskDataFrame
 from datatree import DataTree
 from geopandas import GeoDataFrame
@@ -17,6 +18,7 @@ from spatial_image import SpatialImage
 from tqdm import tqdm
 from xarray import DataArray
 
+from spatialdata._core._elements import Tables
 from spatialdata._core.query._utils import get_bounding_box_corners
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata._logging import logger
@@ -266,7 +268,9 @@ def _(
         new_elements[element_type] = queried_elements
 
     if filter_table:
-        tables = {name: _filter_table_by_elements(table, new_elements) for name, table in sdata.tables.items()}
+        tables: dict[str, AnnData] | Tables = {
+            name: _filter_table_by_elements(table, new_elements) for name, table in sdata.tables.items()
+        }
     else:
         tables = sdata.tables
 
@@ -645,7 +649,9 @@ def _polygon_query(
 
     if filter_table:
         elements = {"shapes": new_shapes, "points": new_points}
-        tables = {name: _filter_table_by_elements(table, elements) for name, table in sdata.tables.items()}
+        tables: dict[str, AnnData] | Tables = {
+            name: _filter_table_by_elements(table, elements) for name, table in sdata.tables.items()
+        }
     else:
         tables = sdata.tables
     return SpatialData(shapes=new_shapes, points=new_points, images=new_images, tables=tables)
@@ -754,7 +760,9 @@ def polygon_query(
         geodataframes[k] = vv
     if filter_table:
         elements = {"shapes": geodataframes}
-        tables = {name: _filter_table_by_elements(table, elements) for name, table in sdata.tables.items()}
+        tables: dict[str, AnnData] | Tables = {
+            name: _filter_table_by_elements(table, elements) for name, table in sdata.tables.items()
+        }
     else:
         tables = sdata.tables
 
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -8,7 +8,6 @@ from typing import Any, Callable
 import dask.array as da
 import numpy as np
 import pandas as pd
-from anndata import AnnData
 from dask.dataframe.core import DataFrame as DaskDataFrame
 from datatree import DataTree
 from geopandas import GeoDataFrame
@@ -18,8 +17,7 @@ from spatial_image import SpatialImage
 from tqdm import tqdm
 from xarray import DataArray
 
-from spatialdata._core._elements import Tables
-from spatialdata._core.query._utils import get_bounding_box_corners
+from spatialdata._core.query._utils import _get_filtered_or_unfiltered_tables, get_bounding_box_corners
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata._logging import logger
 from spatialdata._types import ArrayLike
@@ -250,8 +248,6 @@ def _(
     target_coordinate_system: str,
     filter_table: bool = True,
 ) -> SpatialData:
-    from spatialdata._core.query.relational_query import _filter_table_by_elements
-
     min_coordinate = _parse_list_into_array(min_coordinate)
     max_coordinate = _parse_list_into_array(max_coordinate)
     new_elements = {}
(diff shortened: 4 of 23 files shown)


LucaMarconato:
Thanks @melonora, great work! I am going to start my review. Before submitting it I start with the comment to the PR description. 

### Concatenation

> This is not the case anymore. I propose two options.
> 
> enforce that table names should be different and if not throw an error. This would be the same behaviour as currently happens for 'SpatialElements'.
> Provide argument to the user 'concatenate_table'. If set to 'True' tables if having the same name will be concatenated.
> Personally I would vote for the latter option.

As you suggested, I suggest to have a `concatenate_table` argument. I would set it to `False` by default and go with the option 1, but if the user sets it to `True`, to have the behavior of option 2.

### Dataloaders

> The other issue is regarding the dataloader still assuming 1 table. Question here is whether we still only want to support 1 table for each dataloader instance of whether we want multiple. My proposal would be to have an argument 'table_name' specifying which table should be in the dataloader. Question here is whether there would be a usecase for having more than 1 table in the dataloader.

I agree, we need to extend the dataloaders to this case, I would think about this in a follow up PR.

### Matching tables to elements
> Matching elements to table and vice versa still needs to be better supported, but given the size of the PR I would reserve this for the next PR. Also reserved for a follow up PR would be the filtering of specific tables if the user wishes to do so.

I agree, things will be simplified here. Also we should provide better and more discoverable APIs for the user, and we should show them in the notebooks.

### gen elements
> The function 'gen_elements' iterates also over tables. In order to avoid confusion with 'gen_elements_values' I have renamed the latter to 'gen_spatial_elements_values' as the latter is often assigned to a variable and does not allow for filtering out tables.

Sounds good to me, because they are private function and so they don't strictly need to be backward compatible. We should anyway make them public (or have a single public function with arguments replacing all of them) in a follow up PR.

LucaMarconato (commented):
Hi, I will do this review in pieces. This is some initial feedback on the tests. More to come! So far I can say it's a great work!

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,241 @@
...
+        test_sdata.write(tmpdir)
+        test_sdata = SpatialData.read(tmpdir)
+        assert len(test_sdata.tables) == 2
+
+    def test_single_table_multiple_elements(self, tmp_path: str):
+        tmpdir = Path(tmp_path) / "tmp.zarr"
+        table = _get_table(region=["poly", "multipoly"])
+        subset = table[table.obs.region == "multipoly"]
+        with pytest.raises(ValueError, match="Regions in"):
+            TableModel().validate(subset)
```
LucaMarconato, line 171: This gives the error "ValueError: Regions in the AnnData object and `region` do not match." Which is quite bad in terms of ergonomics, because we want the user to be able to subset a table without having to worry about updating the region element.

I suggest to do the following:
- we write helper functions (static methods of SpatialData or of TableModel) to access "region", "region_key", "instance_key"; in the tutorial we will show that this is what the user should call.
- we never populate adata.uns['spatialdata_attrs']['region'], instead we compute it on the fly when the respective of such methods is called, which simply gives the levels of the categorical column specified by "region_key". If such column is not categorical we compute the unique values and we tell the user that that column should be a categorical.
- when we write, we compute `region` and write it in the Zarr metadata, but we never read it in memory (it's for other tools or for convenient remote access).

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,241 @@
...
+        # # use case example 1
+        # # sorting the shapes visium0 to match the order of the table
+        # sdata["visium0"][sdata.table.obs["__instance_id__"][sdata.table.obs["__spatial_element__"] == "visium0"]]
+        # assert ...
+        # # use case example 2
+        # # subsetting and sorting the table to match the order of the shapes visium0
+        # sub_table = sdata.table[sdata.table.obs["__spatial_element"] == "visium0"]
+        # sub_table.set_index(keys=["__instance_id__"])
+        # sub_table.obs[sdata["visium0"]]
+        # assert ...
```
LucaMarconato, line 192: As commented before, we need to be careful not to remove this comment as the tests that it describes are super important, but we can address them in a follow up PR.

LucaMarconato (commented):
I finished adding my comments on `test_multi_table.py`, going to review the rest in separate incremental reviews.

src/spatialdata/_io/_utils.py
```diff
@@ -282,6 +288,58 @@ def save_transformations(sdata: SpatialData) -> None:
...
+
+def read_table_and_validate(
+    zarr_store_path: str, group: zarr.Group, subgroup: zarr.Group, tables: dict[str, AnnData]
+) -> dict[str, AnnData]:
+    """
+    Read in tables in the tables Zarr.group of a SpatialData Zarr store.
+
+    Parameters
+    ----------
+    zarr_store_path : str
```
LucaMarconato, line 304: types are not needed in the docstring because they are inferred by Sphynx, I will remove them now in a commit.

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,242 @@
...
+            (None, "instance_id", "Instance key instance_id not"),
+        ],
+    )
+    def test_change_annotation_target(self, full_sdata, region_key, instance_key, error_msg):
+        n_obs = full_sdata["table"].n_obs
+        with pytest.raises(ValueError, match="Mismatch found between regions"):
+            full_sdata.set_table_annotation_target("table", "poly")
+
+        del full_sdata["table"].obs["region"]
+        with pytest.raises(ValueError, match="Region key and/or region key"):
```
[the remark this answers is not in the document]
melonora: Refactored with [a708594](https://github.com/scverse/spatialdata/pull/410/commits/a708594b713ddd0650d71f3912f0e2958bc254db)

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,242 @@
...
+            full_sdata.set_table_annotation_target("table", "poly")
+
+        del full_sdata["table"].obs["instance_id"]
+        full_sdata["table"].obs["region"] = ["poly"] * n_obs
+        with pytest.raises(ValueError, match=error_msg):
+            full_sdata.set_table_annotation_target("table", "poly", region_key=region_key, instance_key=instance_key)
```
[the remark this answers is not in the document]
melonora: Same as above

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,242 @@
...
+
+        full_sdata["table"].obs["instance_id"] = range(n_obs)
+        full_sdata.set_table_annotation_target("table", "poly", instance_key="instance_id", region_key=region_key)
+
+        with pytest.raises(ValueError, match="not_existing not present"):
```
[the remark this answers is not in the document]
melonora: same as above

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,242 @@
...
+            (None, None, "Instance key and/or instance"),
+            ("region", None, "Instance key and/or instance"),
+            ("region", "instance_id", "Instance key instance_id not"),
+            (None, "instance_id", "Instance key instance_id not"),
+        ],
+    )
+    def test_change_annotation_target(self, full_sdata, region_key, instance_key, error_msg):
+        n_obs = full_sdata["table"].n_obs
+        with pytest.raises(ValueError, match="Mismatch found between regions"):
+            full_sdata.set_table_annotation_target("table", "poly")
```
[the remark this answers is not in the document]
melonora: This is purely based on a set operation. I wrote it now in a join to have it look more neat, but I think it would be too much to also provide the element type as this is not of particular interest.

tests/io/test_multi_table.py
melonora: Hmm I think this is something to get feedback on perhaps in a community meeting.

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,242 @@
...
+
+        full_sdata["table"].obs["instance_id"] = range(n_obs)
+        full_sdata.set_table_annotation_target("table", "poly", instance_key="instance_id", region_key=region_key)
+
+        with pytest.raises(ValueError, match="not_existing not present"):
+            full_sdata.set_table_annotation_target("table", "circles", region_key="not_existing")
+
+    def test_set_table_nonexisting_target(self, full_sdata):
+        with pytest.raises(ValueError, match="Annotation target"):
+            full_sdata.set_table_annotation_target("table", "non_existing")
```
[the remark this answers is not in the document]
melonora: Refactored.

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def test_set_table_annotation_target(self, full_sdata):
+        del full_sdata["table"].uns[TableModel.ATTRS_KEY]
+        with pytest.raises(TypeError, match="No current annotation"):
```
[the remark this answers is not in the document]
melonora: I disagree here as we are setting the annotation target of an empty table. The argument 'target_element_name' which would populate 'attrs["region"]' is a required argument so if this is not present an error would be thrown earlier.

src/spatialdata/_io/_utils.py
giovp: agree all types should be removed across docstrings

tests/io/test_multi_table.py
giovp: > This gives the error "ValueError: Regions in the AnnData object and region do not match." Which is quite bad in terms of ergonomics, because we want the user to be able to subset a table without having to worry about updating the region element.

I think I disagree, as in it's true it is not ergonomic, but also this table was modified outside of spatialdata and hence it must be validated again before assigning to spatialdata. I think that this type of ergonomics could very well be pushed upstream in anndata. I would keep the error as of now and consider to update the ergonomics in spatialdata

tests/io/test_multi_table.py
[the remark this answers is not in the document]
giovp: > iirc the design discusion, we still use also the region parameter; we should then also add region in the error message.

I agree to this, I think we talked about it in the call @melonora and you agreed affair, let me know if we should address this again.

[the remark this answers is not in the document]
melonora: I reverted this back to using region. The function has been renamed to 'set_table_annotates_spatialelement'. Still not entirely happy with the name but I guess the functionality at least from the function name is clear.

src/spatialdata/_io/_utils.py
melonora: I think all type hints are now removed from doc strings.

melonora:
Perhaps one more question for the behaviour of concatenating `SpatialData` objects, currently if all tables have same `region_key` and `instance_key` the current `region_key` and `instance_key` is used even when a user provides them as arguments. Should this be changed? It might be counter intuitive for the user if they provide the `instance_key` and `region_key`, but other than that I am neutral here.

giovp:
> and instance_key the current region_key and instance_key is used even when a user provides them as arguments. Should this be changed? It might be counter intuitive for the user if they provide the instance_key and region_key, but other than that I am neutral here.

I think not, I would raise an error, basically if the concatenate method can resolve without user input, than there should e no user input, and if the user wishes to do that, it should be done before/after concatenation.

src/spatialdata/_core/_utils.py
```diff
@@ -0,0 +1,22 @@
+from spatialdata._core.spatialdata import SpatialData
+
+
+def find_common_table_keys(sdatas: list[SpatialData]) -> set[str]:
```
giovp, line 4: should this function be private? just wondering, if the function is in a private module, then should it also be appended by `_` or is it redundant?

src/spatialdata/_io/_utils.py
```diff
@@ -282,6 +288,57 @@ def save_transformations(sdata: SpatialData) -> None:
...
+    """
+    count = 0
+    for table_name in subgroup:
+        f_elem = subgroup[table_name]
+        f_elem_store = os.path.join(zarr_store_path, f_elem.path)
+        if isinstance(group.store, zarr.storage.ConsolidatedMetadataStore):
+            tables[table_name] = read_elem(f_elem)
+            # we can replace read_elem with read_anndata_zarr after this PR gets into a release (>= 0.6.5)
+            # https://github.com/scverse/anndata/pull/1057#pullrequestreview-1530623183
+            # table = read_anndata_zarr(f_elem)
```
giovp, line 325: is this comment still valid @LucaMarconato , I am not sure.

giovp (approved):

src/spatialdata/_core/_utils.py
melonora: yeah wasn't certain either, in a way it is redundant, but if someone imports it anyway and then another person would read that code they could think it is public. I am neutral here.

tests/io/test_multi_table.py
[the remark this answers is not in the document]
LucaMarconato: ah ok you are right, `poly` was a elements name not the element type, I read it wrong

src/spatialdata/_core/_utils.py
LucaMarconato: Yeah I'd make it private.

src/spatialdata/_io/_utils.py
LucaMarconato: I think the latest AnnData handles this, gonna open a issue on this and we can do it in a small PR.

LucaMarconato: Opened: https://github.com/scverse/spatialdata/issues/418. Tiny code fix but need to be tested on remote objects, I'd do it separately.

tests/io/test_multi_table.py
LucaMarconato: > hence it must be validated again before assigning to spatialdata. 
Here the problem is triggered by the validation, I would not expect the new object to be invalid just by subsetting. 

I don't think something upstream in AnnData would help here because `region` it's something spatialdata specific.

I think the approach I described is quite ergonomic, we can discuss it maybe at the community meeting today.

tests/io/test_multi_table.py
```diff
@@ -0,0 +1,271 @@
...
+            shapes={f"poly_{i + 1}": test_shapes["poly"], f"multipoly_{i + 1}": test_shapes["multipoly"]},
+            tables={"table": _get_table(region=f"poly_{i + 1}"), "table2": _get_table(region=f"multipoly_{i + 1}")},
+        )
+        for i in range(3)
+    ]
+
+    with pytest.raises(KeyError, match="Tables must have"):
+        concatenate(sdatas)
+
+    merged_sdata = concatenate(sdatas, concatenate_tables=True)
```
LucaMarconato, line 241: Nice! I think here we need to test also the case concatentate_tables=False when the sdata object has > 1 table.

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -348,11 +348,18 @@ def test_no_shared_transformations() -> None:
...
+            if name != element_name:
+                assert test_space not in get_transformation(obj, get_all=True)
+            else:
+                assert test_space in get_transformation(obj, get_all=True)
+
+
+# def test_init_from_elements(full_sdata):
+#     all_elements = {name: el for element_type, name, el in full_sdata._gen_elements() if element_type != "tables"}
+#     sdata = SpatialData.init_from_elements(all_elements, tables=full_sdata.table)
+#     for element_type in ["images", "labels", "points", "shapes", "tables"]:
```
LucaMarconato, line 363: can we remove?

src/spatialdata/_utils.py
```diff
@@ -228,3 +231,68 @@ def _deepcopy_geodataframe(gdf: GeoDataFrame) -> GeoDataFrame:
     new_attrs = deepcopy(gdf.attrs)
     new_gdf.attrs = new_attrs
     return new_gdf
+
+
+# TODO: change to paramspec as soon as we drop support for python 3.9, see https://stackoverflow.com/a/68290080
+def deprecation_alias(**aliases: str) -> Callable[[Callable[..., RT]], Callable[..., RT]]:
```
LucaMarconato, line 237: really cool

src/spatialdata/models/models.py
```diff
@@ -642,24 +643,77 @@ class TableModel:
...
+            The instance_key to be validated and set in table.uns.
+
+        Raises
+        ------
+        ValueError
+            If no instance_key is provided as argument and no instance_key is found in the `uns` attribute of table.
+        ValueError
+            If no instance_key is provided and the instance_key in table.uns does not match any column in table.obs.
+        ValueError
+            If provided instance_key is not present as table.obs column.
```
LucaMarconato, line 669: I think there is a missing case: both instance_key is provided and instance_key is already present in table.uns. I think we should raise an exception to the user.

src/spatialdata/models/models.py
```diff
@@ -642,24 +643,77 @@ class TableModel:
...
+                    "specify instance_key."
+                )
+            if data.obs.get(attrs[self.INSTANCE_KEY]) is None:
+                raise ValueError(
+                    f"Specified instance_key in table.uns '{attrs.get(self.INSTANCE_KEY)}' is not present"
+                    f" as column in table.obs. Please specify instance_key."
+                )
+        if instance_key:
+            if instance_key in data.obs:
+                attrs[self.INSTANCE_KEY] = instance_key
```
LucaMarconato, line 685: Yeah, this replaces the old value in .uns without checking if it was the same

giovp:
just a reminder to close #417 after merging this as it is not needed, since all changes have been implemented here afaik

src/spatialdata/models/models.py
```diff
@@ -678,6 +732,27 @@ def validate(
         if len(set(expected_regions).symmetric_difference(set(found_regions))) > 0:
```
LucaMarconato, line 732: This can be slow, we can make the user aware that this column should be categorical by raising a warning here. If it's categorical we can check the levels, if it's not we can check the uniqueness.

LucaMarconato: Actually the levels is not enought, we should check the counts because there could be a level without a value. Anyway I think with categorical is slightly faster. Or maybe not, wdyt?

src/spatialdata/models/models.py
```diff
@@ -792,3 +869,59 @@ def _validate_and_return(
...
+        The keys in table.uns['spatialdata_attrs']
+    """
+    if table.uns.get(TableModel.ATTRS_KEY):
+        attrs = table.uns[TableModel.ATTRS_KEY]
+        return attrs[TableModel.REGION_KEY], attrs[TableModel.REGION_KEY_KEY], attrs[TableModel.INSTANCE_KEY]
+
+    raise ValueError("No spatialdata_attrs key found in table.uns. Therefore, no table keys found.")
+
+
+def check_target_region_column_symmetry(table: AnnData, region_key: str, target: str | pd.Series) -> None:
```
LucaMarconato, line 899: Useful function. I would add an example in the docstring since I think it would not be immediatly clear to the user when to use it.

tests/io/test_multi_table.py
giovp: how do we know which is "region_key" a priori? we don't know that.

src/spatialdata/models/models.py
giovp: I mean how many regions we expect? maybe order of thousands (although currently not more than dozens)? I would maybe change the line above to 
`data.obs[attr[self.REGION_KEY_KEY]].cat.categories` which would raise error directly if not categorical

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -747,7 +758,12 @@ def polygon_query(
         vv = pd.concat(v)
         vv = vv[~vv.index.duplicated(keep="first")]
         geodataframes[k] = vv
+    if filter_table:
+        elements = {"shapes": geodataframes}
+        tables: dict[str, AnnData] | Tables = {
+            name: _filter_table_by_elements(table, elements) for name, table in sdata.tables.items()
+        }
+    else:
+        tables = sdata.tables
```
LucaMarconato, line 767: this code block appears 3 times, I'd put it into a tiny private function

src/spatialdata/_core/concatenate.py
```diff
@@ -108,16 +114,29 @@ def concatenate(
...
+            )
+    else:
+        common_keys = find_common_table_keys(sdatas)
+        merged_tables = {}
+        for sdata in sdatas:
+            for k, v in sdata.tables.items():
+                if k in common_keys and merged_tables.get(k) is not None:
+                    merged_tables[k] = _concatenate_tables([merged_tables[k], v], region_key, instance_key, **kwargs)
+                else:
+                    merged_tables[k] = v
```
LucaMarconato, line 132: We could reduce the number of _concatenate_tables() being called by calling it once for each unique table name for which there are > 1 tables.

src/spatialdata/models/models.py
LucaMarconato: Not many regions, but the uniqueness is computed on the rows of the table, which can be a large number.

> data.obs[attr[self.REGION_KEY_KEY]].cat.categories which would raise error directly if not categorical

This could work.

But actually I realized that the most low effort change to apply here is to simply change the error into a warning, and the eventually follow up in a different PR.

src/spatialdata/_core/spatialdata.py
```diff
@@ -184,13 +229,167 @@ def from_elements_dict(elements_dict: dict[str, SpatialElement | AnnData]) -> Sp
...
+        table: AnnData,
+        region: str | pd.Series,
+        region_key: str,
+        instance_key: str,
+    ) -> None:
+        """
+        Set the SpatialElement annotation target of an AnnData table.
+
+        This method sets the target element for the table annotation in the provided `table` object based on the
+        specified parameters. It updates the creates the `attrs` dictionary for `table.uns` and only after validation
```
LucaMarconato, line 268: typo

src/spatialdata/models/models.py
```diff
@@ -792,3 +869,59 @@ def _validate_and_return(
...
+    Raises
+    ------
+    ValueError
+        If the regions specified in table.uns["spatialdata_attrs"] are not present in the region column of table.obs.
+    """
+    found_regions = set(table.obs[region_key].unique().tolist())
+    target_element_set = [target] if isinstance(target, str) else target
+    symmetric_difference = found_regions.symmetric_difference(target_element_set)
+    if symmetric_difference:
+        raise ValueError(
```
LucaMarconato, line 924: Also here I think a warning instead of an error would be ok, but I would like to experiment a bit more with the PR before changing this one. So I'd merge it with this like this, and eventually relax.

src/spatialdata/_core/spatialdata.py
```diff
@@ -184,13 +229,167 @@ def from_elements_dict(elements_dict: dict[str, SpatialElement | AnnData]) -> Sp
...
+        if not region_key:
+            if not table_region_key:
+                raise ValueError(
+                    "No region_key in table.uns and no region_key provided as argument. Please specify " "region_key."
+                )
+            if table.obs.get(attrs[TableModel.REGION_KEY_KEY]) is None:
+                raise ValueError(
+                    f"Specified region_key in table.uns '{table_region_key}' is not "
+                    f"present as column in table.obs. Please specify region_key."
+                )
```
LucaMarconato, line 341: Feels a bit confusing that the equivalent of this code for instance_key is in _validate_set_instance_key(). I would create such function also for region_key().

Also _validate_set_instance_key() is called only from here, not from model, so I would move it out of the model and put it in this file.

src/spatialdata/_core/spatialdata.py
```diff
@@ -184,13 +229,167 @@ def from_elements_dict(elements_dict: dict[str, SpatialElement | AnnData]) -> Sp
...
+        Parameters
+        ----------
+        table_name
+            The name of the table to set the annotation target for.
+        region
+            The name of the target element for the annotation. This can either be a string or a pandas Series object.
+        region_key
+            The region key for the annotation. If not specified, defaults to None.
+        instance_key
+            The instance key for the annotation. If not specified, defaults to None.
```
LucaMarconato, line 372: I'd expain that when the table already contains these values they are reused (this is the case in which _change_table_annotation_target() is called)

src/spatialdata/_core/spatialdata.py
```diff
@@ -434,21 +637,27 @@ def filter_by_coordinate_system(self, coordinate_system: str | list[str], filter
...
+                transformations = get_transformation(element, get_all=True)
+                assert isinstance(transformations, dict)
+                for cs in coordinate_system:
+                    if cs in transformations:
+                        if element_type not in elements:
+                            elements[element_type] = {}
+                        elements[element_type][element_name] = element
+                        element_paths_in_coordinate_system.append(element_name)
+
+        # TODO: check whether full table dict should be returned or only those which annotate elements. Also check
```
LucaMarconato, line 650: I'd put a parameter for this, like "also_orphan_tables: bool = False"

src/spatialdata/_core/spatialdata.py
```diff
@@ -434,21 +637,27 @@ def filter_by_coordinate_system(self, coordinate_system: str | list[str], filter
...
+        # filtering with tables having potentially different keys.
```
LucaMarconato, line 651: > Also check filtering with tables having potentially different keys.

I think this is fine, the filtering should be done independently of the table keys.

LucaMarconato (commented):
> just a reminder to close #417 after merging this as it is not needed, since all changes have been implemented here afaik 

I'd maybe to a merge then just to be sure that we are not missing something.

LucaMarconato (commented):
@melonora I have reviewed the PR, really great work, such a needed functionality! I have added (only minor) comments, let's wait for all the conversations to be resolved before the merge. Some additional tasks:

- [ ] please update the docs with the new public functions/methods that have been added
- [ ] please update the changelog.

src/spatialdata/_core/_utils.py
melonora: Resolved with [2576b62](https://github.com/scverse/spatialdata/pull/410/commits/2576b625056978ba94658fbbf7e24b8d84830b1c)

tests/io/test_multi_table.py
melonora: That is done at line 238 no? Or do you mean the case of SpatialData objects without table names being the same?

tests/core/operations/test_spatialdata_operations.py
melonora: Resolved with [dc4bc30](https://github.com/scverse/spatialdata/pull/410/commits/dc4bc30bf5d00a6b6d6450ceebf8561777b86944)

src/spatialdata/models/models.py
melonora: What would be the reason for raising the exception? The validation is a private one and is only used when changing the annotation target. So the intent is in general to overwrite if a value for instance_key is given.

src/spatialdata/models/models.py
melonora: Currently we do raise the warning that the column is changed to categorical so I implemented it with this current behaviour in mind that so far we did not expect a user to make the column categorical until parsed.

I am not certain here whether a warning would be sufficient given that any downstream functionality depending on this would have to check again and raise an error if not symmetric.

My question, should we perhaps enforce the column to be categorical at this stage already?

CHANGELOG.md
```diff
@@ -8,7 +8,20 @@ and this project adheres to [Semantic Versioning][].
...
+-   Implemented support in SpatialData for storing multiple tables. These tables
+    can annotate a SpatialElement but not necessarily so.
+
+#### Minor
+
+-   Added public helper function get_table_keys in spatialdata.models to retrieve annotation information of a given table.
+-   Added public helper function check_target_region_column_symmetry in spatialdata.models to check whether annotation
+    metadata in table.uns['spatialdata_attrs'] corresponds with respective columns in table.obs.
 
 ### Added
```
LucaMarconato, line 26: Please add also the methods of `SpatialData`

src/spatialdata/models/models.py
LucaMarconato: > My question, should we perhaps enforce the column to be categorical at this stage already?

Yes I think that we can enforce the categorical column already, it would simplify things.

> I am not certain here whether a warning would be sufficient given that any downstream functionality depending on this would have to check again and raise an error if not symmetric.

My opinion is that downstream methods should not rely on this design because it's unnecessarily complex and because methods should be flexible anyway. For instance imagine that a table is annotating `sample0` and `sample1`, and the user asks for the plot of `sample1`. Probably internally the table is subset, so it doesn't really matter if the table had `sample0`, `sample1` or even a `sample2` in `region`, because the subset, and the matching of the rows with the geometries (another filtering + reordering) is required anyway. So the information on `region` is not useful. And if the information is really needed one can still check the values of the categorical column. Not all the categories are guaranteed to have rows with that values, but again if this is really really useful, the user can always call `unique()` manually.

src/spatialdata/_core/concatenate.py
melonora: This is essentially what is happening here using `common_keys`. The `common_keys` are the table_names for which in the list of SpatialData objects multiple tables exist.

src/spatialdata/models/models.py
LucaMarconato: I consider the effort to keep those values always in sync unnecessary because most of the time that value will not be used. So better to allow for it to be out-of-sync (and raise a warning), or even better to drop it (follow up PR), and let the user compute it when it's really needed.

tests/io/test_multi_table.py
LucaMarconato: I think it's not done. I mean the case in which there are two spatialdata objects with tables with different names and then calling `concatenate()` with `concatenate_tables=False`.

src/spatialdata/_core/query/spatial_query.py
melonora: Resolved with [0aa05fe](https://github.com/scverse/spatialdata/pull/410/commits/0aa05fe65ea5c4ef03d2dd25d15e00f65686b791)

src/spatialdata/_core/spatialdata.py
melonora: Resolved with [38e2a15](https://github.com/scverse/spatialdata/pull/410/commits/38e2a15514028c82d658da73b1f5668bce8dc07f)

src/spatialdata/_core/concatenate.py
melonora: Marking this as resolved for now as I think the functionality as described is already there.

src/spatialdata/_core/spatialdata.py
melonora: Resolved with [4960b35](https://github.com/scverse/spatialdata/pull/410/commits/4960b3523c1f27d8db607e5295e57117070c50e1).

src/spatialdata/_core/spatialdata.py
melonora: Resolved with [fb2c20c](https://github.com/scverse/spatialdata/pull/410/commits/fb2c20c14ab016aa6ee72959bc4b97fbbf61ec85)

src/spatialdata/models/models.py
melonora: Resolved with [6123f9d](https://github.com/scverse/spatialdata/pull/410/commits/6123f9d03c7ccc839d4a7950d75f46e7626aebff)

src/spatialdata/models/models.py
melonora: Internally the table would be subset, but then would we even validate the table at that stage?

src/spatialdata/models/models.py
```diff
@@ -792,3 +870,73 @@ def _validate_and_return(
...
+        Table annotating specific SpatialElements
+    region_key
+        The column in obs containing for each row which SpatialElement is annotated by that row.
+    target
+         Name of target(s) SpatialElement(s)
+
+    Raises
+    ------
+    ValueError
+        If the regions specified in table.uns["spatialdata_attrs"] are not present in the region column of table.obs.
```
LucaMarconato, line 920: I think this is not correct: the mismatch that is checked here is between values of obs[REGION_KEY] and the target names, not "the regions specified in table.uns['spatialdata_attrs']", or am I missing something?

tests/io/test_multi_table.py
LucaMarconato: Anyway, @giovp and I discussed this in the community meeting, and @melonora and I just discussed this in a call. We agreed on addressing the removal of the region logic from memory (will keeping it on disk) in a follow up PR. Issue to track this here: https://github.com/scverse/spatialdata/issues/419. For the moment using a warning instead of an error is a gentler way to handle this.

src/spatialdata/models/models.py
LucaMarconato: As discussed here: https://github.com/scverse/spatialdata/pull/410#discussion_r1443007852 we are addressing this in a follow up PR https://github.com/scverse/spatialdata/issues/419 and allowing a gentler handling of the mismatch here (warning instead of error).

tests/io/test_multi_table.py
melonora: For the record, @LucaMarconato and I discussed this last week in a brief call. The behaviour of concatenating is getting quite complex as we have no way a priori to know which tables a user would like to concatenate. This would require another argument taking something as a `list` of `list`. Instead we choose for tables having the same name to have an integer suffix in case `concatenate_tables=False` and for those to be concatenate if set to `True`. A user can afterwards still concatenate those tables the user would like to concatenate.

src/spatialdata/_core/spatialdata.py
melonora: resolved with [e6a3893](https://github.com/scverse/spatialdata/pull/410/commits/e6a3893d97ecaf7c12f1facf278400d370890499) although I did keep it in table model in case of in the future being called from elsewhere.

src/spatialdata/models/models.py
melonora: ah yeah it would be  table.uns['spatialdata_attrs']['region'] and obs[REGION_KEY]

tests/io/test_multi_table.py
melonora: resolved with [475c566](https://github.com/scverse/spatialdata/pull/410/commits/475c5666810871bf1340fd29ddf523cd60111d1a)

CHANGELOG.md
melonora: resolved with [bdf0a99](https://github.com/scverse/spatialdata/pull/410/commits/bdf0a996c991d89b630fe42ce8225774b0e700c1)

src/spatialdata/models/models.py
```diff
@@ -643,6 +643,46 @@ class TableModel:
...
 
+    def _validate_set_region_key(self, data: AnnData, region_key: str | None = None) -> None:
+        """
+        Validate the region key in table.uns or set a new region key as the region key column.
+
+        Parameters
+        ----------
+        data
+            The AnnData table.
+        region_key : str, optional
```
LucaMarconato, line 654: please remove type

src/spatialdata/models/models.py
LucaMarconato: Ok makes sense!

LucaMarconato (approved):
Pre-approving! Super important PR, great work!

Let's adjust together the satellite repos/notebooks the next days, looking forward to merge this! 🚀

src/spatialdata/models/models.py
melonora: resolved with [d0422ad](https://github.com/scverse/spatialdata/pull/410/commits/d0422adae7fbd48250cf752470051a59ad055b89)

giovp (approved):
took another look, no outstanding comments/discussions so LGTM!

src/spatialdata/_core/spatialdata.py
```diff
@@ -537,37 +655,93 @@ def filter_by_coordinate_system(self, coordinate_system: str | list[str], filter
...
+                            elements[element_type] = {}
+                        elements[element_type][element_name] = element
+                        element_paths_in_coordinate_system.append(element_name)
+        tables = self._filter_tables(
+            filter_tables, "cs", include_orphan_tables, element_paths=element_paths_in_coordinate_system
+        )
+
+        return SpatialData(**elements, tables=tables)
+
+    def _filter_tables(
```
LucaMarconato, line 693: I would move this function to relational query. That file will need to be refactored to make the functions more organized and set some of them public (we can do it in a follow up PR)

src/spatialdata/_core/spatialdata.py
```diff
@@ -537,37 +655,93 @@ def filter_by_coordinate_system(self, coordinate_system: str | list[str], filter
...
+        ----------
+        filter_tables
+            If True (default), the tables will be filtered to only contain regions
+            of an element belonging to the specified coordinate system(s).
+        by
+            Filter mode. Valid values are "cs" or "elements". Default is None.
+        include_orphan_tables
+            Flag indicating whether to include orphan tables. Default is False.
+        element_paths
+            Check whether this should be changed into element names in coordinate system
```
LucaMarconato, line 714: yes, it's to be changed as discussed thanks

src/spatialdata/_core/spatialdata.py
```diff
@@ -537,37 +655,93 @@ def filter_by_coordinate_system(self, coordinate_system: str | list[str], filter
...
+        elements_dict : Union[Dict[str, Dict[str, Any]], None], optional
```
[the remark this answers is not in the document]
melonora: resolved

src/spatialdata/_core/spatialdata.py
melonora: resolved with refactor to `filter_by_element_names`.

melonora:
Merging as agreed changes have been implemented and checked.

giovp:
amazing work @melonora !

PR #417, opened 2023-12-21
general fixes to adapt table model

giovp:
@melonora these are some additional fixes, I think we need to remove the add_table/store_table etc. due to the merge of #329 and keep the Tables element as close as possible to the rest.

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -177,6 +177,26 @@ class SpatialData:
         self._query = QueryManager(self)
 
     def validate_table_in_spatialdata(self, data: AnnData) -> None:
+        """
+        Validates the presence of the annotation target of a SpatialData table in the SpatialData object.
+
+        This method validates a table in the SpatialData object to ensure that if annotation metadata is present, the
+        annotation target (SpatialElement) is present in the SpatialData object. Otherwise, a warning is raised.
+
+        Parameters
+        ----------
+        data : AnnData
+            The table potentially annotating a SpatialElement
+
+        Returns
+        -------
+        None
+
+        Raises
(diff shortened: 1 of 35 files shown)


melonora:
> @melonora these are some additional fixes, I think we need to remove the add_table/store_table etc. due to the merge of #329 and keep the Tables element as close as possible to the rest.

Agreed, was doing that but keep losing connection in the car:)

melonora:
In the other PR there is currently 1 remaining errors which relates to removing the `store_tables`.
We used to immediately create a zarr group if not there and store the table upon adding so removing this is a backward incompatible change. If we want to keep it this way for this PR, we need to have a check per element whether it is backed or not. What do you think?

LucaMarconato:
@melonora can you please check if this branch can merged/removed?

src/spatialdata/_core/spatialdata.py
```diff
@@ -672,7 +671,7 @@ def filter_by_coordinate_system(
         # TODO: check whether full table dict should be returned or only those which annotate elements. Also check
         # filtering with tables having potentially different keys.
         if filter_tables:
-            tables = {}
+            tables: Tables = Tables(shared_keys=self._shared_keys)
```
melonora, line 674: comparing this to #410 there is a difference here. However, I am not certain that `shared_keys` needs to be passed here since we are returning a new `SpatialData` object. What do you think @giovp @LucaMarconato ?

giovp: yeah I agree, the shared keys should be instantiated in the return line 680. I think this can be closed!

melonora:
I will close this PR as all the changes have already been implemented in #410

PR #420, opened 2024-01-07
Added error message for removed add_elements functions

LucaMarconato:
@giovp I realized that we incremental io PR was not backward compatible. I therefore added some error messages to communicate to the user the new syntax.

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -21,6 +21,7 @@ from spatial_image import SpatialImage
 from spatialdata._core._elements import Images, Labels, Points, Shapes
 from spatialdata._logging import logger
 from spatialdata._types import ArrayLike, Raster_T
+from spatialdata._utils import _error_message_add_element
 from spatialdata.models import (
     Image2DModel,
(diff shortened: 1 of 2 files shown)


src/spatialdata/_core/spatialdata.py
```diff
@@ -820,6 +820,56 @@ def read(file_path: Path | str, selection: tuple[str] | None = None) -> SpatialD
...
+            '\tsdata.images["image_name"] = image\n'
+            '\tsdata.labels["labels_name"] = labels\n'
+            "\t...\n"
+            "\n"
+            "The new syntax does not automatically updates the disk storage, so you need to call sdata.write() when "
+            "the in-memory object is ready to be saved.\n"
+            "To save only a new specific element to an existing Zarr storage please use the functions write_image(), "
+            "write_labels(), write_points(), write_shapes() and write_table(). We are going to make these calls more "
+            "ergonomic in a follow up PR."
+        )
```
giovp, line 837: can this be moved somewhere else (e.g. in utils)?

giovp (approved):
ah that's a good point! completely forgot

PR #422, opened 2024-01-07
Refactor `polygon_query()`

### LINKED ISSUE #255: Refactoring `polygon_query()` and `bounding_box_query()`
TODOs:
- [x] use `sjoin` when querying by multiple polygons at the same time.
*EDIT: I implemented `sjoin` and tested the support for `MuliPolygon`. I have removed the possibility to query by multiple polygons, as the behavior that was implemented was equivalent to querying by a single `Multipolygon`.
- [x] unify APIs interface

LucaMarconato:
Made on top of https://github.com/scverse/spatialdata/pull/409, which should be merged first.

Closes #255.

The PR does the following:
- [x] improved comments
- [x] polygon_query doesn't transform the data but back-transforms the polygon/multipolyon. This is essential for having good performance with multiscale raster data.
- [x] unifies the `polygon_query` and `bounding_box_query` APIs. In particular
    - [x] Uses single dispatch for `polygon_query()`
    - [x] Add `sdata.query.polygon_query`
- [x] makes the functionality of `polygon_query` complete
- [x] unifies and adds more tests, in particular:
	- [x] tests `polygon_query` and `bounding_box_query` within the same test
	- [x] checks that the results are identical
    - [x] tests 3D bounding boxes with polygons
    - [x] tests multipolygons
    - [x] tests different coordinate systems via affine transformations
    - [x] tests Dask dataframes with multiple partitions.
- [x] bounding_box_query for points now doesn't reset indices
- [x] the transformation of queried elements was passed by reference instead of copy, now fixed

Minor:
- [x] `PointsModel.validate()` (which is also called by the parser) now checks that no radius value is <= 0. Added also a test for this.

### DIFF
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -20,8 +20,12 @@ from spatialdata._core.spatialdata import SpatialData
 from spatialdata._types import ArrayLike
 from spatialdata._utils import Number, _parse_list_into_array
 from spatialdata.models import (
+    PointsModel,
+    ShapesModel,
     SpatialElement,
     get_axes_names,
+    points_dask_dataframe_to_geopandas,
+    points_geopandas_to_dask_dataframe,
 )
 from spatialdata.models._utils import ValidAxis_t, get_spatial_axes
 from spatialdata.transformations._utils import compute_coordinates
@@ -99,6 +103,45 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
     return intrinsic_bounding_box_corners, intrinsic_axes
 
 
+def _get_polygons_in_intrinsic_coordinates(
+    element: DaskDataFrame | GeoDataFrame, target_coordinate_system: str, polygons: GeoDataFrame
+) -> GeoDataFrame:
+    from spatialdata._core.operations.transform import transform
+
+    m_without_c, input_axes_without_c, output_axes_without_c = _get_axes_of_tranformation(
+        element, target_coordinate_system
+    )
+    spatial_transform = Affine(m_without_c, input_axes=input_axes_without_c, output_axes=output_axes_without_c)
+
+    m_without_c_linear = m_without_c[:-1, :-1]
+    case = _get_case_of_bounding_box_query(m_without_c_linear, input_axes_without_c, output_axes_without_c)
+    assert case in [1, 5]
+
+    # we need to deal with the equivalent of _adjust_bounding_box_to_real_axes() to account for the fact that the
+    # intrinsic coordinate system of the points could be ('x', 'y', 'z'). The counter image of the polygons in such
+    # coordinate system could be not orthogonal to the 'z' axis; in such a case we would need to project the points to
+    # the plan in which the polygons live. Let's not handle this complexity and simply raise an error if the inverse
+    # transformation is mixing the 'z' axis with the other axes.
+    sorted_input_axes_without_c = ("x", "y", "z")[: len(input_axes_without_c)]
+    spatial_transform_bb_axes = Affine(
+        spatial_transform.to_affine_matrix(input_axes=sorted_input_axes_without_c, output_axes=("x", "y")),
+        input_axes=sorted_input_axes_without_c,
+        output_axes=("x", "y"),
+    )
+    error_message = 'The transformation is mixing the "z" axis with the other axes. This case is not supported.'
+    assert spatial_transform_bb_axes.matrix[2, 0] == 0, error_message
+    assert spatial_transform_bb_axes.matrix[2, 1] == 0, error_message
+
+    # now that we checked the above, we can restrict the transformation to the "x" and "y" axes; this will make it
+    # invertible when the points to query are 3D
+    spatial_transform_bb_axes = spatial_transform_bb_axes.to_affine(input_axes=("x", "y"), output_axes=("x", "y"))
+
+    inverse = spatial_transform_bb_axes.inverse()
+    assert isinstance(inverse, Affine)
+
+    return transform(polygons, inverse)
+
+
 def _get_axes_of_tranformation(
     element: SpatialElement, target_coordinate_system: str
 ) -> tuple[ArrayLike, tuple[str, ...], tuple[str, ...]]:
@@ -160,6 +203,53 @@ def _adjust_bounding_box_to_real_axes(
             min_coordinate = np.append(min_coordinate, -M)
             max_coordinate = np.append(max_coordinate, M)
     return axes, min_coordinate, max_coordinate
+
+
+def _get_case_of_bounding_box_query(
+    m_without_c_linear: ArrayLike,
+    input_axes_without_c: tuple[str, ...],
+    output_axes_without_c: tuple[str, ...],
+) -> int:
+    """Get the case of the current bounding box query.
+
+    See https://github.com/scverse/spatialdata/pull/151 for a detailed overview of the logic of this code,
+    and for the cases the comments refer to.
+    """
+    transform_dimension = np.linalg.matrix_rank(m_without_c_linear)
+    transform_coordinate_length = len(output_axes_without_c)
+    data_dim = len(input_axes_without_c)
+
+    assert data_dim in [2, 3]
+    assert transform_dimension in [2, 3]
+    assert transform_coordinate_length in [2, 3]
+    assert not (data_dim == 2 and transform_dimension == 3)
+    assert not (transform_dimension == 3 and transform_coordinate_length == 2)
+    # see explanation in https://github.com/scverse/spatialdata/pull/151
+    if data_dim == 2 and transform_dimension == 2 and transform_coordinate_length == 2:
+        case = 1
+    elif data_dim == 2 and transform_dimension == 2 and transform_coordinate_length == 3:
+        # currently not handled
+        case = 2
+    elif data_dim == 3 and transform_dimension == 2 and transform_coordinate_length == 2:
+        # currently not handled
+        case = 3
+    elif data_dim == 3 and transform_dimension == 2 and transform_coordinate_length == 3:
+        # currently not handled
+        case = 4
+    elif data_dim == 3 and transform_dimension == 3 and transform_coordinate_length == 3:
+        case = 5
+    else:
+        raise RuntimeError("This should not happen")
+
+    if case in [2, 3, 4]:
+        # to implement case 2: we need to intersect the plane in the extrinsic coordinate system with the 3D bounding
+        # box. The vertices of this polygons needs to be transformed to the intrinsic coordinate system
+        error_message = (
+            f"This case is not supported (data with dimension {data_dim} but transformation with rank "
+            f"{transform_dimension}. Please open a GitHub issue if you want to discuss a use case."
+        )
+        raise ValueError(error_message)
+    return case
 
 
 @dataclass(frozen=True)
@@ -354,13 +444,7 @@ def _(
     max_coordinate: list[Number] | ArrayLike,
     target_coordinate_system: str,
 ) -> SpatialImage | MultiscaleSpatialImage | None:
-    """Implement bounding box query for SpatialImage.
-
-    Notes
-    -----
-    See https://github.com/scverse/spatialdata/pull/151 for a detailed overview of the logic of this code,
-    and for the cases the comments refer to.
-    """
+    """Implement bounding box query for SpatialImage."""
     from spatialdata.transformations import get_transformation, set_transformation
 
     min_coordinate = _parse_list_into_array(min_coordinate)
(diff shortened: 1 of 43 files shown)


tests/conftest.py
```diff
@@ -222,8 +222,7 @@ def _get_shapes() -> dict[str, GeoDataFrame]:
                 MultiPolygon(
                     [
                         Polygon(((0, 0), (0, 1), (1, 10))),
```
LucaMarconato, line 224: this multipolygon was invalid (my_pol.is_valid was False), this created problems with sjoin

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -102,6 +104,165 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+
+    m_without_c_linear = m_without_c[:-1, :-1]
+    case = _get_case_of_bounding_box_query(m_without_c_linear, input_axes_without_c, output_axes_without_c)
+    assert case in [1, 5]
+
+    # we need to deal with the equivalent of _adjust_bounding_box_to_real_axes() to account for the fact that the
+    # intrinsic coordinate system of the points could be ('x', 'y', 'z'). The counter image of the polygon in such
+    # coordinate system could be not orthogonal to the 'z' axis; in such a case we would need to project the points to
+    # the plan in which the polygon live. Let's not handle this complexity and simply raise an error if the inverse
+    # transformation is mixing the 'z' axis with the other axes.
```
giovp, line 127: I'm sorry I don't understand, when would be the case when" the inverse of the transformation is mixing the z axis" ? What does mixing means here?

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -102,6 +104,165 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+    output_axes_without_c
+        The axes of the query coordinate system, without the "c" axis.
+
+    """
+    from spatialdata.transformations import get_transformation
+
+    transform_to_query_space = get_transformation(element, to_coordinate_system=target_coordinate_system)
+    assert isinstance(transform_to_query_space, BaseTransformation)
+    m = _get_affine_for_element(element, transform_to_query_space)
+    input_axes_without_c = tuple([ax for ax in m.input_axes if ax != "c"])
```
giovp, line 181:
```suggestion
    input_axes_without_c = tuple(ax for ax in m.input_axes if ax != "c")
```

giovp: this shouldwork I think ? if yes same below

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -102,6 +104,165 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+    if set(axes) != set(output_axes_without_c):
+        axes_only_in_bb = set(axes) - set(output_axes_without_c)
+        axes_only_in_output = set(output_axes_without_c) - set(axes)
+
+        # let's remove from the bounding box whose axes that are not in the output axes (e.g. querying 2D points with a
+        # 3D bounding box)
+        indices_to_remove_from_bb = [axes.index(ax) for ax in axes_only_in_bb]
+        axes = tuple([ax for ax in axes if ax not in axes_only_in_bb])
+        min_coordinate = np.delete(min_coordinate, indices_to_remove_from_bb)
+        max_coordinate = np.delete(max_coordinate, indices_to_remove_from_bb)
```
giovp, line 207: `min_coordinate` and `max_coordinate` are with respect to axes right? If so

```python
idx_bb = np.in1d(axes, output_axes_without_c)
idx_out = np.in1d(output_axes_without_c, axes)

min_coordinate = min_coordinate[idx_out]
max_coordinate = max_coordinate[idx_out]
```

giovp: if yes, I would change `axes` to `axes_bb` and maybe `output_axes_without_c ` to `axes_out_without_x` to anyway make them consistent

giovp: also, do you need to check (maybe done outside) that min/max coordinate matches in len with some axes (either bb or out)

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -102,6 +104,165 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+
+
+def _get_case_of_bounding_box_query(
+    m_without_c_linear: ArrayLike,
+    input_axes_without_c: tuple[str, ...],
+    output_axes_without_c: tuple[str, ...],
+) -> int:
+    """Get the case of the current bounding box query.
+
+    See https://github.com/scverse/spatialdata/pull/151 for a detailed overview of the logic of this code,
```
giovp, line 226: I think that description was really cool and maybe it's worth to add it as a living document somewhere evnetually

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -619,206 +698,180 @@ def _(
...
-    points
-        Whether to filter points
+        The coordinate system of the polygon/multipolygon.
+    filter_table
+        If `True`, the table is filtered to only contain rows that are annotating regions
+        contained within the query polygon/multipolygon.
+    shapes [Deprecated]
+        This argument is now ignored and will be removed. Please filter the SpatialData object before calling this
+        function.
+    points [Deprecated]
```
giovp, line 739: I don't think this is a good way to specify deprecation, I would just keep the docstrings and call the deprecation warning at functions all

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -102,6 +104,165 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+    transform_dimension = np.linalg.matrix_rank(m_without_c_linear)
+    transform_coordinate_length = len(output_axes_without_c)
+    data_dim = len(input_axes_without_c)
+
+    assert data_dim in [2, 3]
+    assert transform_dimension in [2, 3]
+    assert transform_coordinate_length in [2, 3]
+    assert not (data_dim == 2 and transform_dimension == 3)
+    assert not (transform_dimension == 3 and transform_coordinate_length == 2)
+    # see explanation in https://github.com/scverse/spatialdata/pull/151
```
giovp, line 238: can you add a TL;DR here?

giovp (changes requested):
small things around documentation

src/spatialdata/_core/spatialdata.py
```diff
@@ -1210,24 +1211,8 @@ def bounding_box(
...
-        filter_table
-            If `True`, the table is filtered to only contain rows that are annotating regions
-            contained within the bounding box.
-
-        Returns
-        -------
-        The SpatialData object containing the requested data.
-        Elements with no valid data are omitted.
+        Please see
+        :func:`spatialdata.bounding_box_query` for the complete docstring.
```
kevinyamauchi, line 1215: what is the reason for this change? so you don't have to maintain two docstrings? I think it's fine, but I'm curious

tests/models/test_models.py
```diff
@@ -203,7 +203,7 @@ def test_raster_schema(
     @pytest.mark.parametrize("model", [ShapesModel])
     @pytest.mark.parametrize("path", [POLYGON_PATH, MULTIPOLYGON_PATH, POINT_PATH])
     def test_shapes_model(self, model: ShapesModel, path: Path) -> None:
-        radius = RNG.normal(size=(2,)) if path.name == "points.json" else None
+        radius = np.abs(RNG.normal(size=(2,))) if path.name == "points.json" else None
```
kevinyamauchi, line 206: if we are going to use an RNG, we should set the seed. otherwise, test failures might be probabilistic.

src/spatialdata/_core/query/spatial_query.py
kevinyamauchi: This is also unclear to me. Also, what is "counter image"?

src/spatialdata/_core/query/spatial_query.py
kevinyamauchi: Please link to the specific comment you are referring to. That way people know where to look.

This is a private function, so I am okay with leaving it  less documented. However, I think this would be tough for a maintainer in the future to understand what's going on. Can you please add some comments throughout to explain what is going on?

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -102,6 +104,165 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+            max_coordinate = np.append(max_coordinate, M)
+    return axes, min_coordinate, max_coordinate
+
+
+def _get_case_of_bounding_box_query(
+    m_without_c_linear: ArrayLike,
+    input_axes_without_c: tuple[str, ...],
+    output_axes_without_c: tuple[str, ...],
+) -> int:
+    """Get the case of the current bounding box query.
```
kevinyamauchi, line 224: Can you please clarify what you mean by "case" here? I see below that you intend to link to a larger discussion. However, people need to be able to understand at least at a high level what the function does.

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -585,9 +657,14 @@ def _(
         min_coordinate=min_coordinate,
         max_coordinate=max_coordinate,
     )
-    if bounding_box_mask.sum() == 0:
+    bounding_box_mask = np.where(bounding_box_mask.compute())[0]
```
kevinyamauchi, line 660: nitpick, but these are indices, right? I think a mask is usually a bool.

kevinyamauchi (commented):
Thanks, @LucaMarconato . I think this looks pretty good. I've left some comments below, mostly related to code clarity.

I am okay with the refactors in `test_spatial_query`. I see that they likely reduce the amount of code. However, I think this is at the cost of being able to quickly understand the intention of the test. If that's the trade-off we want, I am fine with it. I would add some comments and docstrings to make it easier to understand though.

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: true, I changed this

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: I will rewrite the comment in a clearer way. Here "counter image" is a bad translation of the Italian "controimmagine". I checked and it's called "preimage" in English.

LucaMarconato: Done

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: true, replaced

LucaMarconato: true, replaced

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: Improved the first line and referring to the comments for a TL;DR

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: I lnked the specific comment, also we can use GitHub as a living document as the core devs can edit the comments of other people.

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: Added.

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: Done.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Exactly, I would like to have it only in one place because it's easy to forget to update the other (happened). Is there any Sphinx trick we can use here? Asking also @giovp. For this PR I'll keep it like this because I'd like to merge some other PRs dependent on this.

LucaMarconato: I see also torch has something similar to what is in this PR: https://pytorch.org/docs/stable/generated/torch.Tensor.sum.html

tests/models/test_models.py
LucaMarconato: true, added

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: > min_coordinate and max_coordinate are with respect to axes right? If so
>
> idx_bb = np.in1d(axes, output_axes_without_c)  
> idx_out = np.in1d(output_axes_without_c, axes)  

this still returns a boolean mask, so it's basically equivalent to what I wrote; I'll keep the old code.

min_coordinate = min_coordinate[idx_out]  
max_coordinate = max_coordinate[idx_out]

LucaMarconato: yes, we check that it matches with bb, and we do it `BoundingBoxRequest.__post_init__()`

Issue #289, opened 2023-06-08
Refactor `get_elements()` from `spatialdata-plot` and push upstream


timtreis:
tbd, will discuss with @LucaMarconato and update

Issue #225, opened 2023-04-13
refactoring query functions


LucaMarconato:
There are multiple ways (some still to be implemented) to subset/query a `SpatialData` object. Some of these methods are living outside this repo. We need to port them here and use a more consistent syntax (for instance always `sdata.query.query_method`).

- [ ] spatial query (bounding box query): `sdata.query.bounding_box()`
- [ ] spatial query (general polygon): not implemented yet
- [ ] query by obs: implemented in `napari-spatialdata`, branch `spatialdata`
- [ ] query by var: not implemented yet
- [ ] query by coordinate system: `sdata.filter_by_coordinate_system()`
- [ ] query by (a list of) element names: implemented in `spatialdata-plot`

FYI @giovp @timtreis

kevinyamauchi:
Currently, the `bounding_box_query` doesn't automatically filter for elements within the target coordinate system. I think we should add that. We could just add a call to `filter_by_coordinate_system` to the beginning of the query. Thoughts?

LucaMarconato:
Also:
- `SpatialData.subset()`

LucaMarconato:
Another point to consider, as observed in https://github.com/scverse/spatialdata/issues/443#issuecomment-1922027061, is that query functions that appear only as methods of `SpatialData` maybe be a bit hidden in the docs. 
- [ ] Modules and functions of modules are better (or we can also put an explicit link to those methods in the docs).

PR #426, opened 2024-01-09
Added subset API; fix behavior with zero-len table

LucaMarconato:
I added an API for subsetting a SpatialData object given a list of elements. This function is similar to `pp.get_elements()` from `spatialdata-plot`, with one of the major differences that it doesn't filter by coordinate system like `pp.get_elements()`. 

Also, I have noticed that when various APIs that use the argument `filter_table=True` where returning a `SpatialData` object with a table with `len` 0, instead of `None`. This was confusing because from the `__repr__` of `SpatialData` it seems that there was no table.

I modified the behavior by setting the table to `None`, but maybe we should allow for tables of zero len and modify the `__repr__` to print such tables. WDYT @giovp @melonora?

---

@timtreis @Sonja-Stockhaus ideally we would have only one function upstream (see https://github.com/scverse/spatialdata/issues/289 and https://github.com/scverse/spatialdata-plot/issues/138). Can you please comment on the first issue by listing what would be needed to have all the functions you need from `spatialdata-plot` upstream? Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning][].
 
 ### Added
 
+-   added SpatialData.subset() API
+
 ### Fixed
 
 #### Minor
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -266,6 +266,8 @@ def _(
         new_elements[element_type] = queried_elements
 
     table = _filter_table_by_elements(sdata.table, new_elements) if filter_table else sdata.table
+    if len(table) == 0:
+        table = None
     return SpatialData(**new_elements, table=table)
 
 
@@ -641,6 +643,8 @@ def _polygon_query(
 
     if filter_table and sdata.table is not None:
         table = _filter_table_by_elements(sdata.table, {"shapes": new_shapes, "points": new_points})
+        if table is not None and len(table) == 0:
+            table = None
     else:
         table = sdata.table
     return SpatialData(shapes=new_shapes, points=new_points, images=new_images, table=table)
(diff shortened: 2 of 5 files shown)


giovp:
that looks great @LucaMarconato ! But what about having `sdata[["image1","label1"]]` instead of `sdata.subset(...)`? or even extending the `query` method and basically having an interface similar to the `pd.query`?

aeisenbarth:
Nice!

I have just written a function (using `get_elements`) with more granularity that additionally filters table names and keys. So maybe it's worth to exchange before adding this new API.

### Developer's use case

I want to provide input to a unit test for a function which adds some sort of data to SpatialData (image, table obs/var name/column). The input is derived as a subset from a fully processed SpatialData with the to-be-added data excluded.

(For example if I want to check whether the function really sets `X` values, I must provide it with a SpatialData that does not have these X values. Also, I filter elements and corresponding regions in the table independently, that means a labels element may be present, but its region has no entries in the table.)

```python
def filter_spatialdata(
    sdata: SpatialData,
    elements: Optional[Iterable[str]] = tuple(),
    regions: Optional[Iterable[str]] = tuple(),
    obs_keys: Optional[Iterable[str]] = tuple(),
    var_keys: Optional[Iterable[str]] = tuple(),
    var_names: Optional[Iterable[str]] = tuple(),
    # I haven't implemented obsm and others.
):
    ...
```

I believe this is too advanced for a public API (I only need it for testing). As an API, it would need to cover all AnnData keys which increases the complexity and potential for bugs, for something most users are not going to use.

### Data analyst's use case

I imagine for a data analyst's use-case, the interest is more in filtering spatial regions but keeping all annotations associated with these regions.

LucaMarconato:
Thanks @aeisenbarth, sounds very useful for development purposes indeed. I suggest to move the conversation on this to the context of this issues: https://github.com/scverse/spatialdata/issues/289 and https://github.com/scverse/spatialdata-plot/issues/138. In fact, when moving out `pp.get_elements()` from `spatialdata-plot`, we could also include your functionality (and you could maybe contribute directly to that PR).

Thanks @giovp for the review. I agree that we could put more order in the query interface and this function could be part of it (together with the one from `spatialdata-plot` and the one from Andreas, after a unification). Nevertheless, I propose to do this in the context of this already open issue https://github.com/scverse/spatialdata/issues/225, so here I prefer to merge the PR to add the new functionality.

PR #437, opened 2024-01-18
fix categorical dtype check

LucaMarconato:
The checks for categorical dtypes in `models.py` had a bug, fixed. Merging right away. CC @giovp


LucaMarconato:
Forgot to update the changelog.

PR #439, opened 2024-01-23
Fixes on dataloaders

LucaMarconato:
This PR improves the `ImageTilesDataset` class in the following ways.
- General minor cleanup, validation of the arguments and improvement of the docstring.
- Now the regions to derive the tiles from can also be labels and multiscale lables; before only polygons, multipolygons and circles were supported. Points are not supported.
- Now the dataset works also with multiscale images, both if `rasterize` is `True` or `False`. The output is always a single scale image.
- The implementation is now simplified thanks to the use of `to_circles()` API; in particular in `_get_tile_coords()` there is no need anymore to perform affine matrix multiplications manually.
- I removed the logic around `polygon.length`. Before if the regions were polygons or multipolygons, the perimeter of these polygon would have been used as the tile width/height. This contrasted with the usage of the diameter for circles. Now any region element is transformed to circles element using `to_circles()` and the tile sizes are the diameters of these circles.

### DIFF
--- a/src/spatialdata/dataloader/datasets.py
+++ b/src/spatialdata/dataloader/datasets.py
@@ -11,6 +11,7 @@ import pandas as pd
 from anndata import AnnData
 from geopandas import GeoDataFrame
 from multiscale_spatial_image import MultiscaleSpatialImage
+from pandas import CategoricalDtype
 from scipy.sparse import issparse
 from torch.utils.data import Dataset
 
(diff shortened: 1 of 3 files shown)


LucaMarconato:
CC @giovp

LucaMarconato:
Further fixes are introduced with https://github.com/scverse/spatialdata/pull/495/files.

PR #444, opened 2024-01-27
Enforce instance key to be dtype int

melonora:
Pretty much as in the title. It also checks whether for each region the values in the `instance_key` column in `obs` are unique.


melonora:
Also one small change with the genes here as single character strings can be interpreted as color.

LucaMarconato:
Looks great to me! Merged 🥳

PR #445, opened 2024-01-28
Join elements table

melonora:
Depends on #444 

This PR adds the possibility of SQL like joins for `SpatialElement`(s) and tables. These functions are private but are accessible through one public function and will replace the current `match_table_to_element`. In case a join would result in an empty dataframe or `AnnData` table, `None` is returned for the specific element. A table can be joined to multiple `SpatialElement` at the time. Joins with of table and images and table to table are not supported through the public function.

Currently, labels are only supported for those joins that do not require filtering of the labels, for example, a `left` join. The joins that are currently supported are `left`, `left_exclusive`, `inner`, `right` and `right_exclusive`.

A further refactor and docstrings will be added once initial review has passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@ and this project adheres to [Semantic Versioning][].
 
 -   Implemented support in SpatialData for storing multiple tables. These tables
     can annotate a SpatialElement but not necessarily so.
+-   Added SQL like joins that can be executed by calling one public function `join_sdata_spatialelement_table`. The following
+    joins are supported: `left`, `left_exclusive`, `right`, `right_exclusive` and `inner`. The function has an option to
+    match rows. For `left` only matching `left` is supported and for `right` join only `right` matching of rows is supported.
+    Not all joins are supported for `Label` elements.
+-   Added function `match_element_to_table` which allows the user to perform a right join of `SpatialElement`(s) with a table
+    with rows matching the row order in the table.
 -   Increased in-memory vs on-disk control: changes performed in-memory (e.g. adding a new image) are not automatically performed on-disk.
 
 #### Minor
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning][].
 -   Added SQL like joins that can be executed by calling one public function `join_sdata_spatialelement_table`. The following
     joins are supported: `left`, `left_exclusive`, `right`, `right_exclusive` and `inner`. The function has an option to
     match rows. For `left` only matching `left` is supported and for `right` join only `right` matching of rows is supported.
-    Not all joins are supported for `Label` elements.
+    Not all joins are supported for `Labels` elements.
 -   Added function `match_element_to_table` which allows the user to perform a right join of `SpatialElement`(s) with a table
     with rows matching the row order in the table.
 -   Increased in-memory vs on-disk control: changes performed in-memory (e.g. adding a new image) are not automatically performed on-disk.
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -208,7 +208,7 @@ def _get_masked_element(
         The spatial element to be masked.
     table_instance_key_column
         The column of a table containing the instance ids
-    match_rows : Literal["left", "right"]
+    match_rows
          Whether to match the indices of the element and table and if so how. If left, element_indices take priority and
         if right table instance ids take priority.
 
@@ -456,8 +456,8 @@ def join_sdata_spatialelement_table(
     joins are symmetric to the `left` joins. In case of an `inner` join of `SpatialElement`(s) and a table, for each an
     element is returned only containing the rows that are present in both the `SpatialElement` and table.
 
-    For `Points` and `Shapes` elements every valid join for argument how is supported. For `Label` elements only `left`
-    joins are fully supported.
+    For `Points` and `Shapes` elements every valid join for argument how is supported. For `Label` elements only the
+    `left` and `right_exclusive` joins are supported.
 
     Parameters
     ----------
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -149,6 +149,81 @@ def _filter_table_by_elements(
     return table
 
 
+def _get_joined_table_indices(
+    joined_indices: pd.Index | None,
+    element_indices: pd.RangeIndex,
+    table_instance_key_column: pd.Series,
+    match_rows: Literal["left", "right"],
+) -> pd.Index:
+    """
+    Get indices of the table that are present in element_indices.
+
+    Parameters
+    ----------
+    joined_indices
+        Current indices that have been found to match indices of an element
+    element_indices
+        Element indices to match against table_instance_key_column.
+    table_instance_key_column
+        The column of a table containing the instance ids.
+    match_rows
+        Whether to match the indices of the element and table and if so how. If left, element_indices take priority and
+        if right table instance ids take priority.
+
+    Returns
+    -------
+        The indices that of the table that match the SpatialElement indices.
+    """
+    mask = table_instance_key_column.isin(element_indices)
+    if joined_indices is None:
+        if match_rows == "left":
+            joined_indices = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
+        else:
+            joined_indices = table_instance_key_column[mask].index
+    else:
+        if match_rows == "left":
+            add_indices = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
+            joined_indices = joined_indices.append(add_indices)
+        # in place append does not work with pd.Index
+        else:
+            joined_indices = joined_indices.append(table_instance_key_column[mask].index)
+    return joined_indices
+
+
+def _get_masked_element(
+    element_indices: pd.RangeIndex,
+    element: SpatialElement,
+    table_instance_key_column: pd.Series,
+    match_rows: Literal["left", "right"],
+) -> SpatialElement:
+    """
+    Get element rows matching the instance ids in the table_instance_key_column.
+
+    Parameters
+    ----------
+    element_indices
+        The indices of an element.
+    element
+        The spatial element to be masked.
+    table_instance_key_column
+        The column of a table containing the instance ids
+    match_rows : Literal["left", "right"]
+         Whether to match the indices of the element and table and if so how. If left, element_indices take priority and
+        if right table instance ids take priority.
+
+    Returns
+    -------
+    The masked spatial element based on the provided indices and match rows.
+    """
+    mask = table_instance_key_column.isin(element_indices)
+    masked_table_instance_key_column = table_instance_key_column[mask]
+    mask_values = mask_values if len(mask_values := masked_table_instance_key_column.values) != 0 else None
+    if match_rows == "right":
+        mask_values = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
+
+    return element.iloc[mask_values, :]
+
+
 def _right_exclusive_join_spatialelement_table(
     element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: str
 ) -> tuple[dict[str, Any], AnnData | None]:
@@ -184,7 +259,7 @@ def _right_exclusive_join_spatialelement_table(
 
 
 def _right_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: str
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: Literal["left", "right"]
 ) -> tuple[dict[str, Any], AnnData]:
     regions, region_column_name, instance_key = get_table_keys(table)
     groups_df = table.obs.groupby(by=region_column_name)
@@ -203,13 +278,7 @@ def _right_join_spatialelement_table(
                     )
                     continue
 
-                mask = table_instance_key_column.isin(element_indices)
-                masked_table_instance_key_column = table_instance_key_column[mask]
-                mask_values = mask_values if len(mask_values := masked_table_instance_key_column.values) != 0 else None
-                if match_rows == "right":
-                    mask_values = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
-
-                masked_element = element.iloc[mask_values, :]
+                masked_element = _get_masked_element(element_indices, element, table_instance_key_column, match_rows)
                 element_dict[element_type][name] = masked_element
             else:
                 warnings.warn(
@@ -220,7 +289,7 @@ def _right_join_spatialelement_table(
 
 
 def _inner_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: str
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: Literal["left", "right"]
 ) -> tuple[dict[str, Any], AnnData]:
     regions, region_column_name, instance_key = get_table_keys(table)
     groups_df = table.obs.groupby(by=region_column_name)
@@ -240,34 +309,19 @@ def _inner_join_spatialelement_table(
                     )
                     continue
 
-                mask = table_instance_key_column.isin(element_indices)
-                masked_table_instance_key_column = table_instance_key_column[mask]
-
-                mask_values = mask_values if len(mask_values := masked_table_instance_key_column.values) != 0 else None
-                if match_rows == "right":
-                    mask_values = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
-
-                masked_element = element.iloc[mask_values, :]
+                masked_element = _get_masked_element(element_indices, element, table_instance_key_column, match_rows)
                 element_dict[element_type][name] = masked_element
 
-                if joined_indices is None:
-                    if match_rows == "left":
-                        joined_indices = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
-                    else:
-                        joined_indices = table_instance_key_column[mask].index
-                else:
-                    if match_rows == "left":
-                        add_indices = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
-                        joined_indices = joined_indices.append(add_indices)
-                    # in place append does not work with pd.Index
-                    else:
-                        joined_indices = joined_indices.append(table_instance_key_column[mask].index)
+                joined_indices = _get_joined_table_indices(
+                    joined_indices, element_indices, table_instance_key_column, match_rows
+                )
             else:
                 warnings.warn(
                     f"The element `{name}` is not annotated by the table. Skipping", UserWarning, stacklevel=2
                 )
                 element_dict[element_type][name] = None
                 continue
+
     joined_table = table[joined_indices, :].copy() if joined_indices is not None else None
     return element_dict, joined_table
 
@@ -304,7 +358,7 @@ def _left_exclusive_join_spatialelement_table(
 
 
 def _left_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: str
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: Literal["left", "right"]
 ) -> tuple[dict[str, Any], AnnData]:
     regions, region_column_name, instance_key = get_table_keys(table)
     groups_df = table.obs.groupby(by=region_column_name)
@@ -319,19 +373,9 @@ def _left_join_spatialelement_table(
                 else:
                     element_indices = _get_unique_label_values_as_index(element)
 
-                mask = table_instance_key_column.isin(element_indices)
-                if joined_indices is None:
-                    if match_rows == "left":
-                        joined_indices = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
-                    else:
-                        joined_indices = table_instance_key_column[mask].index
-                else:
-                    if match_rows == "left":
-                        add_indices = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
-                        joined_indices = joined_indices.append(add_indices)
-                    # in place append does not work with pd.Index
-                    else:
-                        joined_indices = joined_indices.append(table_instance_key_column[mask].index)
+                joined_indices = _get_joined_table_indices(
+                    joined_indices, element_indices, table_instance_key_column, match_rows
+                )
             else:
                 warnings.warn(
                     f"The element `{name}` is not annotated by the table. Skipping", UserWarning, stacklevel=2
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -24,6 +24,7 @@ from spatialdata.models import (
     SpatialElement,
     TableModel,
     get_model,
+    get_table_keys,
 )
 
 
@@ -158,9 +159,7 @@ def _create_element_dict(
 def _right_exclusive_join_spatialelement_table(
     element_dict: dict[str, dict[str, Any]], table: AnnData
 ) -> tuple[dict[str, Any], AnnData | None]:
-    regions = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY]
-    region_column_name = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY_KEY]
-    instance_key = table.uns[TableModel.ATTRS_KEY][TableModel.INSTANCE_KEY]
+    regions, region_column_name, instance_key = get_table_keys(table)
     groups_df = table.obs.groupby(by=region_column_name)
     mask = []
     for element_type, name_element in element_dict.items():
@@ -188,9 +187,7 @@ def _right_exclusive_join_spatialelement_table(
 def _right_join_spatialelement_table(
     element_dict: dict[str, dict[str, Any]], table: AnnData
 ) -> tuple[dict[str, Any], AnnData]:
-    regions = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY]
-    region_column_name = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY_KEY]
-    instance_key = table.uns[TableModel.ATTRS_KEY][TableModel.INSTANCE_KEY]
+    regions, region_column_name, instance_key = get_table_keys(table)
     groups_df = table.obs.groupby(by=region_column_name)
     for element_type, name_element in element_dict.items():
         for name, element in name_element.items():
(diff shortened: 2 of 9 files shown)


src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
     return table
 
 
+def _create_element_dict(
```
LucaMarconato, line 150: I think this function could be replaced by default dicts

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+
+    @classmethod
+    def get(cls, key: str) -> Callable[[dict[str, dict[str, Any]], AnnData], tuple[dict[str, Any], AnnData]] | None:
+        if key in cls.__members__:
+            return cls.__members__.get(key)
+        return None
+
+
+def join_sdata_spatialelement_table(
+    sdata: SpatialData, spatial_element_name: str | list[str], table_name: str, how: str = "LEFT"
```
LucaMarconato, line 359: I'd go for lowercase, as in pandas

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+) -> tuple[dict[str, Any], AnnData]:
```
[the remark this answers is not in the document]
melonora: Yeah agreed, just wanted a first pass to see whether logic was ok before writing docstrings

src/spatialdata/_core/query/relational_query.py
melonora: Ok will check if mypy starts complaining because of lowercase. I think I remember that from napari but not sure

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+    else:
+        raise ValueError(f"`{how}` is not a valid type of join.")
+
+    elements_dict = {
+        name: element for outer_key, dict_val in elements_dict.items() for name, element in dict_val.items()
+    }
+    return elements_dict, table
+
+
 def match_table_to_element(sdata: SpatialData, element_name: str) -> AnnData:
```
LucaMarconato, line 399: This function and the tests associated to this function should be removed/adapted to use the join. In particular I think that the private function `_filter_table_by_elements()`, used internally by `match_table_to_element()` can be removed.

tests/core/query/test_relational_query.py
```diff
@@ -23,6 +23,110 @@ def test_match_table_to_element(sdata_query_aggregation):
...
+        sdata_query_aggregation, "values_polygons", "table", "right_exclusive"
+    )
+    assert table is None
+    assert all(element_dict[key] is None for key in element_dict)
+
+    sdata_query_aggregation["values_polygons"] = sdata_query_aggregation["values_polygons"].drop([10, 11])
+    with pytest.raises(AssertionError, match="No table with"):
+        join_sdata_spatialelement_table(sdata_query_aggregation, "values_polygons", "not_existing_table", "left")
+
+    # Should we reindex before returning the table?
```
LucaMarconato, line 37: Very important point. I think that we should never reindex the geometries (and it doesn't matter if we reindex the table because we never use `table.obs.index` anyway as we use the `region_key` and `instance_key` columns).

Instead, we should definitely match the order of the returned data. I propose an argument: `match_rows_order: bool`. (before I though of making it a string with values `left`, `right`, but thinking about it `bool` is better).

The behavior would be the following: 
- if `False` the order of the returned objects don't change. 
- If `True` and we have a inner join, then a warning is raised that the argument is set to `False`, and the user needs to call, after the inner join, a left or right join with `match_rows_order=True`.
- If `True` and we have a left/right join, then the order is inherited from the left/right.
- If `True` and we have a left/right exclusive join, then we set it to `False` and raise a warning saying that since we have an exclusive join there is no order to match.

LucaMarconato: Actually, not sure if the above is ideal. The most common use case for the user would be a "inner" join, followed by a "right" or "left" join with `match_table_order=True`. So either we allow `match_rows_order` take value `'left'`, `'right'`,`'no'`, either we make the function `match_table_to_element` do exactly this.

tests/core/query/test_relational_query.py
```diff
@@ -23,6 +23,110 @@ def test_match_table_to_element(sdata_query_aggregation):
...
+        [element_dict["values_circles"].index.to_series(), element_dict["values_polygons"].index.to_series()]
+    )
+    assert all(table.obs["instance_id"] == indices.values)
+
+    with pytest.warns(UserWarning, match="The element"):
+        element_dict, table = join_sdata_spatialelement_table(
+            sdata_query_aggregation, ["values_circles", "values_polygons", "by_polygons"], "table", "right_exclusive"
+        )
+    assert all(element_dict[key] is None for key in element_dict)
+    assert all(table.obs.index == ["7", "8", "19", "20"])
```
LucaMarconato, line 68: it's equivalent, but here maybe I'd check that we have region_key, instance_key equal to
```
values_circles, 7  
values_circles, 8  
values_polygons, 10  
values_polygons, 11   
```

tests/core/query/test_relational_query.py
```diff
@@ -23,6 +23,110 @@ def test_match_table_to_element(sdata_query_aggregation):
...
+            sdata_query_aggregation, ["by_polygons", "values_polygons"], "table", "left"
+        )
+    assert "by_polygons" in element_dict
+
+    # check multiple elements joined to table.
+    sdata_query_aggregation["values_circles"] = sdata_query_aggregation["values_circles"].drop([7, 8])
+    element_dict, table = join_sdata_spatialelement_table(
+        sdata_query_aggregation, ["values_circles", "values_polygons"], "table", "left"
+    )
+    indices = pd.concat(
```
LucaMarconato, line 58: Good here. But as implied above it is super crucial that the user knows that this is true only if the initial order between regions and table rows matched, or if the join API is called with `match_rows_order=True`.

tests/core/query/test_relational_query.py
```diff
@@ -23,6 +23,110 @@ def test_match_table_to_element(sdata_query_aggregation):
...
+    assert table is sdata_query_aggregation["table"]
+    assert not {7, 8}.issubset(element_dict["values_circles"].index)
+    assert not {10, 11}.issubset(element_dict["values_polygons"].index)
+
+    element_dict, table = join_sdata_spatialelement_table(
+        sdata_query_aggregation, ["values_circles", "values_polygons"], "table", "left_exclusive"
+    )
+    assert table is None
+    assert not np.array_equal(
+        sdata_query_aggregation["table"].obs.iloc[7:9, 1].values, element_dict["values_circles"].index.values
```
LucaMarconato, line 123: the 1 in the iloc is a bit obscure, I'd replace with `.iloc[7:9]['instance_id'`. Same below.

tests/core/query/test_relational_query.py
```diff
@@ -23,6 +23,110 @@ def test_match_table_to_element(sdata_query_aggregation):
...
+        )
+    assert table is sdata_query_aggregation["table"]
+    assert not {7, 8}.issubset(element_dict["values_circles"].index)
+    assert not {10, 11}.issubset(element_dict["values_polygons"].index)
+
+    element_dict, table = join_sdata_spatialelement_table(
+        sdata_query_aggregation, ["values_circles", "values_polygons"], "table", "left_exclusive"
+    )
+    assert table is None
+    assert not np.array_equal(
```
LucaMarconato, line 122: I find this test confusing. We are not interest in the LHS here, just on the value of the RHS, and that it contains 2 rows, with indices 7 and 8.

tests/core/query/test_relational_query.py
```diff
@@ -23,6 +23,110 @@ def test_match_table_to_element(sdata_query_aggregation):
...
+        sdata_query_aggregation["table"].obs.iloc[7:9, 1].values, element_dict["values_circles"].index.values
+    )
+    assert not np.array_equal(
```
LucaMarconato, line 125: I also would replace this test. Here the LHS is by construction an empty series, so it will be almost always different from the RHS. I would instead verify that the RHS has the indices that we removed `array([10, 11])`.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+        return self.value(*args)
+
+    @classmethod
+    def get(cls, key: str) -> Callable[[dict[str, dict[str, Any]], AnnData], tuple[dict[str, Any], AnnData]] | None:
+        if key in cls.__members__:
+            return cls.__members__.get(key)
+        return None
+
+
+def join_sdata_spatialelement_table(
```
LucaMarconato, line 358: mentioned this also in another comment, but writing it also here: I would add an argument `match_rows_order: bool` with maybe default value `False`. Discussion in the other comment.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+    sdata: SpatialData, spatial_element_name: str | list[str], table_name: str, how: str = "LEFT"
+) -> tuple[dict[str, Any], AnnData]:
+    assert sdata.tables.get(table_name), f"No table with `{table_name}` exists in the SpatialData object."
+    table = sdata.tables[table_name]
+    if isinstance(spatial_element_name, str):
+        spatial_element_name = [spatial_element_name]
+
+    elements_dict: dict[str, dict[str, Any]] = {}
+    for name in spatial_element_name:
+        element_type, _, element = sdata._find_element(name)
```
LucaMarconato, line 368: I'd replace with the public function `sdata[name]`

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+    assert any(key in elements_dict for key in ["labels", "shapes", "points"]), (
+        "No valid element to join in spatial_element_name. Must provide at least one of either `labels`, `points` or "
+        "`shapes`."
+    )
+    if JoinTypes.get(how) is not None:
+        elements_dict, table = JoinTypes[how](elements_dict, table)
+    else:
+        raise ValueError(f"`{how}` is not a valid type of join.")
+
+    elements_dict = {
```
LucaMarconato, line 393: maybe I'd give a different name since the type is different

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+    groups_df = table.obs.groupby(by=region_column_name)
+    mask = []
+    for element_type, name_element in element_dict.items():
+        for name, element in name_element.items():
+            if name in regions:
+                group_df = groups_df.get_group(name)
+                table_instance_key_column = group_df[instance_key]
+                if element_type in ["points", "shapes"]:
+                    element_indices = element.index
+                    element_dict[element_type][name] = None
```
LucaMarconato, line 172: why don't you set the element to `None` for labels? I'd move this line right after the `for name, element in ...`

src/spatialdata/_core/query/relational_query.py
```diff
@@ -129,6 +147,255 @@ def _filter_table_by_elements(
...
+                if element_type in ["points", "shapes"]:
+                    element_indices = element.index
+                else:
+                    element_indices = _get_unique_label_values_as_index(element)
+
+                mask = table_instance_key_column.isin(element_indices)
+                if joined_indices is None:
+                    joined_indices = table_instance_key_column[mask].index
+                else:
+                    # in place append does not work with pd.Index
```
LucaMarconato, line 326: This doesn't preserve the order of the rows of the table but makes the table adopt the ordering given by the elements. This could be the behavior given when `match_rows_order=True`, but otherwise I think we should return the filtered table without reordering.

tests/core/query/test_relational_query.py
LucaMarconato: Iterating more on this. Actually after reviewing the rest of the PR I would actually have `match_rows_order: bool`, with default value `True` and in the case of a inner join return the table sorted to match the elements.

This means:
- with the default value for `match_rows_order`, the exclusive join operations will throw a warning; fine, this will occur rarely because I expect the exclusive joins not to be used often, and the user can always silence it by passing `False`.
- if the user whats to match the order of the shapes to match the table for the inner join, this needs to be done by calling the left join after calling the inner join. A bit tricky, so I would still add teh function `match_table_to_element()`. Note that this operation doesn't make sense when there are multiple elements, so we can restrict `match_table_to_element()` to take a single element instead of a list of them.

LucaMarconato:
Hi, I finished reviewing the PR, super useful functions! 

I have left comments and the most prominent one is regarding the `match_rows_order`.

I have also noticed that some other parts should be added, let's sync here/in Zulip, happy to split the tasks.
- [ ] we should add tests for labels
- [ ] we should add tests for points
- [ ] once the behavior around the `match_rows_order` is implemented, this should have tests
- [ ] we should add the `match_element_to_table()` function (now we have `match_table_to_element()`; I'd remove it). The new function has the order inverted because in this way left and right have the same direction as for `join_spatialelement_table()`.

Also actually I'd postpone the unification of the various joins under a single functions in a follow up PR. I'd focus on functions first.

src/spatialdata/_core/query/relational_query.py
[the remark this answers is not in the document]
melonora: Completed docstring with explicit definition of the different joins.

src/spatialdata/_core/query/relational_query.py
melonora: Should we deprecate this function?

src/spatialdata/_core/query/relational_query.py
melonora: default argument is now lowercase and also all joins are described as lowercase in the  docstring. Kept the enum with capital option though.

melonora: changed enum to lowercase as well.

tests/core/query/test_relational_query.py
melonora: hmm would we really have to throw a warning in case of an exclusive join for something that is by default defined? We can overwrite in the exclusive join function as standard.

tests/core/query/test_relational_query.py
melonora: fixed

tests/core/query/test_relational_query.py
melonora: but because the table is annotating multiple elements index of `10` and `11` is not actually `10` and `11` in the original table. Instead it is `19` and `20`. It also tests this way that the original table in the `SpatialData` object stays the same.

src/spatialdata/_core/query/relational_query.py
melonora: Would cause duplication of throwing key error that is already present in `_find_element`. Unless we create `__get__`.

tests/core/query/test_relational_query.py
```diff
@@ -3,7 +3,7 @@
 import pytest
 from anndata import AnnData
 from spatialdata import get_values, match_table_to_element
-from spatialdata._core.query.relational_query import _locate_value, _ValueOrigin
+from spatialdata._core.query.relational_query import _locate_value, _ValueOrigin, join_sdata_spatialelement_table
 from spatialdata.models.models import TableModel
```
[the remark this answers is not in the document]
melonora: should we remove immediately or first deprecation warning?

src/spatialdata/_core/query/relational_query.py
melonora: if creating a get method in `SpatialData` we could clean the code even more and not having to use default dicts.

tests/core/query/test_relational_query.py
melonora: added

src/spatialdata/_core/query/relational_query.py
melonora: or allow for `"..." in sdata`

tests/core/query/test_relational_query.py
melonora: but it is a left exclusive join so the right hand side should be None and the left hand side should contain the values.

tests/core/query/test_relational_query.py
melonora: just for the record we are not overwriting elements in the `Spatialdata` object when doing joins.

src/spatialdata/_core/query/relational_query.py
melonora: changed to `TypeError`

src/spatialdata/_core/query/relational_query.py
LucaMarconato: true, please feel free to go ahead with that implementation in a tiny PR.

src/spatialdata/_core/query/relational_query.py
LucaMarconato: Yes. I'd add a warning and call `match_element_to_table()` within it.

tests/core/query/test_relational_query.py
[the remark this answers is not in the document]
LucaMarconato: I'd remove the test immediately; I'd put a deprecation warning on `match_table_to_element()` (which will call `match_element_to_table()`, which in turn will use the new join APIs).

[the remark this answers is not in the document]
LucaMarconato: I would also remove right away `_filter_table_by_elements()`. Please in doing it have a final double check that all that that function was doing is now supported by the new join APIs. If not, we should address that.

tests/core/query/test_relational_query.py
LucaMarconato: You are right.

tests/core/query/test_relational_query.py
LucaMarconato: Same as for above, you are correct.

src/spatialdata/_core/query/relational_query.py
LucaMarconato: True, I'd go with the `get`, or momentarily use again `_find_element()`.

tests/core/query/test_relational_query.py
LucaMarconato: Otherwise we can use `match_rows_order: Literal['left', 'right', 'no']` with default value 'no' or 'left'. 

The function `match_element_to_table()` will have the same argument with the same default value.

src/spatialdata/_core/query/relational_query.py
melonora: In sdata is not allowed but did not use it here as we still require the element_type. I used defaultdict now as suggested.

src/spatialdata/_core/query/relational_query.py
melonora: the private function `_filter_table_by_elements` is still used in other parts of the code. Checking whether it can be removed would require more effort beyond the scope of this PR. `match_table_to_element` can use the join internally though.

tests/core/query/test_relational_query.py
melonora: implemented

src/spatialdata/_core/query/relational_query.py
melonora: on second thought, can't do that as it is useful in the join functions to have the `element_type` because the behaviour is different for different elements.

src/spatialdata/_core/query/relational_query.py
melonora: made it a literal as bool does not work for inner join.

src/spatialdata/_core/query/relational_query.py
melonora: changed

src/spatialdata/_core/query/relational_query.py
melonora: it does, see [e7f36f2](https://github.com/scverse/spatialdata/pull/445/commits/e7f36f2fde1f34994c93a3861d95ebfe54960e5d)

tests/core/query/test_relational_query.py
melonora: will clarify in docstring

melonora:
tests for points and labels will be added after merging. The `match_table_to_element` is kept as a friendly function for biologists who may not now about SQL like joins. Also added `match_element_to_table` which is not entirely symmetric but does the equivalent the other way around.

tests/core/query/test_relational_query.py
melonora: clarified but not with `match_rows` as bool, rather literal

tests/core/query/test_relational_query.py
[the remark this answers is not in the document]
melonora: This can't be removed straight away as it is used current status of queries like bounding box queries where we join before adding new elements to sdata.

melonora:
Because of the complicated `aggregate_shapes_by_shapes` test and given that we use private functions in `match_table_to_element` we postpone using the new join here in order to move things ahead. These functions should be refactored at a later stage

tests/core/query/test_relational_query.py
```diff
@@ -132,6 +126,40 @@ def test_left_exclusive_and_right_join(sdata_query_aggregation):
     )
 
 
+def test_match_rows_join(sdata_query_aggregation):
```
LucaMarconato, line 129: Very very good! I see the test for `match_rows='no'` is not included. Do you think it can be skipped or should it be also here?

melonora: I think it can be skipped to be honest

src/spatialdata/_core/query/relational_query.py
```diff
@@ -448,19 +539,53 @@ def match_table_to_element(sdata: SpatialData, element_name: str) -> AnnData:
...
         SpatialData object
     element_name
-        Name of the element to match the table to
+        The name of the spatial elements to be joined with the table.
 
     Returns
     -------
     Table with the rows matching the instances of the element
     """
+    # TODO: refactor this to make use of the new join_sdata_spatialelement_table function.
```
LucaMarconato, line 548: Probably the easiest will be to replace the existing calls of `match_table_to_element()` with `join_sdata_spatialelement_table()`.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -304,7 +360,7 @@ def _left_exclusive_join_spatialelement_table(
 
 
 def _left_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: Literal["left", "right"]
```
LucaMarconato, line 363: Here it should be also `no`

src/spatialdata/_core/query/relational_query.py
```diff
@@ -247,33 +311,25 @@ def _inner_join_spatialelement_table(
...
                 element_dict[element_type][name] = None
                 continue
+
     joined_table = table[joined_indices, :].copy() if joined_indices is not None else None
     return element_dict, joined_table
 
 
 def _left_exclusive_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: str
```
LucaMarconato, line 332: why str and not `Literal['left' ,'no']`?

src/spatialdata/_core/query/relational_query.py
```diff
@@ -227,7 +291,7 @@ def _right_join_spatialelement_table(
 
 
 def _inner_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: Literal["left", "right"]
```
LucaMarconato, line 294: Here `no` should be supported

src/spatialdata/_core/query/relational_query.py
```diff
@@ -191,7 +261,7 @@ def _right_exclusive_join_spatialelement_table(
 
 
 def _right_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: Literal["left", "right"]
```
LucaMarconato, line 264: It should also support `no`

src/spatialdata/_core/query/relational_query.py
```diff
@@ -147,17 +150,83 @@ def _filter_table_by_elements(
...
+    mask_values = mask_values if len(mask_values := masked_table_instance_key_column.values) != 0 else None
+    if match_rows == "right":
+        mask_values = _match_rows(table_instance_key_column, mask, element_indices, match_rows)
+
+    return element.loc[mask_values, :]
 
 
 def _right_exclusive_join_spatialelement_table(
-    element_dict: dict[str, dict[str, Any]], table: AnnData
+    element_dict: dict[str, dict[str, Any]], table: AnnData, match_rows: str
```
LucaMarconato, line 229: I'd replace `str`, for consistency, with `why str and not `Literal['no' ,'right']`

LucaMarconato: Also a check of the valid directions could be added.

LucaMarconato:
@melonora really good work, it will be super useful! I gave my review to the new changes. It's only minor things.

Furthermore, I noticed that the changelog and `api.md` are not updated, please don't forget to do it when new features are added.

Please have a look at the above comments, after that I think we are good to merge.

src/spatialdata/_core/query/relational_query.py
melonora: resolved

src/spatialdata/_core/query/relational_query.py
melonora: resolved

src/spatialdata/_core/query/relational_query.py
melonora: resolved

src/spatialdata/_core/query/relational_query.py
melonora: will add the check later. You are right

CHANGELOG.md
```diff
@@ -16,6 +16,12 @@ and this project adheres to [Semantic Versioning][].
 
 -   Implemented support in SpatialData for storing multiple tables. These tables
     can annotate a SpatialElement but not necessarily so.
+-   Added SQL like joins that can be executed by calling one public function `join_sdata_spatialelement_table`. The following
+    joins are supported: `left`, `left_exclusive`, `right`, `right_exclusive` and `inner`. The function has an option to
+    match rows. For `left` only matching `left` is supported and for `right` join only `right` matching of rows is supported.
+    Not all joins are supported for `Label` elements.
```
LucaMarconato, line 22: minor thing, I changed this to `Labels`

PR #449, opened 2024-02-01
Added `transform_to_data_extent()`, improved `transform()`

LucaMarconato:
# New API:
- Added utils function `transform_to_data_extent()`, useful to create a converted with downscaled images, where all the -transformations are applied and where 1 pixel = 1 physical unit.

# Improved robustness and tests of transformations
- Added `maintain_positioning` argument to`SpatialData.transform_to_coordinate_system()` and `SpatialData.transform_element_to_coordinate_system()`.
- Fixed bug: incorrect offset translation vector for raster data when calling the two above functions, `transform()` was not affected)
- Fixed bug: `transform()` on multiscale raster was returning the offset translation for the most low res scale, not scale0

Most importantly:
- `transform()` now requires the argument `to_coordinate_system` instead of `transformation` (unless `maintain_positioning` is `True`); in fact, non-explicit target coordinate systems may lead to unintuitive results when an element is aligned to multiple coordinate systems. This change is communicated to the user by an informative exception; in some cases backward compatibility could be ensured so only an informative warning is given.

Note:
The above bugs were not affecting the rest of the codebase because `query()`, `aggregate()`, ... make either use of `transform()` for non-raster data, either operate on the transformations directly, without transforming raster data.

### DIFF
--- a/src/spatialdata/_core/data_extent.py
+++ b/src/spatialdata/_core/data_extent.py
@@ -360,7 +360,6 @@ def _compute_extent_in_coordinate_system(
     )
     df = pd.DataFrame(corners.data, columns=corners.axis.data.tolist())
     d = get_transformation(element, get_all=True)
-    assert isinstance(d, dict)
     points = PointsModel.parse(df, coordinates={k: k for k in axes}, transformations=d)
     transformed_corners = pd.DataFrame(transform(points, to_coordinate_system=coordinate_system).compute())
     # Make sure min and max values are in the same order as axes
--- a/src/spatialdata/_core/data_extent.py
+++ b/src/spatialdata/_core/data_extent.py
@@ -289,9 +289,7 @@ def _(e: GeoDataFrame, coordinate_system: str = "global", exact: bool = True) ->
             coordinate_system=coordinate_system,
             extent=extent,
         )
-    t = get_transformation(e, to_coordinate_system=coordinate_system)
-    assert isinstance(t, BaseTransformation)
-    transformed = transform(e, t)
+    transformed = transform(e, to_coordinate_system=coordinate_system)
     return _get_extent_of_shapes(transformed)
 
 
@@ -305,9 +303,7 @@ def _(e: DaskDataFrame, coordinate_system: str = "global", exact: bool = True) -
             coordinate_system=coordinate_system,
             extent=extent,
         )
-    t = get_transformation(e, to_coordinate_system=coordinate_system)
-    assert isinstance(t, BaseTransformation)
-    transformed = transform(e, t)
+    transformed = transform(e, to_coordinate_system=coordinate_system)
     return _get_extent_of_points(transformed)
 
 
--- a/src/spatialdata/_core/data_extent.py
+++ b/src/spatialdata/_core/data_extent.py
@@ -171,7 +171,7 @@ def get_extent(
     The exact extent is the bounding box `[minx, miny, maxx, maxy] = [0, 0, 0, 1.414]`, while the approximate extent is
     the box `[minx, miny, maxx, maxy] = [-0.707, 0, 0.707, 1.414]`.
     """
-    raise ValueError("The object type is not supported.")
+    raise ValueError(f"The object type {type(e)} is not supported.")
 
 
 @get_extent.register
@@ -369,16 +369,16 @@ def _compute_extent_in_coordinate_system(
     return extent
 
 
-def are_extents_equal(sdata0: SpatialData, sdata1: SpatialData, atol: float = 0.1) -> bool:
+def are_extents_equal(extent0: BoundingBoxDescription, extent1: BoundingBoxDescription, atol: float = 0.1) -> bool:
     """
-    Check if the extents of two SpatialData objects are equal.
+    Check if two data extents, as returned by `get_extent()` are equal up to approximation errors.
 
     Parameters
     ----------
-    sdata0
-        The first SpatialData object.
-    sdata1
-        The second SpatialData object.
+    extent0
+        The first data extent.
+    extent1
+        The second data extent.
     atol
         The absolute tolerance to use when comparing the extents.
 
@@ -392,6 +392,4 @@ def are_extents_equal(sdata0: SpatialData, sdata1: SpatialData, atol: float = 0.
     rasterized data slightly different from the extent of the original data. This bug is tracked in
     https://github.com/scverse/spatialdata/issues/165
     """
-    e0 = get_extent(sdata0)
-    e1 = get_extent(sdata1, coordinate_system="transformed_back")
-    return all(np.allclose(e0[k], e1[k], atol=atol) for k in set(e0.keys()).union(e1.keys()))
+    return all(np.allclose(extent0[k], extent1[k], atol=atol) for k in set(extent0.keys()).union(extent1.keys()))
--- a/src/spatialdata/_core/data_extent.py
+++ b/src/spatialdata/_core/data_extent.py
@@ -367,3 +367,31 @@ def _compute_extent_in_coordinate_system(
     for ax in axes:
         extent[ax] = (transformed_corners[ax].min(), transformed_corners[ax].max())
     return extent
+
+
+def are_extents_equal(sdata0: SpatialData, sdata1: SpatialData, atol: float = 0.1) -> bool:
+    """
+    Check if the extents of two SpatialData objects are equal.
+
+    Parameters
+    ----------
+    sdata0
+        The first SpatialData object.
+    sdata1
+        The second SpatialData object.
+    atol
+        The absolute tolerance to use when comparing the extents.
+
+    Returns
+    -------
+    Whether the extents are equal or not.
+
+    Notes
+    -----
+    The default value of `atol` is currently high because of a bug of `rasterize()` that makes the extent of the
+    rasterized data slightly different from the extent of the original data. This bug is tracked in
+    https://github.com/scverse/spatialdata/issues/165
+    """
+    e0 = get_extent(sdata0)
+    e1 = get_extent(sdata1, coordinate_system="transformed_back")
+    return all(np.allclose(e0[k], e1[k], atol=atol) for k in set(e0.keys()).union(e1.keys()))
--- a/src/spatialdata/_core/data_extent.py
+++ b/src/spatialdata/_core/data_extent.py
@@ -20,9 +20,6 @@ from spatialdata.models import get_axes_names
 from spatialdata.models._utils import SpatialElement
 from spatialdata.models.models import PointsModel
 from spatialdata.transformations.operations import get_transformation
-from spatialdata.transformations.transformations import (
-    BaseTransformation,
-)
 
 BoundingBoxDescription = dict[str, tuple[float, float]]
 
@@ -349,8 +346,6 @@ def _compute_extent_in_coordinate_system(
     -------
     The bounding box description in the specified coordinate system.
     """
-    transformation = get_transformation(element, to_coordinate_system=coordinate_system)
-    assert isinstance(transformation, BaseTransformation)
     from spatialdata._core.query._utils import get_bounding_box_corners
 
     axes = get_axes_names(element)
@@ -364,8 +359,10 @@ def _compute_extent_in_coordinate_system(
         max_coordinate=max_coordinates,
     )
     df = pd.DataFrame(corners.data, columns=corners.axis.data.tolist())
-    points = PointsModel.parse(df, coordinates={k: k for k in axes})
-    transformed_corners = pd.DataFrame(transform(points, transformation).compute())
+    d = get_transformation(element, get_all=True)
+    assert isinstance(d, dict)
+    points = PointsModel.parse(df, coordinates={k: k for k in axes}, transformations=d)
+    transformed_corners = pd.DataFrame(transform(points, to_coordinate_system=coordinate_system).compute())
     # Make sure min and max values are in the same order as axes
     extent = {}
     for ax in axes:
--- a/src/spatialdata/_core/operations/_utils.py
+++ b/src/spatialdata/_core/operations/_utils.py
@@ -63,24 +63,27 @@ def transform_to_data_extent(
     sdata_raster = SpatialData(images=dict(sdata.images), labels=dict(sdata.labels))
     sdata_vector_transformed = sdata_vector.transform_to_coordinate_system(coordinate_system)
 
-    de = get_extent(sdata, coordinate_system=coordinate_system)
-    de_axes = tuple(de.keys())
-    translation_to_origin = Translation([-de[ax][0] for ax in de_axes], axes=de_axes)
+    data_extent = get_extent(sdata, coordinate_system=coordinate_system)
+    data_extent_axes = tuple(data_extent.keys())
+    translation_to_origin = Translation([-data_extent[ax][0] for ax in data_extent_axes], axes=data_extent_axes)
 
-    sizes = [de[ax][1] - de[ax][0] for ax in de_axes]
+    sizes = [data_extent[ax][1] - data_extent[ax][0] for ax in data_extent_axes]
     target_width, target_height, target_depth = _compute_target_dimensions(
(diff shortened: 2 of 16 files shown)


src/spatialdata/_core/spatialdata.py
```diff
@@ -490,29 +490,72 @@ def transform_element_to_coordinate_system(
...
         t = get_transformation_between_coordinate_systems(self, element, target_coordinate_system)
-        transformed = transform(element, t, maintain_positioning=False)
-        remove_transformation(transformed, remove_all=True)
-        set_transformation(transformed, Identity(), target_coordinate_system)
-
+        if maintain_positioning:
+            transformed = transform(element, transformation=t, maintain_positioning=maintain_positioning)
+        else:
+            d = get_transformation(element, get_all=True)
+            assert isinstance(d, dict)
```
melonora, line 516: Idem here, assert is not required I believe.
```suggestion
```

src/spatialdata/_core/spatialdata.py
```diff
@@ -490,29 +490,72 @@ def transform_element_to_coordinate_system(
...
+                d[target_coordinate_system] = t
+                to_remove = True
+            transformed = transform(
+                element, to_coordinate_system=target_coordinate_system, maintain_positioning=maintain_positioning
+            )
+            if to_remove:
+                del d[target_coordinate_system]
+        if not maintain_positioning:
+            d = get_transformation(transformed, get_all=True)
+            assert isinstance(d, dict)
```
melonora, line 528:
```suggestion
```

src/spatialdata/_core/spatialdata.py
```diff
@@ -490,29 +490,72 @@ def transform_element_to_coordinate_system(
...
+            t = list(d.values())[0]
+            remove_transformation(transformed, remove_all=True)
+            set_transformation(transformed, t, target_coordinate_system)
+        else:
+            # When maintaining positioning is true, and if the element has a transformation to target_coordinate_system
+            # (this may not be the case because it could be that the element is not directly mapped to that coordinate
+            # system), then the transformation to the target coordinate system is not needed # because the data is now
+            # already transformed; here we remove such transformation.
+            d = get_transformation(transformed, get_all=True)
+            assert isinstance(d, dict)
```
melonora, line 539:
```suggestion
```

src/spatialdata/_core/operations/rasterize.py
```diff
@@ -448,7 +448,8 @@ def _(
     if schema in (Labels2DModel, Labels3DModel):
         kwargs = {"prefilter": False, "order": 0}
     elif schema in (Image2DModel, Image3DModel):
-        kwargs = {}
+        kwargs = {"order": 0}
+        # kwargs = {}
```
melonora, line 452:
```suggestion
```

src/spatialdata/_core/operations/transform.py
```diff
@@ -132,124 +135,192 @@ def _prepend_transformation(
...
         assert raster_translation is None
-        if maintain_positioning:
-            to_prepend = transformation.inverse()
+        to_prepend = transformation.inverse() if maintain_positioning else Identity()
     else:
         raise TypeError(f"Unsupported type {type(element)}")
+    assert isinstance(to_prepend, BaseTransformation)
 
     d = get_transformation(element, get_all=True)
     assert isinstance(d, dict)
```
melonora, line 168:
```suggestion
```

src/spatialdata/_core/operations/transform.py
```diff
@@ -132,124 +135,192 @@ def _prepend_transformation(
...
+    data: SpatialElement,
+    transformation: BaseTransformation | None,
+    maintain_positioning: bool,
+    to_coordinate_system: str | None,
+) -> tuple[BaseTransformation, str | None]:
+    from spatialdata.transformations import BaseTransformation, get_transformation
+
+    if not maintain_positioning:
+        d = get_transformation(data, get_all=True)
+        assert isinstance(d, dict)
```
melonora, line 192:
```suggestion
```

melonora (commented):
Very quick first pass:)

tests/core/operations/test_transform.py
```diff
@@ -117,65 +118,88 @@ def _unpad_rasters(sdata: SpatialData) -> SpatialData:
...
+def _postpone_transformation(
+    sdata: SpatialData, from_coordinate_system: str, to_coordinate_system: str, transformation: BaseTransformation
+):
+    for element in sdata._gen_elements_values():
+        d = get_transformation(element, get_all=True)
+        assert isinstance(d, dict)
+        assert len(d) == 1
+        t = d[from_coordinate_system]
+        sequence = Sequence([t, transformation])
+        set_transformation(element, sequence, to_coordinate_system)
```
melonora, line 130: Perhaps better to move this to a `_test_utils.py` in src/spatialdata.

tests/core/operations/test_transform.py
```diff
@@ -117,65 +118,88 @@ def _unpad_rasters(sdata: SpatialData) -> SpatialData:
...
-    affine = _get_affine(small_translation=False)
-    padded = transform(
-        transform(sdata, affine, maintain_positioning=False), affine.inverse(), maintain_positioning=False
-    )
-    _unpad_rasters(padded)
-    # raise NotImplementedError("TODO: plot the images")
-    # raise NotImplementedError("TODO: compare the transformed images with the original ones")
+def _remove_coordinate_system(sdata: SpatialData, coordinate_system: str):
+    for element in sdata._gen_elements_values():
+        remove_transformation(element, coordinate_system)
```
melonora, line 135: Same as above

tests/core/operations/test_transform.py
```diff
@@ -117,65 +118,88 @@ def _unpad_rasters(sdata: SpatialData) -> SpatialData:
...
-    # TODO: unpad the image
-    # raise NotImplementedError("TODO: compare the transformed images with the original ones")
+def _compare_extents(sdata0: SpatialData, sdata1: SpatialData):
+    e0 = get_extent(sdata0)
+    e1 = get_extent(sdata1, coordinate_system="transformed_back")
 
+    # we are not pixel perfect here, but "0.1-pixel perfect", I think this is due to an equivalent bug than then one
+    # that appears with rasterize (https://github.com/scverse/spatialdata/issues/165), probably originating from
+    # ndinterp(), which is used both by rasterize() and by transform()
+    assert all(np.allclose(e0[k], e1[k], atol=0.1) for k in set(e0.keys()).union(e1.keys()))
```
melonora, line 145: same as above

melonora (approved):
For record, the asserts were due to mypy complaining that a `BaseTransformation` does not have `keys`, while there is already an assert for the case that `get_all` is `True`. The other option other than the assert would be to ignore these mypy errors.

I think the PR can be approved, but I do think for future and with the refactor more comments can be added here and there to explain the code to make it more clear. Right now without the explanation on zoom it could be hard to follow or maybe that is just me:)

src/spatialdata/_core/data_extent.py
```diff
@@ -289,9 +286,7 @@ def _(e: GeoDataFrame, coordinate_system: str = "global", exact: bool = True) ->
             coordinate_system=coordinate_system,
             extent=extent,
         )
-    t = get_transformation(e, to_coordinate_system=coordinate_system)
-    assert isinstance(t, BaseTransformation)
-    transformed = transform(e, t)
+    transformed = transform(e, to_coordinate_system=coordinate_system)
```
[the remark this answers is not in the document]
LucaMarconato: Done.

src/spatialdata/_core/data_extent.py
```diff
@@ -368,8 +359,10 @@ def _compute_extent_in_coordinate_system(
         max_coordinate=max_coordinates,
     )
     df = pd.DataFrame(corners.data, columns=corners.axis.data.tolist())
-    points = PointsModel.parse(df, coordinates={k: k for k in axes})
-    transformed_corners = pd.DataFrame(transform(points, transformation).compute())
+    d = get_transformation(element, get_all=True)
+    assert isinstance(d, dict)
```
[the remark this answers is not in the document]
LucaMarconato: you are right, I have removed it

src/spatialdata/_core/operations/_utils.py
```diff
@@ -0,0 +1,117 @@
...
+    from spatialdata.transformations.transformations import BaseTransformation, Scale, Sequence, Translation
+
+    sdata = sdata.filter_by_coordinate_system(coordinate_system=coordinate_system)
+    # calling transform_to_coordinate_system will likely decrease the resolution, let's use rasterize() instead
+    sdata_vector = SpatialData(shapes=dict(sdata.shapes), points=dict(sdata.points))
+    sdata_raster = SpatialData(images=dict(sdata.images), labels=dict(sdata.labels))
+    sdata_vector_transformed = sdata_vector.transform_to_coordinate_system(coordinate_system)
+
+    de = get_extent(sdata, coordinate_system=coordinate_system)
+    de_axes = tuple(de.keys())
```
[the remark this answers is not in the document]
LucaMarconato: Done, also for `de_axes`

src/spatialdata/_core/operations/transform.py
LucaMarconato: makes mypy complain

src/spatialdata/_core/operations/transform.py
LucaMarconato: makes mypy complain

src/spatialdata/_core/spatialdata.py
LucaMarconato: makes mypy complain

src/spatialdata/_core/spatialdata.py
LucaMarconato: makes mypy complain

src/spatialdata/_core/spatialdata.py
LucaMarconato: makes mypy complain

tests/core/operations/test_transform.py
LucaMarconato: Good point, actually I have renamed it to `are_extents_equal() -> bool` and moved it to `data_extent.py`

tests/core/operations/test_transform.py
LucaMarconato: moved to utils function `remove_transformations_to_coordinate_system()`

tests/core/operations/test_transform.py
LucaMarconato: I would keep this function here, at least for the moment, because I think it's better for the user to do this operation manually. In this case we assume that len(d) == 1, but in more complex scenario you may have an intricate graph of transformations, and this function could compromise some alignments to other coordinate systems.

LucaMarconato:
Thanks for the review. I have addressed all the comments (and left 3 conversations open to see if you agree with my proposal). I'll fix the docs, otherwise if you re-approve we are good to merge.

src/spatialdata/_core/operations/transform.py
```diff
@@ -104,24 +107,24 @@ def _transform_raster(
...
     raster_translation: Translation | None,
     maintain_positioning: bool,
+    to_coordinate_system: str | None = None,
 ) -> None:
-    """Prepend a transformation to an element.
-
-    After an element has been transformed, this method is called to eventually prepend a particular transformation to
-    the existing transformations of the element. The transformation to prepend depends on the type of the element (
-    raster vs non-raster) and on the maintain_positioning flag.
+    """Adjust the transformations of an element after it has been transformed.
```
kevinyamauchi, line 118: Please add a more clear docstring. I am not sure what "adjustments" are being made

src/spatialdata/_core/operations/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+def transform_to_data_extent(
+    sdata: SpatialData,
+    coordinate_system: str,
+    target_unit_to_pixels: float | None = None,
+    target_width: float | None = None,
+    target_height: float | None = None,
+    target_depth: float | None = None,
+) -> SpatialData:
+    """
+    Transform the spatial data to match the data extent, keeping the positioning, and making all transformations equal.
```
kevinyamauchi, line 21: Can this be another operation than scale? If not, I think we should say scale. I think "tranform" is ambiguous. 

Also what is meant by "making all transformations equal"?

tests/core/operations/test_transform.py
```diff
@@ -459,14 +481,111 @@ def test_map_coordinate_systems_long_path(full_sdata):
...
+    for element in full_sdata._gen_elements_values():
+        # from spatialdata.transformations.operations import _build_transformations_graph
+        #
+        # g = _build_transformations_graph(full_sdata)
+        # import networkx as nx
+        #
+        # nx.draw(g, with_labels=True)
+        # import matplotlib.pyplot as plt
+        #
+        # plt.show()
```
kevinyamauchi, line 550: Do we need these commented lines still?

tests/core/operations/test_transform.py
```diff
@@ -459,14 +481,111 @@ def test_map_coordinate_systems_long_path(full_sdata):
...
+            images=dict(full_sdata.images),
+            labels=dict(full_sdata.labels),
+            points=dict(full_sdata.points),
+            shapes=dict(full_sdata.shapes),
+            table=full_sdata.table,
+        )
+        temp["transformed_element"] = transformed_element
+        t = get_transformation_between_coordinate_systems(temp, temp["transformed_element"], "multi_hop_space")
+        a = t.to_affine_matrix(input_axes=("x",), output_axes=("x",))
+        d = get_transformation(transformed_element, get_all=True)
```
kevinyamauchi, line 564: Can you please use more informative variable names?

kevinyamauchi (approved):
I think this looks good. I left some clarifying questions below. Please feel free to merge once you've addressed them.

melonora:
I also believe it can be merged after @kevinyamauchi comments are adressed.

src/spatialdata/_core/operations/transform.py
LucaMarconato: I renamed the function to  `_set_transformation_for_transformed_elements()` and fixed the docstring. In short, transformed elements don't have a transformation and this needs to be put inside. The transformation is not just the old transformation(s), but needs some "adjustments", which I now explain better.

src/spatialdata/_core/operations/_utils.py
LucaMarconato: In `Transform the spatial data to match the data extent, keeping the positioning, and making all transformations equal.`

The `Transform` is referring to the application of the `transform()` operation. The transformation being applied can be anything, not just scale (affine, etc).

The returned object will be such that if we have the point coordinate (x, y), this will correspond to the pixel (y, x). The problem is that in this way we lose the positioning to the original data, so we still need to add scale and translations to make the returned data align with the original data. 

I will add these information to the docstring.

LucaMarconato: I have also added an extra argument, called `maintain_positioning: bool = True`, which if false sets all the transformations to `Identity()`. I explained this argument in the docstring.

tests/core/operations/test_transform.py
LucaMarconato: removed.

tests/core/operations/test_transform.py
LucaMarconato: I agree, here it looks ugly. In general often in the code I use `t` for transformation, `a` for affine and `d` for the dict of transformations returned by `get_transformation(..., get_all=True)` (which is followed by `assert isinstance(d, dict)` to address `mypy`. So I think the `d` letter is ok.

LucaMarconato:
Thank you both for the review, I have addressed the review comments. 

I also added an argument `maintain_positioning: bool = True` to `transform_to_data_extent()`, which can be more intuitive for the user (I explained this in the docstring).

I also added some extra tests for checking better the results of `transform_to_data_extent()`, and test the new argument.

Merging now 🥳

PR #454, opened 2024-02-07
added utils scale_radii()

LucaMarconato:
Tiny PR. I just moved some already existing code into a util function that I need to use in napari-spatialdata to fix https://github.com/scverse/napari-spatialdata/issues/179.


LucaMarconato:
Merging right away because tiny. CC @giovp @kevinyamauchi

PR #459, opened 2024-02-18
add possibility for custom table name

melonora:
When performing aggregations currently it is not possible to give a custom table name. Instead a default name `table` is always given. This PR allows users to define the name of the table resulting from the aggregation.


LucaMarconato:
Nice thanks! Can you please also update `SpatialData.aggregate()`? Then good to merge!

PR #465, opened 2024-02-22
`get_centroids()`

LucaMarconato:
Function to get the centroids of a points, shapes (circles, polygons, multipolygons), labels element (single/multi scale) in a designed coordinate system.

### DIFF
--- a/src/spatialdata/_core/centroids.py
+++ b/src/spatialdata/_core/centroids.py
@@ -1,0 +1,89 @@
+from __future__ import annotations
+
+# Functions to compute the bounding box describing the extent of a SpatialElement or SpatialData object
+from functools import singledispatch
+
+from dask.dataframe.core import DataFrame as DaskDataFrame
+from geopandas import GeoDataFrame
+from multiscale_spatial_image import MultiscaleSpatialImage
+from spatial_image import SpatialImage
+
+from spatialdata._core.operations.transform import transform
+from spatialdata.models import get_axes_names
+from spatialdata.models._utils import SpatialElement
+from spatialdata.models.models import Image2DModel, Image3DModel, Labels2DModel, Labels3DModel, PointsModel, get_model
+from spatialdata.transformations.operations import get_transformation
+from spatialdata.transformations.transformations import BaseTransformation
+
+BoundingBoxDescription = dict[str, tuple[float, float]]
+
+
+def _validate_coordinate_system(e: SpatialElement, coordinate_system: str) -> None:
+    d = get_transformation(e, get_all=True)
+    assert isinstance(d, dict)
+    assert coordinate_system in d, (
+        f"No transformation to coordinate system {coordinate_system} is available for the given element.\n"
+        f"Available coordinate systems: {list(d.keys())}"
+    )
+
+
+@singledispatch
+def get_centroids(
+    e: SpatialElement,
+    coordinate_system: str = "global",
+) -> DaskDataFrame:
+    """
+    Get the centroids of the geometries contained in a SpatialElement, as a new Points element.
+
+    Parameters
+    ----------
+    e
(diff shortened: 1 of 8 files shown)


src/spatialdata/_core/centroids.py
```diff
@@ -0,0 +1,146 @@
...
+def _(e: GeoDataFrame, coordinate_system: str = "global") -> DaskDataFrame:
+    _validate_coordinate_system(e, coordinate_system)
+    t = get_transformation(e, coordinate_system)
+    assert isinstance(t, BaseTransformation)
+    # separate points from (multi-)polygons
+    first_geometry = e["geometry"].iloc[0]
+    if isinstance(first_geometry, Point):
+        xy = e.geometry.get_coordinates().values
+    else:
+        assert isinstance(first_geometry, (Polygon, MultiPolygon))
```
kevinyamauchi, line 128: Can you please add a message to this `assert` so people know what the error is for?

src/spatialdata/_core/centroids.py
```diff
@@ -0,0 +1,146 @@
...
+    all_labels = dict(zip(all_labels_values.compute(), all_labels_counts.compute()))
+    for label_value in centroids:
+        centroids[label_value] /= all_labels[label_value]
+    centroids = dict(sorted(centroids.items(), key=lambda x: x[0]))
+    return pd.DataFrame({axis: centroids.values()}, index=list(centroids.keys()))
+
+
+@get_centroids.register(SpatialImage)
+@get_centroids.register(MultiscaleSpatialImage)
+def _(
```
kevinyamauchi, line 95: For the dispatch methods, I think it would be nice if we added a comment/docstring for each dispatch method to state which element they are for. I know the types are registered, but not everyone will know that points are in a Dask DataFrame (for example).

tests/core/test_centroids.py
```diff
@@ -0,0 +1,155 @@
...
+@pytest.mark.parametrize("is_3d", [False, True])
+def test_get_centroids_points(points, coordinate_system: str, is_3d: bool):
+    element = points["points_0"]
+
+    # by default, the coordinate system is global and the points are 2D; let's modify the points as instructed by the
+    # test arguments
+    if coordinate_system == "aligned":
+        set_transformation(element, transformation=affine, to_coordinate_system=coordinate_system)
+    if is_3d:
+        element["z"] = element["x"]
```
kevinyamauchi, line 27: I personally am not a big fan of tests with many conditionals because I find them difficult to read. I think `parameterize` is better used when multiple inputs can be passed through the same code (you can pass the expected results  as input arguments as well).

I think this would be easier to read if there were two sets of tests, one for global and one for aligned. you can then pass in 2D and 3D points with their expected values as inputs. Something like

```python
@pytest.mark.parametrize("points,expected", [(points_2d, expected_points_2d), (points_3d, expected_points_3d)])
def test_get_centroids_points_global(points, expected):
    # do stuff

@pytest.mark.parametrize("points,expected", [(points_2d, expected_points_2d), (points_3d, expected_points_3d)])
def test_get_centroids_points_aligned(points, expected):
    # do stuff
```

That being said, I think it's fine if you prefer this style. I don't want to block the PR - just wanted to give an alternative.

kevinyamauchi (approved):
LGTM! I left some minor comments below and an optional suggestion on test style. Please feel free to merge when you're ready. Thanks, @LucaMarconato !

tests/core/test_centroids.py
LucaMarconato: Thanks for the comment, you are right the logic of `global` vs `aligned` is different and it would have been better to split the tests. I In the context of this PR I prefer to keep it like this because this PR is blocking other PRs and we need to make a release, but for the next one I will try to reduce the use of `parametrize` to the cases that don't create code bifurcation.

PR #466, opened 2024-02-22
allow table to be None in get_values and _locate_values

melonora:
The `table_name` in `get_values` and `locate_values` was not allowed to be None, but this can sometimes be a requirement when for example there is no table annotating the element. This can be the case in `spatialdata_plot`. This PR allows for this.

This PR also fixes the issue of failing docs build due to `import error, no module called SpatialData`. In order to replicate the issue I tried the doc build on the denbi machine `using spinx-build source-dir out-dir -P`. The `-P` flag allows for a debugger without ide allowing to manually try the import of `spatialdata`. This gave an error because there was no module called `pytest`. Adding `sphinx-pytest` to the doc dependencies fixed the issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ docs = [
     # For notebooks
     "ipython>=8.6.0",
     "sphinx-copybutton",
+    "sphinx-pytest",
 ]
 test = [
     "pytest",
--- a/src/spatialdata/_core/operations/aggregate.py
+++ b/src/spatialdata/_core/operations/aggregate.py
@@ -64,7 +64,7 @@ def aggregate(
     region_key: str = "region",
     instance_key: str = "instance_id",
     deepcopy: bool = True,
-    table_name: str = "table",
+    table_name: str | None = None,
     **kwargs: Any,
 ) -> SpatialData:
     """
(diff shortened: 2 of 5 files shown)


LucaMarconato:
Well spotted! Please let me know when it's ready for review 😊

melonora:
@LucaMarconato could you please comment on the last commit? I think this check is valid but currently some tests fail because the table does not annotate the elements being aggregated.

melonora:
after that I will update docstrings, update changelog etc.

LucaMarconato:
> @LucaMarconato could you please comment on the last commit? I think this check is valid but currently some tests fail because the table does not annotate the elements being aggregated.

The fact that the values are passed as the tuple `values_sdata: SpatialData`, `values: str` instead of `values: SpaitalElement` doesn't imply that the table is actually used. Here one of the test could fail because the values to aggregate may be fetched from one of the columns of `values`.

melonora:
The PR also includes a fix for the failing doc builds now. This is due to a missing dependency `sphinx-pytest`

LucaMarconato (approved):
Looks great! Very nice! 🎉

PR #468, opened 2024-02-23
Added `validate_table_annotation_target()`

LucaMarconato:
Added function to validate the target(s) of a table:
- checks if it exists in the SpatialData object
- check if the dtype of the instances match

This function is called internally in the `__init__` of SpatialData but is also available for downstream use.

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -1574,6 +1574,50 @@ class SpatialData:
         )
         return SpatialData(**elements_dict, tables=tables)
 
+    def validate_table_annotation_target(self, table_name: str) -> None:
+        """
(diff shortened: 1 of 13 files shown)


src/spatialdata/_core/spatialdata.py
```diff
@@ -1574,6 +1574,50 @@ def subset(
         )
         return SpatialData(**elements_dict, tables=tables)
 
+    def validate_table_annotation_target(self, table_name: str) -> None:
```
melonora, line 1577: maybe check how much this compares to `validate_table_in_spatialdata`. Part of the functionality overlaps

LucaMarconato: Thanks! I removed my function and put the extra functionalities inside yours.

melonora (approved):
Looks good to me now!

PR #471, opened 2024-02-25
Fixed and tested index bug points model

### LINKED ISSUE #470: `PointsModel.parse()` resets the index when passing a dataframe
`PointsModel.parse()` resets the index when passing a dataframe

LucaMarconato:
Small PR, the indices of points were reset by the parser.

Closes #470.


melonora (approved):
Seems good to me

PR #476, opened 2024-03-13
Removed int contraint for instance_key column

LucaMarconato:
@melonora @giovp I realized that if we force instance_key to be an int we need to update the storage format because some technologies, in particular Xenium, relies on having instance_key as string. I would therefore, unless this leads to problems in napari-spatialdata, etc, to keep allowing strings for instance_key and eventually change this in the future, when the versioning of the data format is improved and we can provide a migration tool for the data.


LucaMarconato:
Closing this as the change will be directly applied inside https://github.com/scverse/spatialdata/pull/485 (where @melonora verified that the join operations are correct also when the `instance_key` has `dtype` string).

PR #481, opened 2024-03-13
fix dask dataframe dependency

LucaMarconato:
Dask DataFrame is now using `dask-expr` as backend; pinning down the version temporarily until we address the changes required to support it (example there is no `.attrs` in the dataframe).


melonora:
Added #482 as tasks to adjust the code base for compatibility.

melonora (approved):

Issue #308, opened 2023-06-22
Coordinate transformations refactoring


LucaMarconato:

LucaMarconato:
The coordinate transformation refactoring will implement the following:

# Data model
- [ ] `elements.attrs['coordinate_system']` (or maybe with the shorter name `'cs'`), will be added; this will be a single 'string'
- [ ] `elements.attrs['transform'] will be removed; the information on coordinate transformation will be moved to the `SpatialData` level; transformations will be defined between coordinate systems, *not* between elements and coordinate systems, coordinate systems and elements, nor elements and elements
- [ ] the above changes require raster types to use `xarray` coordinates
- [ ] there will be a helper function to convert back and forth between `BaseTransformation` (and so also `NGFFBaseTransformation`) and `xarray` coordinates
- [ ] on disk, `xarray` coordinates will not be saved (this is not supported by NGFF); instead the above conversion will be used

# APIs
- [ ] there will be a helper function to transfer the coordinate transformations metadata easily between `SpatialData` objects. In fact, elements will not contain coordinate transformations anymore.
- [ ] `SpatialData.transform_element_to_coordinate_system()` will be incorporated into `SpatialData.transform_to_coordinate_system()`, by the use of extra arguments. Also it will become a wrapper around `transform()`.
- [ ] `transform(data, transformation, to_coordinate_system)` will have the following behavior
	- [ ] if `transformation` is passed, `to_coordinate_system` must be `None`. This will modify the data in-place, and not involve/modify any coordinate system transformation. It's ok to call this function if there is only one coordinate system, but if multiple coordinate systems are being used, this usage will probably be undesirable, so we should raise a warning.
	- [ ] if `transformation` is `None`, `to_coordinate_system` must be provided. This will transform the data to the desired coordinate system (=change the xarray coordinates, e.g. translations/scale; and also change the data when required, e.g. rotation); also the coordinate system metadata for the element will be updated.
- [ ] The parameter `maintain_positioning` must be removed, because now we can't modify one element and compensate the transformations departing from it with the inverse transformations. In fact, now the transformations will involve the whole coordinate system. We can consider adding an helper function that mimics the current behavior of `maintain_positioning`

# File format
- [ ] The NGFF transformation specification is not merged yet; check if we have to update the io functions to the new draft specs (or if it is merged to main by that time, implement the final version of the specs).

CC @melonora

LucaMarconato:
[...]

PR #491, opened 2024-03-18
Update design_doc.md

LucaMarconato:
- Added multiple tables
- Removed `.obsm['spatial']` references (it won't be needed after the images will live in extrinsic coordinate systems, thanks to the support of xarray coordinates with the planned refactoring https://github.com/scverse/spatialdata/issues/308).
- Adjusted priorities; dropped initial ideas that fell out of scope


LucaMarconato:
CC @giovp @kevinyamauchi.

Issue #486, opened 2024-03-14
Columns order for points element shuffled sometimes after parser/deepcopy


LucaMarconato:
Sometimes (I can't reproduce consistently, it happens rarely), the tests for `deepcopy()` fail because the deepcopied points have columns with shuffled order. See for instance here: https://github.com/scverse/spatialdata/actions/runs/8281155947/job/22659046336?pr=455

I can't reproduce. I think the problem is in the parser, but since the data is still consistent, I will add a note in the docstring of the parser for points and of the deepcopy for points saying that the order of the columns is sometimes not preserved. This could be also a bug of dask dataframe, and maybe with the newly released `dask-expr` implementation it will disappear.

PR #501, opened 2024-03-21
reimplemented incremental io

### LINKED ISSUE #186: Discussion: how incremental addition/modification of elements and IO interact
This is to collect points to discuss in major redesign of how IO interacts with the capabilities of extending an object by adding/removing or modifying particular elements. In other words, we need to decide how the user can modify/extend a `SpatialData` object after its constructor is called, and how this relates with IO.

It evolves from previous discussions in this old issue https://github.com/scverse/spatialdata/issues/137 (and it's related PR https://github.com/scverse/spatialdata/pull/138).

I'll just dump all the points and reformat later on.

* comments on `add_image(), add_labels()` 
    * these functions are immediately writing to disk when called. We could decide removing these functions and let the user write to disk explicitly. But it would be cool to know what is to be written to disk, and what not.
    * the functions also check if the name of the image being added is already present in `sdata.images` or not, or if it is unique across all elements. This check is important, but currently it can be overridden by simply doing `sdata.images['my_image'] = image`. We should decide on one of the following:
        * we make `.images` private and we remove `sdata.add_images('my_image', image)`. The user will not use `sdata.images['my_image']` anymore, leaving only with the option to do `sdata['my_image'] = image`, which internally uses `get_schema()` to know what `image` is.
        * we replace `.images` with an accessor class for which we override `__set_item__()` and `__get_item__()`, and we make these `__set_item__()` call `add_image()` internally.

### LINKED ISSUE #496: Re-enable modular io for elements and transformation
The modular saving of elements and transformations has been refactored out of the previous release but it didn't make yet in the new codebase. Re-enabling this is a high priority.

### LINKED ISSUE #498: automatic backing of sdata.tables
I was testing the new mutlitple tables feature on the main branch, and noticed that automatic backing of  sdata.tables to the backing zarr store no longer works (or I fail to get it working). This is how it used to work:
```

import anndata as ad
import numpy as np
import pandas as pd
import spatialdata
import os

path=""

n_cells = 100
n_genes = 50

# Creating random data
X = np.random.randn(n_cells, n_genes)
obs = pd.DataFrame(index=[f'cell_{i}' for i in range(n_cells)])
var = pd.DataFrame(index=[f'gene_{i}' for i in range(n_genes)])

adata = ad.AnnData(X=X, obs=obs, var=var)

adata.obs['cell_type'] = ['type1' if i < n_cells // 2 else 'type2' for i in range(n_cells)]
adata.var['highly_variable'] = np.random.choice([True, False], size=n_genes)


sdata=SpatialData()
sdata.write( os.path.join( path, "sdata_dummy.zarr" ) )

sdata.tables[ "dummy_table" ] = spatialdata.models.TableModel.parse( adata,
)

assert sdata.table is not None # this is ok

sdata=SpatialData.read( os.path.join( path, "sdata_dummy.zarr" ) )
sdata

assert sdata.table is not None # this gives assertionerror
```

Is this related to https://github.com/scverse/spatialdata/issues/496?

LucaMarconato:
Closes #186
Closes #496
Closes #498

Support for incremental io operations.

New features:
- [x] ability to save additional elements to disk after the `SpatialData` object is created
- [x] ability to remove from disk previously saved objects
- [x] ability to see which elements are present only in-memory and not in the Zarr store and viceversa
- [x] refactored saving of metadata:
    - [x] transformations
    - [x] consolidated metadata
    - [x] set the basis (but not implemented), like empty tests or TODOs with what's missing, for the other metadata: table.uns['spatialdata_attrs'], points.attrs['spatialdata_attrs'] and OMERO metadata for image channel names 

Robustness:
- [x] refactored write function to make it more robust
- [x] improved error messages for the users, with actionable advices
- [x] new concept of "self-contained" `SpatialData` object and "self-contained" elements. Useful for the user to understand the implications of file backing
- [x] added info on Dask-backed files for non "self-contained" elements to `__repr__()`

Testing:
- [x] improved existing tests for io
- [x] extensive testing for modular io
- [x] improved testing for comparision of metadata after io and after a deepcopy

Other:
- [x] fixed bug of points columns being shuffled after a query operation #486 

This PR also sets the basis for (not implemented here) the ability to load in-memory objects that are Dask-backed.


LucaMarconato:
@ArneDefauw @aeisenbarth tagging you because you opened at some point a issue regarding incremental IO. In this PR incremental IO is implemented, happy to receive feedback in case you want to play around with it😊 

I will make a notebook to showcase it, but in short to save an element `lables`, `table`, etc you can use the new `sdata.write_element('element_name')`. If the element already exists in the storage an exception will be raised. You can work around the exception for instance with the strategies shown here: https://github.com/scverse/spatialdata/blob/87dd1a88b41139d2657b56a5191a14e239aa26a2/tests/io/test_readwrite.py#L159.

Please note that those strategies are not guarantee to work in various scenarios, including multi-threaded application, network storages, etc. So please use with care.

Currently the whole table needs to be replaced and the whole table needs to be stored in-memory, but recent progress in `anndata` + `dask` will be used also in `spatialdata` to allow lazy loading and the replacement of particular elements (like adding a single obs column). This PR clean up the previous code and is a step in that direction.

LucaMarconato:
Currently the whole table needs to be replaced and the whole table needs to be stored in-memory, but recent progress in `anndata` + `dask` will be used also in `spatialdata` to allow lazy loading and the replacement of particular elements (like adding a single obs column). This PR clean up the previous code and is a step in that direction.

src/spatialdata/_utils.py
```diff
@@ -156,7 +156,11 @@ def _compute_paddings(data: DataArray, axis: str) -> tuple[int, int]:
 def multiscale_spatial_image_from_data_tree(data_tree: DataTree) -> MultiscaleSpatialImage:
     d = {}
     for k, dt in data_tree.items():
-        v = dt.values()
+        try:
+            v = dt.values()
+        except TypeError as e:
+            pass
+            raise e
```
kevinyamauchi, line 163: What is the purpose of this? Is it just re-raising the same error?

src/spatialdata/_io/_utils.py
```diff
@@ -308,11 +343,14 @@ def save_transformations(sdata: SpatialData) -> None:
...
     """
-    from spatialdata.transformations import get_transformation, set_transformation
-
-    for element in sdata._gen_spatial_element_values():
-        transformations = get_transformation(element, get_all=True)
-        set_transformation(element, transformations, set_all=True, write_to_sdata=sdata)
+    warnings.warn(
+        "This function is deprecated and should be replaced by `SpatialData.write_transformations()` or "
+        "`SpatialData.write_metadata()`, which gives more control over which metadata to write. The execution will "
+        "conitnue by calling the former",
```
kevinyamauchi, line 349: > The execution will conitnue by calling the former

I'm not sure what this means. Also see the small typo below

```suggestion
        "`SpatialData.write_metadata()`, which gives more control over which metadata to write. The execution will "
        "continue by calling the former",
```

src/spatialdata/_core/spatialdata.py
```diff
@@ -931,195 +879,557 @@ def transform_to_coordinate_system(
                 elements[element_type][element_name] = transformed
         return SpatialData(**elements, tables=sdata.tables)
 
-    def write(
+    def describe_elements_are_self_contained(self) -> dict[str, bool]:
```
kevinyamauchi, line 882: Minor nitpick: I'm not a huge fan of this name. Perhaps something more like `elements_are_self_contained()`

src/spatialdata/_core/spatialdata.py
```diff
@@ -492,93 +496,83 @@ def aggregate(
...
     @property
     def path(self) -> Path | None:
         """Path to the Zarr storage."""
-        return self._path
+        if self._path is None:
+            return None
+        if isinstance(self._path, str):
+            return Path(self._path)
+        if isinstance(self._path, Path):
+            return self._path
```
kevinyamauchi, line 511: I would prefer that we do the type coercion on the setter. That is, `SpatialData._path` should always be a `Path` object. That way the internal representation is consistent.

kevinyamauchi (commented):
Thanks @LucaMarconato ! I left a review with some minor points below. I think it looks good, but i didn't have time for a super in depth review. Given that this is a big change, I think the approval should be given by somebody who can look more closely.

melonora:
> Thanks @LucaMarconato ! I left a review with some minor points below. I think it looks good, but i didn't have time for a super in depth review. Given that this is a big change, I think the approval should be given by somebody who can look more closely.

Thanks for the review @kevinyamauchi. I will have a look at this PR later today as well.

ArneDefauw:
> @ArneDefauw @aeisenbarth tagging you because you opened at some point a issue regarding incremental IO. In this PR incremental IO is implemented, happy to receive feedback in case you want to play around with it😊
> 
> I will make a notebook to showcase it, but in short to save an element `lables`, `table`, etc you can use the new `sdata.write_element('element_name')`. If the element already exists in the storage an exception will be raised. You can work around the exception for instance with the strategies shown here:
> 
> https://github.com/scverse/spatialdata/blob/87dd1a88b41139d2657b56a5191a14e239aa26a2/tests/io/test_readwrite.py#L159
> 
> Please note that those strategies are not guarantee to work in various scenarios, including multi-threaded application, network storages, etc. So please use with care.




Thanks for the quick response and fix!

I've tested the incremental io for my use case, and up to now everything seems to works as expected, except one thing. If I follow the approach suggested here: 

https://github.com/scverse/spatialdata/blob/87dd1a88b41139d2657b56a5191a14e239aa26a2/tests/io/test_readwrite.py#L159

I get a ValueError when I load my SpatialData object back from the zarr store and try to overwrite it:
`
ValueError: The file path specified is a parent directory of one or more files used for backing for one or more elements in the SpatialData object. Deleting the data would corrupt the SpatialData object.`


The fix was to first delete the attribute from the SpatialData object, and then remove the element on disk. Minimal example below of a typical workflow in my image processing pipelines:

```
import spatialdata
import dask.array as da

img_layer="test_image"

sdata=SpatialData()

sdata.write( os.path.join( path, "sdata.zarr" ) )

dummy_array = da.random.random(size=(1,10000, 10000), chunks=(1,1000, 1000))

se=spatialdata.models.Image2DModel.parse(
                data=dummy_array,
            )

sdata.images[ img_layer ] = se

if sdata.is_backed():
    sdata.write_element("test_image", overwrite=True)

# need to read back from zarr store, otherwise graph in the in-memory sdata would not be executed
sdata=read_zarr( sdata.path )

# now overwrite:

# Here I needed to first delete the attribute:

# first delete attribute
element_type = sdata._element_type_from_element_name(  img_layer )
del getattr(sdata, element_type)[ img_layer ]
# then on disk
if sdata.is_backed():
    sdata.delete_element_from_disk( img_layer )


sdata.images[ img_layer ] = se

if sdata.is_backed():
    sdata.write_element(img_layer, overwrite=True)

sdata=read_zarr( sdata.path )
```

I think what the unit test you refered to lacks is the reading back from the zarr store, after an element is written to a zarr store.

In version 0.0.15 of SpatialData, when sdata.add_image(...) was executed, it was not necessary to read back from the zarr store. I understand that current implementation allows for more control, but the inplace update of the SpatialData object was kinda convenient.

src/spatialdata/_io/_utils.py
LucaMarconato: "The execution will continue by calling the former" with "This function will call `SpatialData.write_transformations()`; please call this function directly."

src/spatialdata/_core/spatialdata.py
LucaMarconato: Good point! Done.

src/spatialdata/_utils.py
LucaMarconato: I forgot it there while debugging, I will remove it.

LucaMarconato:
Thanks for the reviews. I addressed the points from @ArneDefauw and merged his PR here. @giovp, when you have time could you also please give a pass to this?

tests/io/test_readwrite.py
```diff
@@ -28,134 +27,198 @@
...
-                sdata[f"incremental_{k}"] = v
-            break
+                sdata.points[f"additional_{k}"] = v
+                sdata[f"additional_{k}"] = v
+            with pytest.raises(KeyError, match="Key `table` already exists."):
+                sdata["table"] = v
+
+    @pytest.mark.parametrize("dask_backed", [True, False])
+    @pytest.mark.parametrize("workaround", [1, 2])
+    def test_incremental_io_on_disk(
```
giovp, line 136: this should 💯 be a tutorial

src/spatialdata/_core/spatialdata.py
```diff
@@ -962,195 +908,564 @@ def transform_to_coordinate_system(
...
+        return description
+
+    def is_self_contained(self, element_name: str | None = None) -> bool:
+        """
+        Check if an object is self-contained; self-contained objects have a simpler disk storage layout.
+
+        A SpatialData object is said to be self-contained if all its SpatialElements or AnnData tables are
+        self-contained. A SpatialElement or AnnData table is said to be self-contained when it does not depend on a
+        Dask computational graph (i.e. it is not "lazy") or when it is Dask-backed and each file that is read in the
+        Dask computational graph is contained within the Zarr store associated with the SpatialElement.
```
giovp, line 938: could this description fo self-containing also be in the `elements_are_self_contained` above?

src/spatialdata/_core/spatialdata.py
```diff
@@ -962,195 +908,564 @@ def transform_to_coordinate_system(
...
+                    "Please note that only Zarr stores not currently in used by the current SpatialData object can be "
+                    "overwritten."
                 )
-            if any(Path(fp).resolve().is_relative_to(file_path.resolve()) for fp in get_dask_backing_files(self)):
+            ERROR_MSG = (
+                "Cannot overwrite. The target path of the write operation is in use. Please save the data to a "
+                "different location. "
+            )
+            WORKAROUND = (
+                "\nWorkaround: please see discussion here https://github.com/scverse/spatialdata/discussions/520."
```
giovp, line 1072: this link is dead, I think because of the .

giovp, line 1072:
```suggestion
                "\nWorkaround: please see discussion here https://github.com/scverse/spatialdata/discussions/520 ."
```

src/spatialdata/_core/spatialdata.py
```diff
@@ -962,195 +908,564 @@ def transform_to_coordinate_system(
...
-                    )
+    def _check_element_not_on_disk_with_different_type(self, element_type: str, element_name: str) -> None:
+        only_on_disk = self.elements_paths_on_disk()
+        for disk_path in only_on_disk:
+            disk_element_type, disk_element_name = self._element_type_and_name_from_element_path(disk_path)
+            if disk_element_name == element_name and disk_element_type != element_type:
+                raise ValueError(
+                    f"Element {element_name} is found in the Zarr store as a {disk_element_type}, but it is found "
+                    f"in-memory as a {element_type}. The in-memory object should have a different name."
+                )
```
giovp, line 1292: I think this is a bit tricky, if a user see this error message, what should they do?

ArneDefauw:
> Thank you @ArneDefauw for trying the code and for the explanation, I will now look into your PR.
> 
> > In version 0.0.15 of SpatialData, when sdata.add_image(...) was executed, it was not necessary to read back from the zarr store. I understand that current implementation allows for more control, but the inplace update of the SpatialData object was kinda convenient.
> 
> The reason why we refactored this part is because with `add_image()`, if the user had a in-memory image and was writing it to disk, the image would have then immediately lazy loaded. This is good and ergonomic if the image needs to be written only once, but if the user tries to write image again (for instance in a notebook when a cell may get manually executed twice), then it would have lead to an error.

Hi @LucaMarconato  ,
thanks for the reply and the fixes!
I've tested your suggestions (https://github.com/scverse/spatialdata/blob/582622f689a7e05421e9d066f98baf702549978f/tests/io/test_readwrite.py#L136) for my use cases and everything seems to work fine (both for images,labels, points, labels and shapes ).

Workaround 1 , looks rather safe in most scenarios. If I understand correctly it covers following scenario:
Having "x" in sdata, doing something on "x" (i.e. defining a dask graph), and then writing to "x".

How I would usually work is:
Having "x" in sdata, doing something on "x", and writing to "y" (where "y" already exists).

The latter feels less dangerous, and looka pretty standard in image processing pipelines, e.g. tuning of hyper parameters for image cleaning or segmentation.

In the latter case, I guess the following would be sufficient:
```

arr=sdata["x"].data
arr=arr*2
spatial_element=spatialdata.models.Image2DModel.parse(
                arr,
            )
del sdata["y"]
sdata.delete_element_from_disk("y")
sdata["y"]=spatial_element
sdata.write_element("y")
sdata=read_zarr( sdata.path )
```

LucaMarconato:
Yes, I agree that the approach that you described is generally a good practice when processing data and safe, since the original data is not modified.

The use cases that I described are instead for the case in which the data itself is replaced. I think I should add in the comments that this approach should be avoided when possible, and clarify that the workaround that I described are only if really needed.

namsaraeva:
Thank you for this PR, I am using it right now. One question: would it be possible to pass a list of strings onto `write_element()` instead of just one element? @LucaMarconato

LucaMarconato:
@namsaraeva thanks for the suggestion, it's indeed handier to have list of names. I have added the support for this for `write_element()` and `delete_element_from_disk()`.

melonora:
personally I don't see any blockers currently for this PR.

melonora (approved):

LucaMarconato:
> personally I don't see any blockers currently for this PR.

@melonora I wanted to check this PR before merging: https://github.com/scverse/spatialdata/pull/525/files. I will do this this weekend.

tests/io/test_readwrite.py
LucaMarconato: I agree; for the moment I refer to this code in this GitHub conversation: https://github.com/scverse/spatialdata/discussions/520. Also, in the models notebooks I refer to the same GitHub conversation. The user would have to jump a bit but at least this code is discoverable.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Fixed it now.

src/spatialdata/_core/spatialdata.py
LucaMarconato: I added an actionable explanation: "If you want to maintain both objects, please rename the in-memory object. Alternatively, you can rename the element on disk (manually)."

Issue #483, opened 2024-03-13
Replace some `AssertionError` with other exceptions


LucaMarconato:
As a user reported (I can't find the discussion at the moment), assertions are ignored when the code runs in production. We should verify that no assertion is actually required for the logic of the execution, and in case they are required, replace them with other types of exception. Many assertions are used for `mypy`, but I see some of them are required for a correct execution, so better to replace them.

PR #512, opened 2024-03-25
Join outside sdata

melonora:
This PR adds a function `join_spatialelement_table` which is equivalent to `join_sdata_spatialelement_table` but does not require the elements to be part of a SpatialData object.

### DIFF
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -19,6 +19,8 @@ from spatial_image import SpatialImage
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata._utils import _inplace_fix_subset_categorical_obs
 from spatialdata.models import (
+    Image2DModel,
+    Image3DModel,
     Labels2DModel,
     Labels3DModel,
     PointsModel,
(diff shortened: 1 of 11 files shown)


src/spatialdata/_core/query/relational_query.py
```diff
@@ -462,14 +464,78 @@ class MatchTypes(Enum):
...
 
-def join_sdata_spatialelement_table(
-    sdata: SpatialData,
-    spatial_element_name: str | list[str],
-    table_name: str,
-    how: str = "left",
+def _create_sdata_elements_dict_for_join(
+    sdata: SpatialData, spatial_element_name: str | list[str], table_name: str
+) -> tuple[dict[str, dict[str, Any]], AnnData]:
+    assert sdata.tables.get(table_name), f"No table with `{table_name}` exists in the SpatialData object."
```
LucaMarconato, line 470: I'll change this to a `ValueError`, see https://github.com/scverse/spatialdata/issues/483

src/spatialdata/_core/query/relational_query.py
```diff
@@ -462,14 +464,78 @@ class MatchTypes(Enum):
...
+                UserWarning,
+                stacklevel=2,
+            )
+        else:
+            element_type, _, element = sdata._find_element(name)
+            elements_dict[element_type][name] = element
+    return elements_dict, table
+
+
+def _create_elements_dict_for_join(
```
LucaMarconato, line 494: This function, except for the warnings part, can be replaced by `SpatialData.init_from_elements()`.

LucaMarconato: I did this.

PR #515, opened 2024-03-26
update test read write on disk

ArneDefauw:
update unit test incremental io on disk, see https://github.com/scverse/spatialdata/pull/501#issuecomment-2016568527

### DIFF
--- a/tests/io/test_readwrite.py
+++ b/tests/io/test_readwrite.py
@@ -147,7 +147,7 @@ class TestReadWrite:
         ]:
             sdata[name] = full_sdata[name]
             sdata.write_element(name)
-            sdata=read_zarr( sdata.path )
+            sdata = read_zarr(sdata.path)
 
             with pytest.raises(
                 ValueError, match="The Zarr store already exists. Use `overwrite=True` to try overwriting the store."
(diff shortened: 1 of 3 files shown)


ArneDefauw:
I changed status of PR, there is not really something missing, but for points, there still seem to be issues with workaround 2 ,as suggested, I am not sure if this is expected or not...

tests/io/test_readwrite.py
```diff
@@ -147,13 +147,14 @@ def test_incremental_io_on_disk(self, tmp_path: str, full_sdata: SpatialData):
...
             sdata.write_element(name)
+            sdata = read_zarr(sdata.path)
 
             with pytest.raises(
                 ValueError, match="The Zarr store already exists. Use `overwrite=True` to try overwriting the store."
             ):
                 sdata.write_element(name)
 
-            with pytest.raises(ValueError, match="Currently, overwriting existing elements is not supported."):
+            with pytest.raises(ValueError):  # TODO this gives different messages for different element types
```
LucaMarconato, line 157: This is confusion, I will adjust it, but correct. Points, images and labels may be backed from disk. In that case the problem is that even with `overwrite=True`, we can't overwrite those elements, because overwriting them would mean to actually delete the source of the backed data.

The second error message occurs with non-backed images, labels, points or with shapes and tables (that currently are always non-backed). In that case we could overwrite the data but we don't do it because this could lead to data loss anyway (the case above is instead a "systematic data loss").

The only overwrite case that is supported is overwriting into a different Zarr store, for instance an old, non-loaded, spatialdata object.

LucaMarconato: The workarounds I show in the tests work for both cases; I will document them in an example notebook and eventually we could automate their execution (after warning the user that this works only under particular assumptions).

LucaMarconato: I will now make the error messages less cryptic.

LucaMarconato: Done.

LucaMarconato:
Thanks for reporting and for the code. The use case of wanting to overwrite lazy data is an important one and now this is considered in tests. For clarify, I have split the test into 4 subcases: workaround 1 and 2, and lazy vs non-lazy data. In this way the user can see which case to use and have some code to start with.

A few more comments:
- the error messages, now made clearer, point to a discussion (https://github.com/scverse/spatialdata/discussions/520), which currently points to the test considered in this PR, to show some code to start with.
- currently we don't allow to reload single elements, or to write the sdata/element and immediately reload it; I plan to add this in the future. Tracking it here: https://github.com/scverse/spatialdata/issues/521.
- Finally, as now it is also being clarified in the test, the workaround 2 should **not** be used for lazy data. It is expected for it to fail for points, images and labels. Unfortunately the approach that you suggested for workaround 2 leads to data loss. The error given for points is expected and I was surprised it was working for images and labels. I looked into it better: somehow the `xarray.DataArray` doesn't throw an error when it tries to load from the Zarr store that had been deleted, but if you compute the data you will see that the tensor is now a tensor of zeros.

PR #517, opened 2024-03-26
update image IO to write and read omero metadata correctly

giovp:
Late followup PR of spatialdata from this: https://github.com/ome/ome-zarr-py/pull/261

also should point out that this is current only in the latest spec, not v0.4:  https://github.com/ome/ngff/pull/202


src/spatialdata/_io/io_raster.py
```diff
@@ -105,7 +110,6 @@ def _write_raster(
     fmt: Format = CurrentRasterFormat(),
     storage_options: Optional[Union[JSONDict, list[JSONDict]]] = None,
     label_metadata: Optional[JSONDict] = None,
-    channels_metadata: Optional[JSONDict] = None,
```
giovp, line 108: removed cause it's unused

src/spatialdata/_io/io_raster.py
```diff
@@ -130,9 +134,12 @@ def _get_group_for_writing_transformations() -> zarr.Group:
             return group.require_group(name)
         return group["labels"][name]
 
-    # convert channel names to channel metadata
+    # convert channel names to channel metadata in omero
     if raster_type == "image":
-        group_data.attrs["channels_metadata"] = fmt.channels_to_metadata(raster_data, channels_metadata)
+        metadata["metadata"] = {"omero": {"channels": []}}
```
giovp, line 139: so "technically" here I should be checking that metadata doesn't already have this key, and also with respect to the omero dict, yet the only time that the `_write_raster` is exposed, via [`_write_image`](https://github.com/scverse/spatialdata/blob/18b32e76d8b87600b15f5253e5e3b582f8d7bf1f/src/spatialdata/_core/spatialdata.py#L1042), it is not allowed for the metadata to be passed. What would you suggest? I don't really see a real use case, so I would suggest to keep it as is and use them effectively as kwargs as it is now.

giovp:
uh oh, I thought 0.8.4 was release with that fix with https://github.com/ome/ome-zarr-py/pull/261#issuecomment-1825869715 but apparently it was now, shall I ask Josh or we want to put this on hold?

LucaMarconato:
Thanks for the PR, great to see this implemented. I would ask if a 0.8.4 can be released and wait to merge this PR until then.

src/spatialdata/_io/io_raster.py
LucaMarconato: I agree, I also would keep this as it is.

LucaMarconato (approved):
All good for me.

LucaMarconato:
Now `ome_zarr` 0.8.4 (and more) is released. If the tests pass we are good to merge. @giovp

giovp:
ah saw it, yes LGTM then!

PR #525, opened 2024-03-29
Incremental io

ArneDefauw:
PR for illustration of possible workflow for adding elements to spatialdata object inspired by https://github.com/scverse/spatialdata/blob/582622f689a7e05421e9d066f98baf702549978f/tests/io/test_readwrite.py#L136 and related to discussion https://github.com/scverse/spatialdata/discussions/520.

Currently only implemented for image and labels, but similar approaches would work for points, shapes and tables.

See unit test for example usage.
https://github.com/ArneDefauw/spatialdata/blob/2e71e63ca9ccbd849879efe1f26c7221cdd60b34/tests/io/test_readwrite.py#L662

Approach also allows for writing to multiscale, and avoids recomputation via writing to intermediate slot.

Some things are not covered yet, such as automatic reload, and proper passing of transformations to `spatialdata.models.Image2Dmodel.parse`

I am still unsure to what extent spatialdata should implement such workflows (e.g. overwriting elements on disk), or should merely give suggestions about possible workflows (via notebooks/documentation)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,31 @@
+name: Release
+
+on:
+    release:
+        types: [published]
+
+jobs:
+    release:
+        # requires that you have put your twine API key in your
+        # github secrets (see readme for details)
+        runs-on: ubuntu-latest
+        if: contains(github.ref, 'tags')
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v3
+
+            - name: Set up Python 3.10
+              uses: actions/setup-python@v4
+              with:
+                  python-version: "3.10"
+
+            - name: Install hatch
+              run: pip install hatch
+
+            - name: Build project for distribution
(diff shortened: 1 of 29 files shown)


LucaMarconato:
Notes from the in-person discussion.

- [ ] we want to give more control to the user for in-memory vs non-memory objects, by giving control of how each element is stored https://github.com/scverse/spatialdata/issues/293 We should consider also the option to make the objects lazy but persistent (`.persist()`.
- [ ] we identified a possible bug in ome-zarr-py that could lead up to ~10x slower writing performance when dealing with large multiscale images; I will report it
- [ ] the overwrite behavior got tested by @ArneDefauw on large data pipelines and no problems were highlighted; we should enable it inside `write()` and `write_element()` in a follow-up PR.

Thanks again for the demo of this PR and for bringing attention to the above topics.

Issue #311, opened 2023-06-27
Option to save columns in shapes layer


lopollar:
I think this was mentioned before, but I couldn't find it back.
When adding columns with annotation to the shapes layer, these aren't saved. 
I have different types of shapes I like to add, where all of them are labeled. 
I now have to save them as all separate shapes layer, but one shapes layer with an annotation column would be cleaner. 
Is this an option for the future?

LucaMarconato:
Hi Lotte, thanks for reporting. We are soon going to improve the IO in a major refactoring and we will make sure that this is possible. You can see the progress of the refactoring here https://github.com/scverse/spatialdata/issues/293 (I think it will take a few weeks, please be patient).

EliHei2:
I think it would be great to have the flexibility to add more columns to the shepes' geopandas dataframes, this is super useful, as sometimes having shapes that are not already in the table, one needs to associate some metadata to them. Having two versions segmentation masks for cells, and annotating them is one example. Obviously, one can add columns to the gdfs when in memory, but writing and reading again it doesn't load the added columns.

PR #542, opened 2024-04-08
GeoParquet support for shapes; refactor format

LucaMarconato:
This PR introduces the use of GeoParquet to save shapes and replaces the old ragged array-based storage. 

I also fixes the issue that only the geometry column was saved, now all the columns of the dataframe are saved.
Closes https://github.com/scverse/spatialdata/issues/311

Notes:
- No performance change is observed for writing (they are both binary formats). But GeoParquet improves interoperability. 
- The PR is backward compatible: all previous datasets can be read, and the user can choose if writing the shapes with the latest format or with a previous format. This choice can be performed per-element.
- Tests are added to ensure the compatibility between the shapes "v01" and "v02"

These last two points required a overhaul of format.py and related code, code that now is more capable and robust. A few notes on this:
- For raster types we didn't introduce a "spatialdata format" because we were optimistic that transformations etc would have landed in NGFF faster; since they are not there yet, I believe that we need versioning also for raster types (which I now implemented). 
- To avoid confusion between the `version` property, that always gives 0.4 and the spatialdata version, I renamed the `version` property to `spatialdata_format_version`. When writing points, shapes and tables, we need just the second one (because NGFF doesn't support them); instead, while writing labels and images we need both: `ome-zarr-py` needs to know that we are working with `0.4`, and spatialdata needs to know which spatial data format for raster we are using to augment the NGFF storage.
- I have introduced a function `_parse_version()` that retrieves the spatialdata zarr version for a given element from its Zarr store.
- The function `_parse_version()` less clean that I hoped, this because of the versioning was done in a heterogenous way for the elements:
	- Shapes and points saved and still save the version under `.attrs['spatialdata_attrs']['version']`
	- Tables saved the version under `.attrs['version']`, they still save the version there for backward compatibility even if switching to `.attrs['spatialdata_attrs']['version']` would be cleaner.
	- Images and labels didn't save the version at all, now they save it under `.attrs['spatialdata_attrs']['version']`. If `_parse_version()` can't find the version for raster types, it is assumed that the spatialdata `RasterFormatV01` is used.


LucaMarconato:
Ready for review, thanks 😊

LucaMarconato:
Todo:
- [ ] update changelog

LucaMarconato:
The docs started failing in a different PR, I'd address that separately.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1114,8 +1115,30 @@ def write(
...
+        ----------
+        file_path
+            The path to the Zarr store to write to.
+        overwrite
+            If `True`, overwrite the Zarr store if it already exists. If `False`, `write()` will fail if the Zarr store
+            already exists.
+        consolidate_metadata
+            If `True`, consolidate metadata (write all the metadata in a single file at the root directory of the
+            store). This optimizes the data to be cloud accessible, and it is even a requirement for certain cloud
+            stores (such with S3).
```
giovp, line 1133:
```suggestion
            If `True`, triggers :func:`zarr.convenience.consolidate_metadata`, which writes all the metadata in a single 			file at the root directory of the store. This makes the data cloud accessible, which is required for certain cloud
            stores (such as S3).
```

src/spatialdata/_core/spatialdata.py
```diff
@@ -1114,8 +1115,30 @@ def write(
...
+        format
+            The format to use for writing the elements of the `SpatialData` object. It is recommended to leave this
+            parameter equal to `None` (default to latest format for all the elements). If not `None`, it must be
+            either a format for an element (formats are defined in `spatialdata._io.format`), or a list of formats.
```
giovp, line 1137: shall we add `SpatialDataFormat` to the docstrings and create interlink here?

src/spatialdata/_io/format.py
```diff
@@ -69,12 +100,18 @@ def validate_coordinate_transformations(
...
+        return "0.1"
+
 
-class ShapesFormatV01(SpatialDataFormatV01):
+class ShapesFormatV01(SpatialDataFormat):
     """Formatter for shapes."""
 
     @property
-    def version(self) -> str:
+    def spatialdata_format_version(self) -> str:
```
giovp, line 114: I don't undersatnd the renaming to spatialdata_format_version, if it's really just the version of the element?

src/spatialdata/_io/format.py
```diff
@@ -149,6 +198,62 @@ def validate_table(
...
+        "shapes": CurrentShapesFormat(),
+        "points": CurrentPointsFormat(),
+        "tables": CurrentTablesFormat(),
+    }
+    if formats is None:
+        return parsed
+    if not isinstance(formats, list):
+        formats = [formats]
+
+    modified = {
```
giovp, line 232: what does this modified do?

src/spatialdata/_io/io_shapes.py
```diff
@@ -24,24 +29,34 @@
...
+        if typ.name == "POINT":
+            radius = np.array(f["radius"])
+            geometry = from_ragged_array(typ, coords)
+            geo_df = GeoDataFrame({"geometry": geometry, "radius": radius}, index=index)
+        else:
+            offsets_keys = [k for k in f if k.startswith("offset")]
+            offsets = tuple(np.array(f[k]).flatten() for k in offsets_keys)
+            geometry = from_ragged_array(typ, coords, offsets)
+            geo_df = GeoDataFrame({"geometry": geometry}, index=index)
+    elif isinstance(format, ShapesFormatV02):
```
giovp, line 53: really nice!

giovp (approved):
looks great! Much needed update, small comments

src/spatialdata/_io/format.py
LucaMarconato: The problem was that the various format classes (`PointsFormatV01`, ...) are inheriting from `ome_zarr.format.CurrentFormat`, so the property `format` was overriding the one from `ome_zarr`. This is ok for points, shapes and tables, but not for images and labels. In fact, since we use `ome_zarr` for some parts of IO for images and labels, we need the original `format`. But we need also the spatialdata `format`. So I thought that renaming the spatialdata format to not have this clash is cleaner.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Good point, I have updated the docs. More in a comment below.

LucaMarconato:
As you prompted, I have updated the docs. Now the API page from the docs has this in the end:
image

Please noticed that the first class is non-clickable (pink). This because we define the `SpatialDataFormat` class simply as 

```python
class SpatialDataFormat(CurrentFormat):
    pass
```

But if we had it to the docs, since we inherit from `ome_zarr.format.CurrentFormat`, the user would get this confusing page:
Screenshot 2024-07-22 at 18 37 02

src/spatialdata/_io/format.py
LucaMarconato: It is to check that no duplicate formats are passed by the user as the list of format used for writing. I added a comment.

src/spatialdata/_io/format.py
LucaMarconato: I think it's better to have these two property separate.

PR #544, opened 2024-04-11
Persist color channels in `rasterize()`

clwgg:
Fixes scverse/spatialdata-plot#249

Please see the above `spatialdata-plot` issue for a problem description.

This is just a draft of how this issue could be addressed. I could see several different routes this could be tackled, including:
- splitting the tests into a separate function
- using `if schema in (Image2DModel, Image3DModel):` logic instead of the current `try: except:`
- using `get_channels()` instead of `xdata.coords["c"].values`, although this is a bit awkward since `xdata` is not the correct type for `get_channels()`, so one would have to either modify `get_channels()` or call it on `data` instead of `xdata` which feels a little less clean

Let me know what you think!


giovp (approved):
thanks @clwgg , looks good! would you mind adding a release note here? https://github.com/scverse/spatialdata/blob/main/CHANGELOG.md thanks!

LucaMarconato:
Thanks @clwgg! I went for "using if schema in (Image2DModel, Image3DModel): logic instead of the current try: except:"; merging now 😊

Issue #432, opened 2024-01-16
Remove `pygeos` dependency


LucaMarconato:
As pointed out here https://github.com/ckmah/bento-tools/pull/104#discussion_r1454146578, we should be able to drop the `pygeos` dependency. We need to try it out.

Also this could be a solution to this behavior: https://github.com/scverse/spatialdata/issues/272.

leoschwarz:
Hi, thank you for your work on spatialdata. I think it would be great if pygeos could be removed as a dependency in the next release, because currently it prevents the installation with Python 3.12 since it uses some config parse related functionality that was removed from Python.

I haven't tested the full effects of removing it yet.

PR #545, opened 2024-04-16
drop unmaintained pygeos for shapely 2.0

omsai:
The requirement pygeos fails to install with python 3.12 with the error described here: https://github.com/pygeos/pygeos/issues/463

The maintainer explains that pygeos is no longer maintained and has been merged into the shapley 2.0 dependency of geopandas. Since version 0.14, geopandas uses shapley instead of pygeos: https://geopandas.org/en/stable/docs/user_guide/pygeos_to_shapely.html

Therefore, the PR drops the unmaintained pygeos and requires geopandas >= 0.14 to ensure the default shapley version is used instead.


clwgg:
(Issue tracked here:  #432)

giovp:
thanks @omsai for this PR, I wonder we could probably also remove this right https://github.com/scverse/spatialdata/blob/main/src/spatialdata/_compat.py#L1 ?

omsai:
@giovp Indeed; I've dropped that as well.

giovp:
thank you @omsai , I think it should be fine to not have it backward compatible, would like to have other devs chime in

giovp (approved):

LucaMarconato:
Thanks for the PR! I have also tested it, with Python 3.12, and all works as expected.

Minor comment, I have removed the line `os.environ["USE_PYGEOS"] = "0"` from `spatialdata/tests/conftest.py` (and from other scripts in the ecosystem) because now it'd not needed anymore. Some notebooks still have it, so if we see it let's remove it (but no rush, that line currently has no effect).

LucaMarconato:
codecov woke up after 3 months 😅

PR #560, opened 2024-05-15
Vectorize labels into polygons

quentinblampey:
Converting a 2D Labels object (multiscale or not) into a GeoDataFrame

### Notes
- vectorization is applied on chunks
- for each chunks, `scikit_image` is used to find contours (for each label, a cropped view is extracted, then vectorized, and then translated according to the bounding box coords)
- polygons are also translated according to the chunk coords
- labels on multiple chunks are merged via the `dissolve` method


src/spatialdata/_core/operations/vectorize.py
```diff
@@ -146,11 +150,86 @@ def to_polygons(
...
+    assert isinstance(transformations, dict)
+
+    return ShapesModel.parse(gdf, transformations=transformations.copy())
+
+
+def _region_props_to_polygons(region_props: RegionProperties) -> list[Polygon]:
+    mask = np.pad(region_props.image, 1)
+    contours = skimage.measure.find_contours(mask, 0.5)
+
+    polygons = [Polygon(contour[:, [1, 0]]) for contour in contours if contour.shape[0] >= 4]
```
LucaMarconato, line 209: I would add a comment explaining the choice of 4.

LucaMarconato:
Excellent PR! I just added a minor comment, but otherwise looks great! I will check/implement the following minor cases (copying them from the hackmd):

```
### small things
- [ ] to_circles() seems not to be implemented for points
- [ ] to_circles() seems not to be implemented for circles
- [ ] circles_to_polygons() should be more explicit: if values_geotypes == ['Point'], elif is.subset(['Polygon', 'MultiPolygon']) ... , else fail

### other small things
- [ ] remove `circles_to_polygons()` and use `to_polygons()` instead
- [ ] `to_polygons()` should work for polygons, multipolygons, circles and points (or maybe fail on points)
```

I will merge after I complete them.

src/spatialdata/_core/operations/vectorize.py
quentinblampey: I added a comment in the code.
When the number of vertices is 3, the shape is basically a line (2 points + repetition of the first point), which would raise a Shapely error. This is why I added this filtering.
Though, we should never have a shape with less than three coordinates: even a single pixel would be converted into a square. I just added this filtering for safety, but I'm not sure if skimage can produce a shape with less than 3 coordinates.

PR #562, opened 2024-05-22
updated deprecation_alias

melonora:
I have updated the `deprecation_alias` to be flexible enough for implementation across the ecosystems and to handle different arguments with deprecations in different versions.

Example
```
@deprecation_alias(table="tables", version="version 0.1")
def your_function(tables):
      do something...

your_function(table)
```
This will now give a message that table is set for deprecation from version 0.1 onwards. The message will also include the library so if a deprecation_alias decorator is applied in a function in spatialdata-plot then it will indicate the spatialdata-plot library in the message.

@timtreis probably useful for the spatialdata-plot refactor.


LucaMarconato (approved):
Very useful!

PR #565, opened 2024-05-24
improve data loader performance

giovp:
so I've been wanting to take another look at this for a long time, I used https://github.com/benfred/py-spy with `speedscope`  format, you can see screenshot below. 
image

I've been doing this on the `xenium_rep_1` dataset from the paper, and been using the following code (adapting  from @LucaMarconato code ):

<details>
<summary>Details</summary>

```python
import json

import numpy as np
import pandas as pd
import torchvision.transforms.v2 as T
from spatialdata.dataloader.datasets import ImageTilesDataset
from spatialdata.transformations import Scale, get_transformation
from spatialdata.transformations import Sequence as SequenceTransformation
from torch.utils.data import DataLoader
from tqdm import tqdm

from pathlib import Path
import spatialdata as sd

xeniumrep1 = Path(
  "/path/to/xenium_rep1_data_aligned.zarr"
)
sdata1 = sd.read_zarr(xeniumrep1)

visium = Path(
    "/path/to/visium_data_aligned.zarr"
)
sdata3 = sd.read_zarr(visium)

TILE_SCALE = 10.0
REGION = "xeniumrep1"
sdata = sdata1
sdata.images["hne"] = sdata3.images["CytAssist_FFPE_Human_Breast_Cancer_full_image"]

def get_ds(sdata: sd.SpatialData):
    img_size = 224

    transform_tv = T.Compose(
        [
            T.ToImage(),
            T.Resize((img_size, img_size), antialias=True, interpolation=T.InterpolationMode.BICUBIC),
            T.ToTensor(),
        ]
    )

    def transform(output):
        image, anno = output
        instance_id, celltype = anno[:, 0].squeeze(), anno[:, 1].squeeze()
        image = transform_tv(image.data.transpose(1, 2, 0).compute(scheduler="single-threaded"))
        out = {"img": image, "instance_id": instance_id.tolist(), "celltype": celltype.tolist()}
        return out

    mu = sdata.shapes["cell_circles"]["radius"].mean()
    std = sdata.shapes["cell_circles"]["radius"].std()
    # large radius to cover most of the cells
    large_radius = mu + 2 * std
    neighbors_contex = large_radius
    sdata.shapes["cell_circles"]["radius"] = neighbors_contex
    instance_key = sdata.tables["table"].uns["spatialdata_attrs"]["instance_key"]

    ds = ImageTilesDataset(
        sdata=sdata,
        regions_to_images={"cell_circles": "hne"},
        regions_to_coordinate_systems={"cell_circles": "aligned"},
        return_annotations=[instance_key, "celltype_major"],
        tile_scale=TILE_SCALE,
        transform=transform,
        table_name="table",
    )
    return ds

ds = get_ds(sdata)
dl = DataLoader(
    ds,
    batch_size=256,
    num_workers=0,
    shuffle=False,
)

```

</details>

this made me realize that, if we want to return the array, than there is an unnecessary step of instantiating the `SpatialImage|MultiscaleSpatialImage` that is not necessary, and the dask array could be simply returned. This halved the `fetch` step (across 6 iterations) from ~43s  to ~23s total, see below
image

I think the `fetch` step is what ultimately we want to improve, as it's the one that stream the tiles from the zarr array to the GPU. Now the two main blocks are the `transform` call and the `compute` call. The `transform` call is visualized under `compute` but it's effectively the `wrapper` call, where all the `DataArray.isel` happen, which is where the crops are defined, transformed and set, before the computation is actually triggered with `compute`.
image
I wonder what could be the next step here to chase performance gain: I think one option would be to basically "prepare" the transformation before on the full array, and then trigger it only at the tile creation in the `compute` (whereas now, transformation and tile creation is done jointly for each tile). This I think would require significant refactoring though so I wonder if it makes sense at all, and if anyone has other ideas to explore @scverse/spatialdata

### DIFF
--- a/src/spatialdata/dataloader/datasets.py
+++ b/src/spatialdata/dataloader/datasets.py
@@ -96,6 +96,9 @@ class ImageTilesDataset(Dataset):
         It is a `Callable`, with `Any` as return type, that takes as input the (image, table_value) tuple (when
         `return_annotations` is not `None`) or a `Callable` that takes as input the `SpatialData` object (when
         `return_annotations` is `None`).
+    return_array
+        If `True`, the tile is returned as an :class:`dask.array.Array` object; otherwise, it is returned as a
+        :class:`spatial_image.SpatialImage` object.
     rasterize_kwargs
         Keyword arguments passed to :func:`spatialdata.rasterize` if `rasterize` is `True`.
         This argument can be used in particular to choose the pixel dimension of the produced image tiles; please refer
@@ -122,6 +125,7 @@ class ImageTilesDataset(Dataset):
         return_annotations: str | list[str] | None = None,
         table_name: str | None = None,
         transform: Callable[[Any], Any] | None = None,
+        return_array: bool = False,
         rasterize_kwargs: Mapping[str, Any] = MappingProxyType({}),
     ):
         from spatialdata import bounding_box_query
@@ -144,9 +148,9 @@ class ImageTilesDataset(Dataset):
                 **dict(rasterize_kwargs),
             )
             if rasterize
-            else bounding_box_query  # type: ignore[assignment]
+            else partial(bounding_box_query, return_array=return_array)
         )
-        self._return = self._get_return(return_annotations, table_name)
+        self._return = self._get_return(return_annotations, table_name, return_array=return_array)
         self.transform = transform
 
     def _validate(
@@ -299,8 +303,10 @@ class ImageTilesDataset(Dataset):
         dataset_index: pd.DataFrame,
         table_name: str | None,
         return_annot: str | list[str] | None,
+        return_array: bool = False,
     ) -> tuple[Any, Any] | SpatialData:
-        tile = ImageTilesDataset._ensure_single_scale(tile)
+        if not return_array:
+            tile = ImageTilesDataset._ensure_single_scale(tile)
         if return_annot is not None:
             # table is always returned as array shape (1, len(return_annot))
             # where return_table can be a single column or a list of columns
@@ -334,6 +340,7 @@ class ImageTilesDataset(Dataset):
         self,
         return_annot: str | list[str] | None,
         table_name: str | None,
+        return_array: bool = False,
     ) -> Callable[[int, Any], tuple[Any, Any] | SpatialData]:
         """Get function to return values from the table of the dataset."""
         return partial(
@@ -342,6 +349,7 @@ class ImageTilesDataset(Dataset):
             dataset_index=self.dataset_index,
             table_name=table_name,
             return_annot=return_annot,
+            return_array=return_array,
         )
 
     def __len__(self) -> int:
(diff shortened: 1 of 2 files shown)


LucaMarconato:
Super cool analysis! I'll also try it out (which commands did you use to open `py-spy`? Or did you set it up to be integrated with your IDE?)

If most of the time is spent outside `dask_image.ndinterp.affine_transform()` (the core function used in `transform()`), then I think that preparing everything before and calling `affine_transform()` at the end would be a good approach.

But my bet (I need to check by running the profiler), is that the problem is that we load multiple times the same chunks. I think that maybe using `.persist()` to automatically cache some Dask chunks, and to order the cells so that we randomize the chunks first, and then the cells inside a chunk, would lead to performance improvements. 

This second approach has the advantage that it involves only the dataloader class and does not require changes in the transformation code.

LucaMarconato:
I reviewed the code, looks good to me. We could merge this already or explore first the `.persist()` approach above in this PR.

src/spatialdata/dataloader/datasets.py
```diff
@@ -81,6 +81,10 @@ class ImageTilesDataset(Dataset):
         system; this back-transforms the target tile into the pixel coordinates. If the back-transformed tile is not
         aligned with the pixel grid, the returned tile will correspond to the bounding box of the back-transformed tile
         (so that the returned tile is axis-aligned to the pixel grid).
+    return_genes:
```
LucaMarconato, line 84: Nice! Two comments: 
1. I would specify that the layers are `AnnData` layers and the default layer is `X`. 
2. I would also allow to pass just a list instead of a dict, that would be interpreted as `{'X': genes_list}`

giovp:
> I'll also try it out (which commands did you use to open py-spy? Or did you set it up to be integrated with your IDE?)

I've just changed the format in py-spy
`py-spy record --format speedscope -o profile.speedscope.json -- python process_xenium.py`

this was just a push to get the code in another machine. But let me explain what's next.

I've realized that the calculation of the transformed bounding box in the implicit coordinate system takes a fair amount of time and it could in fact be done only in the same way the tile coords dataset is built. I will therefore:
- move out the transformation from the bounding box query and do it only once at init. 
- Enable to return gexp data from different layers.

The dataset will have only type of output which will be dictionary of the following
```python
{
	"tile":tile,
	"annotations":list of annotations,
	"gexp": list of gexp,
}
```
wdyt?

What I won't do here but would be useful to work on next is:
- enable to work with multiple tables (doing something similar to https://docs.lamin.ai/lamindb.core.mappedcollection )
- enable to encode labels

LucaMarconato:
Thanks for the explanation. Yes, I think that operating on the transformation at the preprocessing stage is a good approach to improve performance. Also, the option to specify the layer will be useful.

Regarding the return type, would you remove the `SpatialData` return type or still leave it as an option?

giovp:
> Regarding the return type, would you remove the SpatialData return type or still leave it as an option?

that's a good question, I would potentially leave it but then technically the dataloader would fail as the default collate_fn only accepts array/mapping[str, array]/list[array], wdyt?

LucaMarconato:
Ok, then I would probably move the default away from returning `SpatialData` (but still leave it as an option to the users). I think a good default would be one compatible with the default `collate_fn`.

giovp:
close in favour of #687

PR #566, opened 2024-05-24
Rasterize shapes

quentinblampey:
Shapes rasterization

Use cases:
| values type | Shapes or Points | return_single_channel | datashader reduction |
|-------------|-----------------------------------|-----------------------|----------------------|
| None        | Point                             | NA                    | count                |
| None        | Shapes                              | True                  | first                |
| None        | Shapes                              | False                 | count_cat            |
| category    | NA                                | True                  | first                |
| category    | NA                                | False                 | count_cat            |
| int / float | NA                                | NA                    | sum                  |

Speed: 30s on 160k cells locally

Checklist:
- [x] Add some tests (rectangles with MapAxis transform)
- [x] Add notebook tutorial
- [x] Same for points
- [x] Add dict in attrs for the categories mapping


LucaMarconato:
Quick comments from discussion. Please add the following:
- [ ] the table of this PR in the docs. Also: 
- [ ] I would explain that count (first row) acts like having a column of ones; 
- [ ] count_cat could be paired with a text example

Comments on the code
- [ ] please add the test for points
- [ ] test for an anndata object annotating the shapes/points

LucaMarconato:
Luca's task before merging:
- [ ] implement the single dispatch against `SpatialData` objects
- [ ] labels should also accept table annotations

Tasks for a follow up PR:
- [ ] extend `value_key` to accept a list of strings, as we do in `aggregate()`
- [ ] implement the missing cases of `aggregate()` that now will defer the work to `rasterize()`
- [ ] use `rasterize()` in `spatialdata-plot` instead of reimplementing the usage of datashader

quentinblampey:
I removed `instance_key_as_default_value_key`, and updated the above table of use case. What do you think of the new table @LucaMarconato?

LucaMarconato:
Thanks looks great! I am working on the tasks listed in my two comments; I will tag you for a review of that part when I finish, thanks 😊

LucaMarconato:
> Luca's task before merging:
> 
> * [x]  implement the single dispatch against `SpatialData` objects
> * [x]  labels should also accept table annotations
> * [x]  testing the rasterization of shapes and points into labels when the index of the element has not int dtype but str
> * [x]  test when rasterizing an entire `SpatialData` object

@quentinblampey I have just finalized the implementation. Could you please give a last round of review to the new changes? Thanks!

quentinblampey (commented):
Looks good to me!
One small question: are we necessarily supposed to provide `element_name` if we want to get values from the table? Because I don't see how we can get the values if `element_name is None` (according to the current state of the code), which is not very convenient
(I actually have another small question but I can ask you tomorrow)

LucaMarconato:
Thanks for checking the code. Actually what you spotted is an idiosyncrasy in the APIs: the public API `rasterize()` does not contain the `element_name` parameter, since this is specified in `data` when `sdata` is `None`, while the internal APIs to rasterized images/labels and points/shapes always have `data` as a `SpatialElement` and, if provided `element_name` as `str`.

Therefore, if we want a value from a table, `sdata` must be specified and `data` must be the element name.

LucaMarconato:
> (I actually have another small question but I can ask you tomorrow)

Ok nice, then gonna merge now 😊

PR #578, opened 2024-06-11
Rasterize bins

quentinblampey:
Added API to rasterize grid-like bins data like Visium HD.


quentinblampey:
Remaining tasks:
- [ ] adding tests
- [ ] updating the notebook on Visium HD (?)

LucaMarconato:
I have extended `rasterize_bins()` to work also with points and I have added tests. This is the data that is being used in the tests: it's binned data where the data has been rotated by 30*, but also where an explicit transformation which scales the `x` axis 2-fold is used. 

CleanShot 2024-06-14 at 14 51 23

LucaMarconato:
The above example also shows that one in principle could use this function for Visium data. More details on this by explaining the way `rasterize_bins()` works.

`rasterize_bins()` assumes that each data instance (point/shape) can be indexed by a `(i_row, i_col)` pair. It will then construct an image where each pixel is indexed via `(i_row, i_col)` and finally assign to this image, the coordinate transformations already present in the original data.

But there is a further detail: it doesn't matter if the original data *pre-transformation* lies on a grid or not, just that the instances can be indexed via `(i_row, i_col)` (in the example in the test the data actually lies on a grid that is rotated by `pi/4`). After rasterizing the data, and before assign the coordinate transformations that were present in the data, the transformation that maps the indices `(i_row, i_col)` to the original data is computed and pre-pended to the coordinate transformation (in the example in the test such transformation is the rotation by `pi/4`, then the data has also an explicit coordinate transformation `y -> 2 * y`).

LucaMarconato:
To give a real example, consider this real Visium HD data. The data here doesn't have any explicit coordinate transformation (the transformation to transform the data is `Identity`); still the data lies on a grid that is very slightly rotated.
image

Rasterizing the data into a axis-aligned pixel grid would lead to an incorrect result. This is the reason why the `rasterize_bins()` function learns the tiny rotation and assigns it to the rasterized image before the original coordinate transformation (which in this case, as said, is just an `Identity()`).

quentinblampey:
Super cool work @LucaMarconato, looking amazing!🚀

LucaMarconato:
> Super cool work @LucaMarconato, looking amazing!🚀

Super nice teamwork! 🔥

Btw, ready for review, could you give a last pass to my edits before merging? Thanks 😊

LucaMarconato:
I will update the Visium HD notebook in a follow up PR in `spatialdata-notebooks`. For the moment, here are some preview of the features enabled with this PR on Visium HD data.

Gif 1: now for binned data (like Visium HD/Stereo-seq) one can choose between "squares/points" representation (vector) and "raster" representation. The raster representation respects the image alignment (affine transformations).

016 bins

Gif 2: the channel-wise rasterization is computed on the fly, here is shown how fast (2µm bins). In memory you have a standard `xarray.DataArray` object that you can treat like a standard image, but now each channel is computed on-demand.

002 bins

quentinblampey:
I had a look to your changes, it looks good to me :)

t-a-m-i:
Amazing work, really excited and looking forward to the updated notebook on the usage for Visium HD! 
When do you think rasterize_bins() will be part of a stable release? It works fine for me when I only install the dev version of spatialdata in my conda environment, but nothing works anymore as soon as I also download e.g. napari. (hope it´s okay to post this here)

LucaMarconato:
Hi, happy that you tried this out and that you find it useful 😊 It should be out this week!

PR #579, opened 2024-06-11
support AnnData table with get_values

melonora:
This PR allows to also pass an `AnnData` table as `element`. If `element_name` is `None` then a whole column is returned but if it is specified than the column is a subset only including those rows annotating the `element_name`. Docstrings have been updated to reflect this.

Furthermore, this PR also allows for getting values from `obsm`.

### DIFF
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -717,6 +717,8 @@ def _get_table_origins(
     # check if the value_key is in the var
     elif value_key in element.var_names:
         origins.append(_ValueOrigin(origin="var", is_categorical=False, value_key=value_key))
+    elif value_key in element.obsm:
+        origins.append(_ValueOrigin(origin="obsm", is_categorical=False, value_key=value_key))
     return origins
 
 
@@ -739,13 +741,6 @@ def _locate_value(
         origins.append(_ValueOrigin(origin="df", is_categorical=is_categorical, value_key=value_key))
     if model == TableModel:
         origins = _get_table_origins(element=el, value_key=value_key, origins=origins)
-        # if value_key in el.obs.columns:
-        #     value = el.obs[value_key]
-        #     is_categorical = pd.api.types.is_categorical_dtype(value)
-        #     origins.append(_ValueOrigin(origin="obs", is_categorical=is_categorical, value_key=value_key))
-        # # check if the value_key is in the var
-        # elif value_key in el.var_names:
-        #     origins.append(_ValueOrigin(origin="var", is_categorical=False, value_key=value_key))
 
     # adding from the obs columns or var
     if model in [ShapesModel, PointsModel, Labels2DModel, Labels3DModel] and sdata is not None:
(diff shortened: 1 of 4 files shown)


src/spatialdata/_core/query/relational_query.py
```diff
@@ -730,23 +751,19 @@ def _locate_value(
...
 def get_values(
     value_key: str | list[str],
-    element: SpatialElement | None = None,
+    element: SpatialElement | AnnData | None = None,
     sdata: SpatialData | None = None,
     element_name: str | None = None,
     table_name: str | None = None,
-) -> pd.DataFrame:
+    return_obsm_as_is: bool = False,
+) -> pd.DataFrame | npt.NDArray[Any]:
```
LucaMarconato, line 766: I will replace this with `ArrayLike` from `from spatialdata._types import ArrayLike` in the next commit.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -10,6 +10,7 @@
 
 import dask.array as da
 import numpy as np
+import numpy.typing as npt
```
LucaMarconato, line 13: I'm removing this in the next commit, using `from spatialdata._types import ArrayLike` instead.

tests/conftest.py
```diff
@@ -448,3 +460,58 @@ def _make_sdata_for_testing_querying_and_aggretation() -> SpatialData:
...
+                    "tissue_segmentation_scalef": 1,
+                }
+            }
+        },
+        "spatialdata_attrs": {"region": "test", "region_key": "region", "instance_key": "instance_id"},
+    }
+    obsm_labels = {
+        "spatial": rng.integers(0, blobs.shape[0], size=(n_obs_labels, 2)),
+        "spatial_copy": rng.integers(0, blobs.shape[0], size=(n_obs_labels, 2)),
+    }
```
LucaMarconato, line 516: I renamed `spatial` and `spatial_copy` to `tensor` and `tensor_copy` since `spatial` is from a legacy format so this may confuse the reader.

tests/conftest.py
```diff
@@ -448,3 +460,58 @@ def _make_sdata_for_testing_querying_and_aggretation() -> SpatialData:
...
+    uns_labels = {
+        "spatial": {
+            "labels": {
+                "scalefactors": {
+                    "spot_diameter_fullres": 10,
+                    "tissue_hires_scalef": 1,
+                    "tissue_segmentation_scalef": 1,
+                }
+            }
+        },
```
LucaMarconato, line 510: I deleted this because this shows legacy data that we don't want to reintroduce in the codebase.

tests/conftest.py
```diff
@@ -34,6 +36,16 @@
 
 RNG = default_rng(seed=0)
 
+try:
+    from numpy.typing import NDArray
+
+    NDArrayA = NDArray[Any]
+except (ImportError, TypeError):
+    NDArray = np.ndarray  # type: ignore[misc]
+    NDArrayA = np.ndarray  # type: ignore[misc]
```
LucaMarconato, line 45: I replaced this with `from spatialdata._types import ArrayLike`

LucaMarconato:
I reviewed and applied some (minor) code changes. Good work! Merging.

PR #582, opened 2024-06-12
add `get_element_instances`

giovp:


src/spatialdata/_core/centroids.py
```diff
@@ -98,9 +98,7 @@ def _(
 ) -> DaskDataFrame:
     """Get the centroids of a Labels element (2D or 3D)."""
     model = get_model(e)
-    if model in [Image2DModel, Image3DModel]:
-        raise ValueError("Cannot compute centroids for images.")
-    assert model in [Labels2DModel, Labels3DModel]
+    assert model in [Labels2DModel, Labels3DModel], "Expected a `Labels` element. Found an `Image` instead."
```
LucaMarconato, line 101: I replaced this with a `if ... raise` since the `assert` are not triggered during `python -O` (optimized mode).

giovp: sounds good, but then I would keep only the if raise instead of assert right? no need for double. Also, should we have if raise always?

LucaMarconato:
Looks great! I applied some minor changes, in particular I added the new function as a public API, I added it to the docs and I updated the changelog. Ready to merge!

PR #587, opened 2024-06-16
Support dataarray

### LINKED ISSUE #395: Replace SpatialImage and MultiscaleSpatialImage with DataArray and DataTree
The new releases of `spatial_image` and `multiscale_spatial_image` use these the base classes `DataArray` and `DataTree`. We need to:
- [ ] add support for these classes
- [ ] remove the version pin from the dependencies in `pyproject.toml`

Note. This modification will introduce a breaking change. I propose to develop this feature in a parallel branch that we release together with the rest of IO refactoring. CC @giovp @melonora @kevinyamauchi

melonora:
closes #395 
Supporting DataArray and DataTree


melonora:
So `SpatialImage` returns a `SpatialImage` while `to_spatial_image` returns a `DataArray`. Because of this there was the problem of reading data back in as here we use `SpatialImage`. I did a hacky way now to adjust the class, but not certain if there are better ways.

pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
     "pyarrow",
     "rich",
     "shapely>=2.0.1",
-    "spatial_image==0.3.0",
+    "spatial_image==1.1.0",
```
LucaMarconato, line 40: i'd make it >=

melonora: yeah had it locally, but did not push yet

src/spatialdata/models/_utils.py
```diff
@@ -7,15 +7,17 @@
...
 from multiscale_spatial_image import MultiscaleSpatialImage
 from shapely.geometry import MultiPolygon, Point, Polygon
 from spatial_image import SpatialImage
+from xarray import DataArray
 
 from spatialdata._logging import logger
 from spatialdata.transformations.transformations import BaseTransformation
 
-SpatialElement = Union[SpatialImage, MultiscaleSpatialImage, GeoDataFrame, DaskDataFrame]
+SpatialElement = Union[SpatialImage, DataArray, MultiscaleSpatialImage, DataTree, GeoDataFrame, DaskDataFrame]
```
LucaMarconato, line 20: Here we should remove `SpatialImage` and `MultiscaleSpatialImage` from the `SpatialElements`.

The idea is to use `to_multiscale()` internally, and in `RasterModel.parse()` call internally `to_spatial_image()` and `to_multiscale()` to trigger validation via the `spatial_image` and `multiscale_spatial_image` package, but the user would only be exposed to `DataArray` and `DataTree`.

melonora: yeah this indeed should still be removed

melonora:
For adding `DataTree` as a possible type to parse I would add a separate PR as I feel like it goes beyond the scope of this PR.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
     "dask<=2024.2.1",
     "fsspec<=2023.6",
     "geopandas>=0.14",
-    "multiscale_spatial_image==0.11.2",
```
LucaMarconato, line 31: I"d add >=

melonora: yeah it was like that, but was just checking what the issue is with pre CI.

tests/conftest.py
```diff
@@ -313,14 +313,13 @@ def sdata_blobs() -> SpatialData:
...
         sdata.shapes[k] = _deepcopy(v)
-    from spatialdata._utils import multiscale_spatial_image_from_data_tree
 
-    sdata.images["blobs_multiscale_image"] = multiscale_spatial_image_from_data_tree(
-        sdata.images["blobs_multiscale_image"]
-    )
-    sdata.labels["blobs_multiscale_labels"] = multiscale_spatial_image_from_data_tree(
-        sdata.labels["blobs_multiscale_labels"]
-    )
+    # sdata.images["blobs_multiscale_image"] = multiscale_spatial_image_from_data_tree(
```
LucaMarconato, line 317: you can remove the comments here

tests/utils/test_testing.py
```diff
@@ -73,15 +70,12 @@ def test_assert_elements_are_identical_metadata(full_sdata):
...
+            if not isinstance(copied[element_name], DataTree):
+                assert_elements_are_identical(full_sdata[element_name], copied[element_name])
             _change_metadata_image(copied, element_name, coords=True, transformations=True)
         elif get_model(element) in (Labels2DModel, Labels3DModel):
             _change_metadata_labels(copied, element_name)
-            # TODO: bug in data_tree, need to report
-            if not isinstance(copied[element_name], MultiscaleSpatialImage):
+            if not isinstance(copied[element_name], DataTree):
                 with pytest.raises(AssertionError):
                     assert_elements_are_identical(full_sdata[element_name], copied[element_name])
```
LucaMarconato, line 80: you removed one of the two `with.pytest.raises()`, I would bet that either both are needed or both are not. If both are needed is to silence a bug with copying that I had found in `data_tree` (I actually don't remember what the bug was, I should have written it in the comment). If you end up keeping the `with.pytest.raises()` please restore the comment to explain that the objects should be identical but a bug in `data_tree` makes the comparison fail.

tests/utils/test_testing.py
```diff
@@ -68,46 +48,13 @@ def test_assert_elements_are_identical_metadata(full_sdata):
 
     to_iter = list(copied.gen_elements())
     for _, element_name, element in to_iter:
-        if get_model(element) in (Image2DModel, Image3DModel):
```
LucaMarconato, line 73: Here I was testing if `assert_elements_are_identical` is able to compare not just the data but also the metadata. Removing them reduces the strength of the test. The code is quite ugly, but I'd rather simplify/keep it like this than remove it, since it's used only in the tests.

src/spatialdata/transformations/_utils.py
```diff
@@ -72,13 +71,13 @@ def _set_transformations(e: SpatialElement, transformations: MappingToCoordinate
     raise TypeError(f"Unsupported type: {type(e)}")
 
 
-@_get_transformations.register(SpatialImage)
-def _(e: SpatialImage) -> Optional[MappingToCoordinateSystem_t]:
+@_get_transformations.register(DataArray)
+def _(e: DataArray) -> Optional[MappingToCoordinateSystem_t]:
```
giovp, line 75: nitpicky but changing `Union` to `|` breaks singledispatch?

giovp (approved):
amazing @melonora super important!

tests/utils/test_testing.py
melonora: Added it back, but without duplication of asserts and now the tests are run first with elements being equal and then with elements not being equal.

src/spatialdata/transformations/_utils.py
melonora: Nitpicky change pushed. Also silenced some is_categorical_dtype deprecation warnings.

PR #591, opened 2024-06-18
Fix pre ci

melonora:
Pre ci was failing due to an issue with geopandas 1.0.0rc1 which doesn't support the current implementation of point geometry to polygon as it expects an array of geometry objects


melonora:
closed to have clean pr of branch from main

PR #610, opened 2024-07-02
Fix #580 region column to categorical

### LINKED ISSUE #580: `set_table_annotates_spatialelement` doesn't return a valid table
https://github.com/scverse/spatialdata/blob/9e07583f13a70be0855c94ce54b18f133f5ab5b8/src/spatialdata/_core/spatialdata.py#L432

using this, it returns a table that is not valid. Specifically, it fails at write time due to region not being categorical. I believe because the table is never actually parsed somehow, but important to check where does this check fail

cc. @lotte-vandevreken @lopollar

melonora:
closes #580 

When setting a new annotation target of the table, the region column was not checked to be a categorical so if the user didn't do this it would fail on write. This PR fixes that issue by adding a new function to check whether the column is categorical and converting it. This is used in two places, in the parsing method of the table model and in the function that sets the annotation target.

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -37,7 +37,7 @@ from spatialdata.models import (
     get_model,
     get_table_keys,
 )
-from spatialdata.models._utils import SpatialElement, get_axes_names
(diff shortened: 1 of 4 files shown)


giovp (approved):
thanks @melonora ! looks great!

PR #615, opened 2024-07-04
Update dask dependency

melonora:
The current dask dependency constraint can cause issues when using python 3.11 as reported here: https://github.com/dask/dask/issues/11038. This PR updates the dependency to the version in which the fix has been implemented.


LucaMarconato:
Well done @melonora!

@ilan-gold this should fix your issue.

PR #617, opened 2024-07-05
refactor bounding box

giovp:
I would need this refactor in order to move on with the data loader performance PR. @LucaMarconato what this PR basically does is unifying the building request code for `DataArray|DataTree` with the already existing `_get_bounding_box_corners_in_intrinsic_coordinates` for `DaskDataFrame|GeoPandasDataFrame`. I *think* this is correct, as in the tests for `spatial_query` pass, but I am not sure that the logic is correct.

I had to disable the tests for the dataloaders but they will be worked out in a separate PR. This PR is ready to review, but I would not merge it for now, and wait for finishing the dataloader improvement work.

### DIFF
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -98,10 +98,6 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
     if set(axes) != set(output_axes_without_c):
         raise ValueError("The axes of the bounding box must match the axes of the transformation.")
 
-    # in the case of non-raster data, we need to swap the axes
-    if not isinstance(element, (DataArray, DataTree)):
-        input_axes_without_c, axes = axes, input_axes_without_c
-
     # let's get the bounding box corners and inverse transform then to the intrinsic coordinate system; since we are
     # in case 1 or 5, the transformation is invertible
     spatial_transform_bb_axes = Affine(
@@ -252,6 +248,11 @@ def _adjust_bounding_box_to_real_axes(
             M = np.finfo(np.float32).max - 1
             min_coordinate = np.append(min_coordinate, -M)
             max_coordinate = np.append(max_coordinate, M)
+    else:
+        indices = [axes_bb.index(ax) for ax in axes_out_without_c]
+        min_coordinate = min_coordinate[np.array(indices)]
+        max_coordinate = max_coordinate[np.array(indices)]
+        axes_bb = axes_out_without_c
     return axes_bb, min_coordinate, max_coordinate
 
 
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -432,6 +432,7 @@ def bounding_box_query(
     min_coordinate: list[Number] | ArrayLike,
     max_coordinate: list[Number] | ArrayLike,
     target_coordinate_system: str,
+    return_request_only: bool = False,
     filter_table: bool = True,
     **kwargs: Any,
 ) -> SpatialElement | SpatialData | None:
@@ -451,6 +452,9 @@ def bounding_box_query(
     filter_table
         If `True`, the table is filtered to only contain rows that are annotating regions
         contained within the bounding box.
+    return_request_only
+        If `True`, the function returns the bounding box coordinates in the target coordinate system.
+        Only valid with `DataArray` and `DataTree` elements.
 
     Returns
     -------
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -501,7 +501,7 @@ def _(
     max_coordinate: list[Number] | ArrayLike,
     target_coordinate_system: str,
     return_request_only: bool = False,
-) -> DataArray | DataTree | Mapping[str, list[int]] | None:
+) -> DataArray | DataTree | Mapping[str, slice] | None:
     """Implement bounding box query for Spatialdata supported DataArray.
 
     Notes
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -834,6 +834,7 @@ def _(
     image: DataArray | DataTree,
     polygon: Polygon | MultiPolygon,
     target_coordinate_system: str,
+    return_request_only: bool = False,
     **kwargs: Any,
 ) -> DataArray | DataTree | None:
     _check_deprecated_kwargs(kwargs)
@@ -845,6 +846,7 @@ def _(
         max_coordinate=[max_x, max_y],
         axes=("x", "y"),
         target_coordinate_system=target_coordinate_system,
+        return_request_only=return_request_only,
     )
 
 
--- a/src/spatialdata/_core/query/spatial_query.py
+++ b/src/spatialdata/_core/query/spatial_query.py
@@ -50,7 +50,7 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
     min_coordinate: list[Number] | ArrayLike,
     max_coordinate: list[Number] | ArrayLike,
     target_coordinate_system: str,
-) -> tuple[ArrayLike | DataArray, tuple[str, ...]]:
+) -> tuple[DataArray, tuple[str, ...]]:
     """Get all corners of a bounding box in the intrinsic coordinates of an element.
 
     Parameters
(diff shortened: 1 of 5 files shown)


src/spatialdata/_core/query/spatial_query.py
```diff
@@ -74,38 +75,65 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
         element, target_coordinate_system
     )
+    spatial_transform = Affine(m_without_c, input_axes=input_axes_without_c, output_axes=output_axes_without_c)
+
+    # we identified 5 cases (see the responsible function for details), cases 1 and 5 correspond to invertible
+    # transformations; we focus on them
+    if isinstance(element, (DataArray, DataTree)):
+        m_without_c_linear = m_without_c[:-1, :-1]
+        case = _get_case_of_bounding_box_query(m_without_c_linear, input_axes_without_c, output_axes_without_c)
+        assert case in [1, 5]
```
LucaMarconato, line 93: we should replace this `assert` with `if ... raise`, so that this is triggered even when running `python` with `-O`

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -74,38 +75,65 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+
+    # adjust the bounding box to the real axes, dropping or adding eventually mismatching axes; the order of the axes is
+    # not adjusted
     axes, min_coordinate, max_coordinate = _adjust_bounding_box_to_real_axes(
         axes, min_coordinate, max_coordinate, output_axes_without_c
     )
+    assert set(axes) == set(output_axes_without_c)
```
LucaMarconato, line 100: same for the assert here and in other parts of the code.

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: (I am the one to (git) blame for having used the asserts in the first place 😅)

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -74,38 +75,65 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+    # compute the output axes of the transformation, remove c from input and output axes, return the matrix without c
+    # and then build an affine transformation from that
     m_without_c, input_axes_without_c, output_axes_without_c = _get_axes_of_tranformation(
         element, target_coordinate_system
     )
+    spatial_transform = Affine(m_without_c, input_axes=input_axes_without_c, output_axes=output_axes_without_c)
+
+    # we identified 5 cases (see the responsible function for details), cases 1 and 5 correspond to invertible
+    # transformations; we focus on them
+    if isinstance(element, (DataArray, DataTree)):
```
LucaMarconato, line 90: we actually need to invert the matrix also when the elment is not raster, so I would drop this if and always make the check.

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -49,7 +50,7 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
     min_coordinate: list[Number] | ArrayLike,
     max_coordinate: list[Number] | ArrayLike,
     target_coordinate_system: str,
-) -> tuple[ArrayLike, tuple[str, ...]]:
+) -> tuple[ArrayLike | DataArray, tuple[str, ...]]:
```
LucaMarconato, line 53: to make the code a bit easier I would always return a `DataArray` for all the types of elements, and adjust the code using this function to deal with the returned `DataArray` instead of the returned `ArrayLike`.

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -74,38 +75,65 @@ def _get_bounding_box_corners_in_intrinsic_coordinates(
...
+    # adjust the bounding box to the real axes, dropping or adding eventually mismatching axes; the order of the axes is
+    # not adjusted
     axes, min_coordinate, max_coordinate = _adjust_bounding_box_to_real_axes(
         axes, min_coordinate, max_coordinate, output_axes_without_c
     )
+    assert set(axes) == set(output_axes_without_c)
+
+    # in the case of non-raster data, we need to swap the axes
+    if not isinstance(element, (DataArray, DataTree)):
+        input_axes_without_c, axes = axes, input_axes_without_c
```
LucaMarconato, line 104: are you sure? I think before we were using this variable after calling inverse, and now we first construct the affine and then call inverse. So I don't think it's a case "raster" vs "non-raster", and the same code should work for both.

LucaMarconato (commented):
Great PR! It simplifies the code and will make the torch dataset class more performant!

I added a few comments of changes I would make. They are all minor except for one that may require more care (the one where `axes` and `output_axes_without_c` are swapped). I can also have a deeper look at that case if you want.

LucaMarconato (changes requested):
Great PR! It simplifies the code and will make the torch dataset class more performant!

I added a few comments of changes I would make. They are all minor except for one that may require more care (the one where `axes` and `output_axes_without_c` are swapped). I can also have a deeper look at that case if you want.

src/spatialdata/_core/query/spatial_query.py
giovp: ok, I actually noticed that it is raised already as an error here so I would remove this check entirely https://github.com/scverse/spatialdata/blob/cc85a7fca124bfb205021cee33b275419fc0fae0/src/spatialdata/_core/query/spatial_query.py#L308-L316

src/spatialdata/_core/query/spatial_query.py
giovp: ok!

src/spatialdata/_core/query/spatial_query.py
giovp: Done

src/spatialdata/_core/query/spatial_query.py
giovp: yeah I am not 100% sure but if I don't swap the axes, this test fails
```
FAILED tests/core/query/test_spatial_query.py::test_query_spatial_data - AssertionError: DataFrame are different
```

giovp: I think it has to do more with the fact that I have to unify the variables, and that I am now using the same variables for raster data, also for non-raster data, you can see how this function is used in the non-raster case before this PR. There maybe be a better way that doesn't require this but then the input/output axes of the transformation in the non raster case would have to be changed.

src/spatialdata/_core/query/spatial_query.py
giovp: Done

giovp:
Hi @LucaMarconato , I've addressed all comments, except this one

> one that may require more care (the one where axes and output_axes_without_c are swapped)

If I don't swap the axes, there is one test that fails (see review comment). I think it might be solvable by changing the input/output axes of`spatial_transform_bb_axes` but I am not sure the best solution there. The fact that it's also only that test for spatialdata, and not the e.g. `test_query_points` also makes it a bit suspicious. Would really appreciate your help!

giovp:
one more question, is `return_request_only` a good argument name? and should it be in all cases or only raster? I would argue only raster because I don't see how much value it has for the shapes/points element case. In fact, the only use case so far is for the dataloader.

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: Thanks for the info. I'll also have a look at the code.

LucaMarconato:
@giovp I checked the code and the issue was a bug in the function `_adjust_bounding_box_to_real_axes()`, which was not adjust the bounding box to the real axes in the case in which the bounding box axes and the real axes were a permutation of each other. I have addressed that and now the swap between the two variables for non-raster types is removed.

Before the swap was fixing the bug because of a coincidence. I will add a targeted test for this case, and after that we can merge.

giovp:
> Before the swap was fixing the bug because of a coincidence. I will add a targeted test for this case, and after that we can merge.

🙏 🙏

src/spatialdata/_core/query/spatial_query.py
LucaMarconato: Fixed, I explained this in the conversation.

LucaMarconato:
I have finished adding tests, good to merge 😊

LucaMarconato (approved):

PR #619, opened 2024-07-06
Fix filter tables

melonora:
There was a problem causing napari-spatialdata to fail in case someone has a SpatialData object with a table that does not annotate an element. This was due to filtering the spatialdata object by coordinate system. In case `include_orphan_tables` was `False` for filtering ultimately `None` was attempted to be assigned as a table in the SpatialData object. This PR adds a conditional to prevent this from happening and also adds a test.


LucaMarconato (approved):
Looks great to me, good fix!

PR #621, opened 2024-07-08
add argument `return_background` for `get_centroids` and `get_element_instances` in the case of labels

### LINKED ISSUE #614: `get_element_instances` and `get_centroids` should not (?) return zero element
https://github.com/scverse/spatialdata/blob/f94d0c7df823389076f8bb097c1d887fe643598e/src/spatialdata/_core/query/relational_query.py#L112-L114

I think that is counter-intuitive to return zero, since it's not an instance. I would maybe consider to add an optional argument `return_zero: bool = False` wdyt?

giovp:
closes #614

### DIFF
--- a/src/spatialdata/_core/centroids.py
+++ b/src/spatialdata/_core/centroids.py
@@ -69,7 +69,7 @@ def _get_centroids_for_axis(xdata: xr.DataArray, axis: str) -> pd.DataFrame:
     -------
     pd.DataFrame
         A DataFrame containing one column, named after "axis", with the centroids of the labels along that axis.
-        The index of the DataFrame is the collection of label values, sorted ascendingly.
+        The index of the DataFrame is the collection of label values, sorted in ascending order.
     """
     centroids: dict[int, float] = defaultdict(float)
     for i in xdata[axis]:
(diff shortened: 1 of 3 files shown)


LucaMarconato (approved):

tests/core/query/test_relational_query.py
```diff
@@ -699,14 +699,15 @@ def test_labels_table_joins(full_sdata):
         table_name="table",
         how="left",
     )
-    assert all(table.obs["instance_id"] == range(100))
+
+    assert all(table.obs["instance_id"] == range(1, 100))
```
giovp, line 703: here the labels has values 0,100 but the 0 element is background, so is effectively filtered in the left/right joins when `get_element_instances` is called on labels. Hence the indices of the instances are different.

giovp:
@melonora here I would also need your green light since, due to the change of `get_element_instances`, the results of the tests had to be modified. In short, because now the `get_element_instances` doesn't return the background as default, hence the joins with the table produce a different results. Please see inline comments above, let me know if anything is unclear and if you agree with this.
I guess one option would be to expose the `return_background` in join, but I really don't see it to be of any value at the moment. Wdyt?

melonora (approved):
Hi @giovp, LGTM! However as you said I agree that it does not make sense to have it in the join, but I would add it in the extra explanation of the docstring for the function before merging, but for now I pre-approve. Did you already check whether any tests on the other repos fail?

giovp:
@melonora indeed some tests in `spatialdata-plot` fails, so I will open a separate PR before merging this

PR #623, opened 2024-07-08
allow for uint

melonora:
Segmentation models can return `uint`. This wasn't supported. This PR adds the `uint` dtypes to be accepted when validating the table annotation metadata.

### DIFF
--- a/src/spatialdata/models/models.py
+++ b/src/spatialdata/models/models.py
@@ -853,12 +853,19 @@ class TableModel:
             raise ValueError(f"`{attr[self.REGION_KEY_KEY]}` not found in `adata.obs`. Please create the column.")
         if attr[self.INSTANCE_KEY] not in data.obs:
             raise ValueError(f"`{attr[self.INSTANCE_KEY]}` not found in `adata.obs`. Please create the column.")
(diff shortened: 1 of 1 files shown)


LucaMarconato:
@ilan-gold is there any implications with regards to JS compatibility? I think some of these types were not supported in JS.

melonora:
ah yeah ofcourse, ok so JS does not have built in support but they do have typed  ArrayBuffer. If this is not an option we need to add it to the docs that returned arrays need to be of this dtype for the reason of JS support as segmentation models do provide uint32.

melonora:
https://stackoverflow.com/questions/6041124/javascript-typed-arrays-64-bit-integers

ilan-gold:
The problem would not be on the CPU side but rather the GPU side where 64 bit anything is not supported.  We'd have to downcast to 32 bit integers.  `uint` in general is fine as that is indeed supported up to 32 bits.  I will say, you probably should add a warning about 64 bit integers.  If someone has more than 2**32 - 1 segmentations, I don't know what to say....

melonora:
OK I will open a PR to adjust:) Thanks for the info!

Issue #556, opened 2024-05-01
Subset spatialdata by list of cell ids


RAC1924:
Hi, 

Thank you for the spatialdata package. I'm just wondering if it's possible to subset our spatialdata object directly by a list of cell ids? I couldn't find it in the documentation

BW, 
Ros

aeisenbarth:
The method [SpatialData.subset](https://spatialdata.scverse.org/en/latest/generated/spatialdata.SpatialData.html#spatialdata.SpatialData.subset) takes only element names.

But you can further filter the returned object:
```python
from collections.abc import Iterable
from typing import Optional
import numpy as np


def filter_spatialdata(
    sdata: "SpatialData",
    elements: Optional[Iterable[str]] = None,
    regions: Optional[Iterable[str]] = None,
    obs_keys: Optional[Iterable[str]] = None,
    var_keys: Optional[Iterable[str]] = None,
    var_names: Optional[Iterable[str]] = None,
    layers: Optional[Iterable[str]] = None,
    region_key: str = "region",
    instance_key: str = "instance_id",
) -> "SpatialData":
    """
    Filter a SpatialData object to contain only specified elements or table entries.

    Args:
        sdata: A SpatialData object
        elements: Names of elements to include. Defaults to [].
        regions: Regions to include in the table. Defaults to regions of all selected elements.
        obs_keys: Names of obs columns to include. Defaults to [].
        var_keys: Names of var columns to include. Defaults to [].
        var_names: Names of variables (X columns) to include. Defaults to [].
        layers: Names of X layers to include. Defaults to [].

    Returns:
        A new SpatialData instance
    """
    from anndata import AnnData
    from spatialdata import SpatialData
    from spatialdata.models import TableModel
    from spatialdata_plot.pp import PreprocessingAccessor

    elements = [] if elements is None else list(elements)

    sdata.pp: PreprocessingAccessor  # noqa: F401
    sdata_subset = (
        sdata.subset(element_names=elements, filter_tables=True) if elements else SpatialData()
    )
    # Ensure the returned SpatialData is not backed to the original reference dataset,
    # so that it can be safely modified.
    assert not sdata_subset.is_backed()
    # Further filtering on the table
    if (table := sdata_subset.tables.get("table")) is not None:
        regions = elements if regions is None else regions
        obs_keys = [] if obs_keys is None else list(obs_keys)
        if instance_key not in obs_keys:
            obs_keys.insert(0, instance_key)
        if region_key not in obs_keys:
[...]

Issue #284, opened 2023-06-05
Feature request: spatial cropping from select table rows


LucaMarconato:
From @EliHei2:

> do we have reverse spatial queries? i.e., can one select some rows from the table and get the cropped image? easier case rectangular crops and more advanced exact curvatures?

I am currently working on similar things, so I adding this to my todo list.

sinanugur:
Hi,

Is this feature available?

I want to visualize the whole image where the spots are available but I am getting the visualization of the entire image. I have to create a boundind_box manually. Is there a way to only show where the spots are?

Cheers,
Sinan

LucaMarconato:
Hi, for shapes and points you could use the new API (not available last year) `join_spatialelement_table()` with `how='right'` to return a `SpatialElement` that contains only the geometries corresponding to the table. 

If you are using labels, `how='right'` is not available, but with the new `to_polygons()` API you can easily convert the labels to polygons/multipolygons.

Finally, after you have the polygons that match the rows in one of the two ways described above, you can use `get_extent()` (also new) to compute the bounding box of the geometries matching to the rows. You can use this to crop eventual images or even the whole `SpatialData` object.

Please let me know if it fits your use cases and if you have further questions.

Issue #280, opened 2023-05-31
Filter spatialData 


lopollar:
Idea for enhancement: 
Add a filter option to the spatialdata object, where it is possible to filter out certain cells, based on a column in the table. 
This can be relevant to for example large cells. 
In line with the `sc.pp.filter_cells`. 
A similar function exists in `spatialdata`: `sdata.pp.get_elements`, but this does not work on obs or shapes information. 
This would then work on shapes and table simultaneously. 
Currently, I only get this to work by overwriting the objects, as shown here.
Is there a better way? 
```python
for i in [*sdata.shapes]:
    sdata[i].index = sdata[i].index.astype("str")
    sdata.add_shapes(
        name=i,
        shapes=spatialdata.models.ShapesModel.parse(
            sdata[i][np.isin(sdata[i].index.values, sdata.table.obs.index.values)]
        ),
        overwrite=True,
    )
```

giovp:
this is very much in the roadmap @lopollar , thanks for bringing it up! 

I think this could be done in two different ways (both could be supported)

1)
```python
sdata = sdata[sdata.table.obs.celltype == "celltypeA"]
>>> sdata.table.obs.celltype.unique()
["celltypeA"]
```

2)
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html

wdyt?

lopollar:
Hi, 
For me, the first feels like how I wouldcode it myself, so I would prefer that one! 
Thank you for taking my suggestion into account!

PR #620, opened 2024-07-06
Draft SpatialData.filter()

LucaMarconato:
Closes #280 
Closes #284 
Closes #556

PR #626, opened 2024-07-08
Draft SpatialData.filter()

aeisenbarth:
(In reference to https://github.com/scverse/spatialdata/pull/620)

This PR imlements an more advanced filtering options than `subset`, allowing to create a new SpatialData object that contains only specific tables, layers, obs keys, var keys.

**Use cases**

- From a concatenated SpatialData, one can extract parts of it.
- When testing an operation that adds elements or table columns, one can extract from an expected reference dataset the input data and pass it to the operation, then compare the processed data against the reference.
- …

Closes https://github.com/scverse/spatialdata/issues/280
Closes https://github.com/scverse/spatialdata/issues/284
Closes https://github.com/scverse/spatialdata/issues/556

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning][].
 
 ## [0.x.x] - 2024-xx-xx
 
+-   Added `SpatialData.filter()` method for subsetting by `obs` and `var` @aeisenbarth
+
 ## [0.2.1] - 2024-07-04
 
 ### Minor
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import hashlib
 import os
 import warnings
-from collections.abc import Generator
+from collections.abc import Generator, Iterable
 from itertools import chain
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal
 
+import numpy as np
 import pandas as pd
 import zarr
 from anndata import AnnData
@@ -2062,6 +2063,80 @@ class SpatialData:
             elements_dict=elements_dict,
         )
         return SpatialData(**elements_dict, tables=tables)
+
+    def filter(
+        self,
+        elements: Iterable[str] | None = None,
+        regions: Iterable[str] | None = None,
+        tables: Iterable[str] | None = None,
+        obs_keys: Iterable[str] | None = None,
+        var_keys: Iterable[str] | None = None,
+        var_names: Iterable[str] | None = None,
+        layers: Iterable[str] | None = None,
+    ) -> SpatialData:
+        """
+        Filter a SpatialData object to contain only specified elements or table entries.
+
+        Parameters
+        ----------
+            elements: Names of elements to include. Defaults to [].
+            regions: Regions to include in the table. Defaults to regions of all selected elements.
+            tables: Names of tables to include. Defaults to ["table"].
+            obs_keys: Names of obs columns to include. Defaults to [].
+            var_keys: Names of var columns to include. Defaults to [].
+            var_names: Names of variables (X columns) to include. Defaults to [].
+            layers: Names of X layers to include. Defaults to [].
+
+        Returns
+        -------
+            A new SpatialData object
+        """
+        elements = [] if elements is None else list(elements)
+        regions = elements if regions is None else regions
+        obs_keys = [] if obs_keys is None else obs_keys
+        var_keys = [] if var_keys is None else var_keys
+        var_names = [] if var_names is None else list(var_names)  # iterable and sized
+        tables = ["table"] if tables is None else tables
+        layers = [] if layers is None else layers
+
+        sdata_subset = self.subset(element_names=elements, filter_tables=True) if elements else SpatialData()
+        # We rely on `subset` returning an unbacked copy, so we don't modifying the original data.
+        assert not sdata_subset.is_backed()
+        # Further filtering on the tables
+        for table_name, table in list(sdata_subset.tables.items()):
+            if table_name not in tables:
+                del sdata_subset.tables[table_name]
+                continue
+            _, region_key, instance_key = get_table_keys(table)
+            obs_keys = list(obs_keys)
+            if instance_key not in obs_keys:
+                obs_keys.insert(0, instance_key)
+            if region_key not in obs_keys:
+                obs_keys.insert(0, region_key)
+            # Preserve order by checking "isin" instead of slicing. Also guarantees no duplicates.
+            table_subset = table[
+                table.obs[region_key].isin(regions),
+                table.var_names.isin(var_names),
+            ]
+            layers_subset = (
+                {key: layer for key, layer in table_subset.layers.items() if key in layers}
+                if table_subset.layers is not None and len(var_names) > 0
+                else None
+            )
+            table_subset = TableModel.parse(
+                AnnData(
+                    X=table_subset.X if len(var_names) > 0 else None,
+                    obs=table_subset.obs.loc[:, table_subset.obs.columns.isin(obs_keys)],
+                    var=table_subset.var.loc[:, table_subset.var.columns.isin(var_keys)],
+                    layers=layers_subset,
+                ),
+                region_key=region_key,
+                instance_key=instance_key,
+                region=np.unique(table_subset.obs[region_key]).tolist(),
+            )
+            del sdata_subset.tables[table_name]
+            sdata_subset.tables[table_name] = table_subset
+        return sdata_subset
 
     def __getitem__(self, item: str) -> SpatialElement:
         """


aeisenbarth:
In the current state, it does not yet complete the issues that were aimed to resolve.
- https://github.com/scverse/spatialdata/issues/556:
  - A parameter `instances` could be added. If provided, rows of these instances will be selected in the table, if not provided, all instances are returned. But it should only allow a single region/element (otherwise it gets complicated).
  - Shapes/points elements should also be filtered (easy)
  - Still unanswered is what effect it should have on the labels. Shall we create a new labels image with the instances eliminated (0 = background), or leave the labels unchanged and just have no reference to them in the table?
- https://github.com/scverse/spatialdata/issues/280:
  - This involves a condition. In my opinion, implementing a parameter with a condition as function or [query expression](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html) is out of scope due to its complexity. I would do this in two steps, users use Pandas to get a list of instances, then pass the instances to the SpatialData filter function.
  - The user also asked about adjusting shapes to match the filtered instances in the table.
- https://github.com/scverse/spatialdata/issues/284:
  - Filtering labels/shapes/points elements

LucaMarconato:
Thanks @aeisenbarth, after discussing with @melonora, we are going to first turn the code https://github.com/scverse/spatialdata/pull/627 into an internal function, merge, and then continue working on your PR. The idea is to provide a single entry point for filtering `filter()` and use for instance `subset()` or the function from Wouter internally.

owenwilkins:
is there somewhere is the domentation that now describes how to filter a spatialdata object by cell IDs? this is valuable for several reasons, e.g. filtering cells removed by QC in analysis using other libraries

LucaMarconato:
I went back to this and to https://github.com/scverse/spatialdata/pull/627 today and realized that we maybe do not need to add a new API, since all the points covered by this PR and by the linked PR, including all the points listed in this message here: https://github.com/scverse/spatialdata/pull/626#issuecomment-2215009660 are essentially covered by the example below, which uses the currently available APIs:


```python
##
# constructing the example data
from spatialdata.datasets import blobs_annotating_element
from spatialdata import concatenate
from spatialdata import join_spatialelement_table

sdata1 = blobs_annotating_element("blobs_polygons")
sdata2 = blobs_annotating_element("blobs_polygons")

sdata = concatenate({"sdata1": sdata1, "sdata2": sdata2}, concatenate_tables=True)
print(sdata)

##
# filtering the data
table_name = "table"
filtered_table = sdata[table_name][sdata[table_name].obs.instance_id < 3]
annotated_regions = sdata.get_annotated_regions(sdata[table_name])
elements, table = join_spatialelement_table(
    sdata, spatial_element_names=annotated_regions, table=filtered_table, how="inner"
)
sdata_filtered = SpatialData.init_from_elements(elements | {table_name: table})
print(sdata_filtered)
```

Explicitly, the code above first filters the table with standard `pandas`/`anndata` operations (and thus is very general), and then reuses the join operations to filter the `SpatialData` object (which again are very general and allow for several cases). In doing this:
- we limit code redundancy
- we can use any query/condition to filter the object, including:
	- we can filter by a threshold
	- we can filter by instances
	- we can manually subset certain obs/var/layers of the table
- shapes and points are filtered to match the table
- if a table annotates multiple elements, they are all filtered
- filtering labels is delegated to the join operations (currently not supported, but if in the future it will, it would be also supported here)

I think we could proceed as follow:
- we do not add any new API, and put the example above visible in the docs
- we add a very minimalistic API that essentially reproduces the example above. So convenient for most cases, and for more general cases the user can modify the code
- we add a feature complete API (similar to this PR); we use the code above internally so we don't have code redundancy.

aeisenbarth:
This PR indeed overlaps with existing APIs.

I agree that redundancy should be avoided. But I think an API should help to minimize gluing code. The above example solves the task, but contains two filtering steps and two intermediate function calls.

So from my side, I would favor extending the existing API to be more feature-complete. But I would not see it high priority.

LucaMarconato:
> I agree that redundancy should be avoided. But I think an API should help to minimize gluing code. The above example solves the task, but contains two filtering steps and two intermediate function calls.

You are right, that approach is not ergonomic enough. I thought about this and now in the linked PR I introduce a new API `match_sdata_to_table()` that bundles the code (@aeisenbarth could you please review the PR?).

All the use cases mentioned in your message are included in the [tests](https://github.com/melonora/spatialdata/blob/filter/tests/core/query/test_relational_query_match_sdata_to_table.py). I wonder if now the function is ergonomic enough or if we should still add a `filter()` function.

Here are my thoughts on this:

**Ergonomics limitation of the `match_sdata_to_table()`**
- you need to import the function and pass the `sdata` argument to it; a method in `SpatialData` could be added to fix this.
- while its usage is powerful and simple, as the tests linked above show, it may not be immediate how to use it without looking at the tests. We could add some examples in the docstring for this.
- the name tells exactly what the function does, but it's not what people would look for when trying to subset a `SpatialData` object by a threshold on `obs`/`var`.

**Limitations of the `SpatialData.filter()`**
- filter may suggest a threshold-based filtering, which we probably do not want to implement as the same result can be achieved via `pandas`/`anndata` APIs + `match_sdata_to_table()`, as shown in this [test ](https://github.com/melonora/spatialdata/blob/d79e5335c60d7f55d1e991a0f3be4c7893fe3be9/tests/core/query/test_relational_query_match_sdata_to_table.py#L56).
- code redundancy

## my preferred take
I think my preferred approach would be to gather feedback from the users, and then actually provide a new function (`filter()`? `query()`?) that simplifies the usage. This is in line to what reported here: https://github.com/scverse/spatialdata/issues/225. I'd strategize this by closing the current PR and reworking the code when we are ready into a new function, that eventually covers also the bounding box and polygon query cases. But happy to hear what you think @aeisenbarth @melonora.

LucaMarconato:
Hi @srivarra! Your package looks super cool! (CC @giovp) The syntax reads great and it removes a lot of verbosity.

A question, to loop-in the discussion above. What's your opinion of filtering the `AnnData` object followed by calling `match_sdata_to_table`(see tests here https://github.com/scverse/spatialdata/pull/627/files#diff-98b5bcd7dc2c08e7cd12d22b25e446f191b57f71b7be3f8255c813f58771d154): do you think that this 2-step process would be convenient enough for users that want to subset a `SpatialData` object by matching it to a `table` or should there be a single function allowing for all of it, and maybe more (`subset` to certain elements, spatial query, etc.)?

My opinion is that for the moment we could merge the linked PR, which introduces `match_sdata_to_table()`, and make a notebook showing what we do in the tests above. We could use your package in the notebook so that the syntax is even shorter.

I'd then close this PR and instead try things out for a while. Maybe with the pipe functionality that you wrote a while ago https://github.com/scverse/spatialdata/pull/722 we would not even need to have a single function that allows to filter everything and we could still achieve an ergonomic workflow. The pipe functionality in-fact would allow to combine `subset()`, `bounding_box_query()`, subsetting the `AnnData` table and matching it to elements in "one line".

WDYT?

By the way, if you have some time to leave a review for https://github.com/scverse/spatialdata/pull/627, I'd be grateful 😊

srivarra:
I’ve given this some thought, and while the two step approach can work, my perspective as a user is that a one step filtering method, similar to those found in DataFrame libraries like Pandas and Xarray would make for a more intuitive experience. Many users of SpatialData are likely familiar with those patterns, so providing a single, unified entry point seems preferable.

At first glance, the name `match_sdata_to_table` in #627 doesn’t immediately convey its purpose. After reading the docs, I have a rough idea of what it does, but renaming it to something like `filter_sdata_by_table` would better express its purpose (though naming is a secondary concern, I understand it does do some table matching / joins, but the experience on the user side is it's effectively a filter).

The proposed 2 step process, filtering the AnnData object first, then calling `match_sdata_to_table` feels like a temporary compromise.

> Thanks @aeisenbarth, after discussing with @melonora, we are going to first turn the code #627 into an internal function, merge, and then continue working on your PR. The idea is to provide a single entry point for filtering `filter()` and use for instance `subset()` or the function from Wouter internally.

I like this idea of making `match_sdata_to_table` an internal function to help handle the details.

For example, my vision for a unified filter function with expression functionality is as follows:

```python
Expressions = Expression | Iterable[Expression]

class SpatialData:
    # Other methods and attributes...

    def filter(
        self,
        elements: Iterable[str] | None,  # Elements to filter on (optional)
        # regions: Users would optionally set this in `obs_expr` if needed
        obs_expr: Expressions | None,     # Accept Narwhals expressions or an iterable of strings for observations
        var_expr: Expressions | None,     # Same as obs_expr, but for variable selection
        x_expr: Expressions | None,        # For filtering based on variable expression values
        obs_names_expr: Expressions | None, # For filtering based on table.obs_names
        var_names_expr: Expressions | None, # For filtering based on table.var_names
        table_name,                           # Ideally, this would work on a single table at a time to keep it simple
        layer: str | None,                # Filter based on a specific layer of the table, if provided, and using `x_expr`
        how: Literal["left", "left_exclusive", "inner", "right", "right_exclusive"] = "right",
    ):
        sdata_subset = self.subset(element_names=elements, filter_tables=True) if elements else self.sdata

        filtered_adata = sdata_subset[table_name].an.filter(obs_expr, var_expr, x_epxr, var_names, obs_names, layer)

        filtered_sdata = match_sdata_to_table(sdata=sdata_subset, table=filtered_adata, how=how)

        return filtered_sdata
```

Of course there would be a handful of extra checks and handling for corner cases and whatnot, but this is the general idea.

> Maybe with the pipe functionality that you wrote a while ago #722

Oh man, I completely forgot about this, looks like I accidentally closed it when going through and deleting up some forks I had. I'll see if I can recover that PR somehow.

If the current 2 step process was to be merged in today, I'd be satisfied using it, but hoping it would be touched up later down the line into one step.

And, if you'd like I could try to whip up a tiny implementation of filtering with Annsel / Narwhals expressions within a couple of weeks.

Thanks for your guys' great work, and let me know your thoughts!

LucaMarconato:
Thanks for the discussion! I really like how the API would look like and the arguments are convincing. I will check with other devs and try to first merge the other PRs, keep this PR as a discussion (closed PR) and ideally go for the new approach you described.

> Oh man, I completely forgot about this, looks like I accidentally closed it when going through and deleting up some forks I had. I'll see if I can recover that PR somehow.

Thanks a lot!

srivarra:
@melonora

> Personally, I have been a fan of the Polar's API but then again a lot of people are used to the pandas API so good tutorials / education is a must though I find it more clean than Pandas API.

Yeah I love the Polars API, it's super expressive and *pleasant* to use.

> Regarding function name I would probably not name it plain filter but something like `filter_by_table_query`. I know it is a bit more verbose but is more informative regarding what kind of filtering is being performed. But overall am in favour of this.

Yeah I think that's a very good name and I agree.

> Regarding the name `match_sdata_to_table`, I think this follows more the function definition we already had. It is hard to get feedback here from users whether they found the other function definitions intuitive or not so I am quite neutral about it.

I see that it follows similar functions like [`match_element_to_table()`][match_element_to_table_docs] and [`match_table_to_element()`][match_table_to_element_docs], so it makes sense to me.

[match_element_to_table_docs]: https://spatialdata.scverse.org/en/stable/api/operations.html#spatialdata.match_element_to_table


[match_table_to_element_docs]: https://spatialdata.scverse.org/en/stable/api/operations.html#spatialdata.match_table_to_element

LucaMarconato:
The linked PR introducing `merge_sdata_to_table()` is now merge, so as explained in https://github.com/scverse/spatialdata/pull/626#issuecomment-2653484912, I will close this PR.

> And, if you'd like I could try to whip up a tiny implementation of filtering with Annsel / Narwhals expressions within a couple of weeks.

As pointed out above having a `filter()` function would improve the ergonomics for the users. @srivarra if you could try out an implementation it would be great 😊 thanks!

Issue #217, opened 2023-04-05
Bug with `INSTANCE_KEY` for points?


LucaMarconato:
For the design doc, for points elements we have

> It MAY contains additional information in dask.dataframe.DataFrame().attrs["spatialdata_attrs"], specifically:
> 
> "instance_key": the column name of the table where unique instance ids that this point refers to are stored, if available.

I think this is not sufficient to identify instances, we need also a `region_key` column, since there could be multiple regions elements that share the same indices for identifying the instances.

We need to:
- [ ] check and fix this in the schema
- [ ] update the design doc
- [ ] add tests
- [ ] add the validation for these relationships, see here: https://github.com/scverse/spatialdata/issues/218

PR #627, opened 2024-07-09
allow filtering by ids

### LINKED ISSUE #280: Filter spatialData

### LINKED ISSUE #284: Feature request: spatial cropping from select table rows

### LINKED ISSUE #556: Subset spatialdata by list of cell ids

melonora:
closes #556 

This PR allows for filtering elements (points, shapes, tables) on instance IDs. For labels I decided just as with the joins to not support it as it is potentially an expensive operation, where I am not certain what the usecase would be (one can adjust the color to a background color in any case for particular labels or even set it transparent).

The PR is related to #626, however I think it is better to separate out filtering on instance  IDs. The function here can ultimately still be used in his function.

The PR will not cause an error in case of filtering on instance IDs that are not there. Furthermore it is not a problem to filter on instances in a table that is annotating more than 1 element. If instances for both annotated elements need to be filtered in the table this will be retrieved. If `region_names` is defined, a view of the table with rows annotating elements in `region_names` is first created before filtering on instance IDs.

After filtering of the table, the metadata for annotation of regions is checked and updated after which the table is validated.

## Changelog
- added new API `match_sdata_to_table()` which allows for complex subsetting of `SpatialData` objects
- changed `SpatialData.get_annotated_regions()` to compute on-the-fly the regions annotated by a table, instead of returning `table.uns['spatialdata_attrs']['region']
- added `overwrite_metadata` parameter to `TableModel.parse()` to conveniently replace `region`, `region_key`, `instance_key` metadata.

Update Luca (as shown with the code in the [tests](https://github.com/scverse/spatialdata/pull/627/files#diff-98b5bcd7dc2c08e7cd12d22b25e446f191b57f71b7be3f8255c813f58771d154)):
closes #280 
closes #284


LucaMarconato:
@melonora converting to draft, what's missing for the moment, just the tests?

LucaMarconato:
- [ ] ah also the function should be private so it can be used internally by a single "entry point" `filter()`.

LucaMarconato:
Hi, I reviewed the PR. First please check my comment in this linked PR: https://github.com/scverse/spatialdata/pull/626#issuecomment-2615467570.

In this PR, I refined the function described in the comment linked above, into a new public API `match_sdata_to_table()`. As shown in the tests, all the use cases of the linked PR are supported. The use cases of filtering by instance_ids from points/shapes are supported by the functions `match_table_to_element()` (or by the join operations).

Since the purpose of the current PR was to add a private function that could support the `filter()` method of the linked PR, and since now the new public function allows for this. I removed the `filter()` method. 

This is supported also by the following reasons:
- the `filter()` function was supposed to be private, but now we do not need such private function
- the method introduces code redundancy that can be avoided by using the join APIs (this is how I implemented `match_sdata_to_table()`

Finally, since two comments that I had from the review of the `filter()` method (even if not relevant anymore if the method is removed).
- the points were not filtered by the index, but by the `instance_key` column, whose role is to match the transcript to a cell index to which the transcript below to. I had fixed this in [`122aa5f` (#627)](https://github.com/scverse/spatialdata/pull/627/commits/122aa5f22ecbd0cc8c3193dd0ff55a4d25dec1f7). Note that the semantic of `instance_key` column for points is not well defined as reported here https://github.com/scverse/spatialdata/issues/217.
- the old API (with the parameter `element_names`) did not allow to specify particular instances for particular elements. For instance given two segmentation masks (shapes) "shapes0" and "shapes1" annotated by a table. It was not possible to select the cells [0, 1] for "shapes0", but only the cell [0] for "shapes1". The new implementation allows for these stratified types of filters.

LucaMarconato:
Ready for review! I tag both @melonora and @aeisenbarth please.

src/spatialdata/models/models.py
```diff
@@ -1129,3 +1137,20 @@ def get_table_keys(table: AnnData) -> tuple[str | list[str], str, str]:
...
+        warnings.warn(
+            f"The region key column `{region_key}` is not of type `pd.Categorical`. Consider casting it to "
+            f"improve performance.",
+            UserWarning,
+            stacklevel=2,
+        )
+        annotated_regions = region_key_column.unique().tolist()
+    else:
+        annotated_regions = table.obs[region_key].cat.remove_unused_categories().cat.categories.unique().tolist()
+    assert isinstance(annotated_regions, list)
```
melonora, line 1155: why this assert?

tests/core/query/test_relational_query.py
```diff
@@ -901,3 +901,11 @@ def test_get_element_annotators(full_sdata):
...
     names = get_element_annotators(full_sdata, "labels2d")
     assert names == {"another_table", "table"}
+
+
+# def test_match_table_to_element(sdata_query_aggregation):
+#     matched_table = match_table_to_element(sdata=sdata_query_aggregation, element_name="values_circles")
+#     arr = np.array(list(reversed(sdata_query_aggregation["values_circles"].index)))
+#     sdata_query_aggregation["values_circles"].index = arr
+#     matched_table_reversed = match_table_to_element(sdata=sdata_query_aggregation, element_name="values_circles")
+#     assert matched_table.obs.index.tolist() == list(reversed(matched_table_reversed.obs.index.tolist()))
```
melonora, line 911:
```suggestion
```
Remove please unless you plan to use it:)

src/spatialdata/_core/query/relational_query.py
```diff
@@ -773,6 +773,45 @@ def match_element_to_table(
...
+    Filter the elements of a SpatialData object to match only the rows present in the table.
+
+    Parameters
+    ----------
+    sdata
+        SpatialData object containing all the elements and tables.
+    table
+        The table to join with the spatial elements. Has precedence over `table_name`.
+    table_name
+        The name of the table to join with the SpatialData object if `table` is not provided. Also, `table_name` is used
```
melonora, line 792:
```suggestion
        The name of the table to join with the SpatialData object if `table` is not provided. If table is provided, `table_name` is used
```

tests/core/query/test_relational_query_match_sdata_to_table.py
```diff
@@ -0,0 +1,144 @@
...
+    assert "blobs_polygons-sdata1" in matched
+    assert "blobs_polygons-sdata2" not in matched
+
+
+def test_match_sdata_to_table_shapes_and_points():
+    """
+    The function works both for shapes (examples above) and points.
+    Changes the target of the table to labels.
+    """
+    sdata = _make_test_data()
```
melonora, line 83:
```suggestion
```
defeats the purpose of the global variable no?

src/spatialdata/models/models.py
LucaMarconato: true, removed.

tests/core/query/test_relational_query.py
LucaMarconato: Removed.

src/spatialdata/models/models.py
LucaMarconato: `annotated_regions` could be `Any`; this assert is for `mypy`.

PR #636, opened 2024-07-15
small fixes rasterize_bins() and get_values()

LucaMarconato:
- only csr matrices were supported in `get_values()`, fixed
- improved docs for `rasterize_bins()` to show how to plot the rasterized image with `spatialdata-plot`


LucaMarconato:
Docs build locally, not sure what's the problem. Gonna fix another time.

PR #647, opened 2024-07-30
Remove macos arm workaround description

Czaki:
Same like https://github.com/scverse/napari-spatialdata/pull/279

There are pyqt5 wheels for macos available. 

The current list of dependencies in `extra` do not trigger pyqt5 installation. so maybe `"napari-spatialdata>=0.2.4"` should be changed to `"napari-spatialdata[all]>=0.2.4" or installation instruction should contain information about Qt backend installation. For example, link to napari documentation https://napari.org/stable/tutorials/fundamentals/installation.html#choosing-a-different-qt-backend

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,7 @@ torch = [
     "torch"
 ]
 extra  = [
-    "napari-spatialdata>=0.2.4",
+    "napari-spatialdata[all]>=0.2.4",
     "spatialdata-plot",
     "spatialdata-io",
 ]


LucaMarconato:
Thanks for the PR. Before merging I would resolve your open question regarding the dependencies.

Citing the docs you linked:
> Running python -m pip install "napari[all]" will install the default framework, which is currently PyQt5–but this could change in the future. However, if you have a Mac with the newer arm64 architecture (Apple Silicon), this will not work–see [note on m1](https://napari.org/stable/tutorials/fundamentals/installation.html#note-m1).

If I understood correctly, replacing `napari-spatialdata>=0.2.4` with `napari-spatialdata[all]>=0.2.4` would still not work on M1. But `napari-spatialdata[pyqt5]>=0.2.4` should work.

In that case I would use this last dependency. What do you think?

Czaki:
Ahh. We need to update this part of documentation. It is the same artifact as this one that we remove in This PR. 

The only backend that does not support macOS M chips is pyside2.

PR #654, opened 2024-08-01
Fix dataloader table names

LLehner:
The `ImageTilesDataset` class cannot be used if the table you want to work with is not called "table", since there are lines where this name is hard-coded, even though the class allows for a `table_name` to be specified. 

An example where this fails is if you follow this [tutorial](https://spatialdata.scverse.org/en/latest/tutorials/notebooks/notebooks/examples/densenet.html) but specify a different table name when creating the dataset. 

This PR replaces instances of "table" with `table_name`.


LucaMarconato:
Thanks @LLehner for the contribution. Only minor adjustments were neeed. Merging now and fixing the docs before release.

PR #656, opened 2024-08-02
Simplified points model handling of divisions

LucaMarconato:
I simplified the handling of divisions in the points model when dealing with a non-ascending index and added tests.


LucaMarconato:
CC @giovp

PR #660, opened 2024-08-05
Update installation link and instructions

namsaraeva:
I had trouble finding the installation instructions for M1 Mac, so I updated the link. Thank you for the package!


LucaMarconato:
Ah thanks! Actually with the incoming release (maybe we are doing it today), the workaround for Apple Silicon macs will not be needed anymore https://github.com/scverse/spatialdata/pull/647.

I will therefore edit this PR to drop that line from the readme, but merge it only right before making the release.

Issue #597, opened 2024-06-24
check on `c_coords` should be one the output data in `map_raster`


giovp:
You are right, the check on `c_coords` should be one the output data. I merged already because I need to merge #594, which was opened against the current PR. Please can you make a patch for fixing that behavior?

_Originally posted by @LucaMarconato in https://github.com/scverse/spatialdata/issues/588#issuecomment-2182817806_

LucaMarconato:
CC @ArneDefauw

LucaMarconato:
Thanks! That's an interesting API! It could be implemented making use of the new `map_raster()` right?

ArneDefauw:
yes, indeed. Would you prefer adding it as a parameter to `map_raster()`? Say a boolean flag, `relabel_blocks`, ignored if output type is not a labels layer or if blockwise is False.
Or would you prefer wrapping map_raster in a seperate function, which will take care of relabeling chunks if output layer is a labels layer?

Wouldn't be too difficult to implement, I can prepare a PR if you would be interested.

LucaMarconato:
I think I would wrap it in a separate function, @giovp @melonora wdyt?

giovp:
that sounds really useful @ArneDefauw ! I think it depends a bit, if it's just a single argument, then maybe can be done in map_raster? as in, we have already a very extensive list of functions/methods in the repo, but I also understand on striking a balance of composability and fleixiblity

LucaMarconato:
Good point! If using `map_raster()` for that use case would feel unnatural, or if the functionality would be "too hidden" I would go for a separate function with an obvious name.

ArneDefauw:
I agree, I think we can first try to wrap it in map_raster.

@LucaMarconato , I don't think I understand the type of relabel (`relabel: dict[str:str]`) you propose.  I think relabel can be a boolean. 
I was planning to implement relabeling via use of a bit shift. This is how it was done in squidpy, and is also how I typically implement it to avoid collisions when I do not know which labels will be predicted for each block. I.e. something like:
...
        num_blocks = x.numblocks
        shift = int(np.prod(num_blocks[0] * num_blocks[1] * num_blocks[2]) - 1).bit_length()

...
        block_num = block_id[0] * (num_blocks[1] * num_blocks[2]) + block_id[1] * (num_blocks[2]) + block_id[2]
        mask: NDArray = labels > 0
        labels[mask] = (labels[mask] << shift) | block_num

giovp:
[...]

PR #664, opened 2024-08-07
relabel block

ArneDefauw:
Follow up PR for https://github.com/scverse/spatialdata/pull/588, see discussion https://github.com/scverse/spatialdata/issues/597 for relabeling in case output data is a labels layer (i.e.  `dims` does not contain `c`) via use of a bit shift.

### DIFF
--- a/src/spatialdata/_core/operations/map.py
+++ b/src/spatialdata/_core/operations/map.py
@@ -1,14 +1,19 @@
 from __future__ import annotations
 
+import math
+import operator
 from collections.abc import Iterable, Mapping
+from functools import reduce
 from types import MappingProxyType
 from typing import TYPE_CHECKING, Any, Callable
 
 import dask.array as da
+import numpy as np
 from dask.array.overlap import coerce_depth
 from datatree import DataTree
 from xarray import DataArray
 
+from spatialdata._types import ArrayLike
 from spatialdata.models._utils import get_axes_names, get_channels, get_raster_model_from_data_dims
 from spatialdata.transformations import get_transformation
 
@@ -25,6 +30,7 @@ def map_raster(
     c_coords: Iterable[int] | Iterable[str] | None = None,
     dims: tuple[str, ...] | None = None,
     transformations: dict[str, Any] | None = None,
+    relabel: bool = True,
     **kwargs: Any,
 ) -> DataArray:
     """
@@ -69,6 +75,11 @@ def map_raster(
     transformations
         The transformations of the output data. If not provided, the transformations of the input data are copied to the
         output data. It should be specified if the callable changes the data transformations.
+    relabel
+        Whether to relabel the blocks of the output data.
+        This option is ignored when the output data is not a labels layer (i.e., when `dims` does not contain `c`).
+        It is recommended to enable relabeling if `func` returns labels that are not unique across chunks.
+        Relabeling will be done by performing a bit shift.
     kwargs
         Additional keyword arguments to pass to :func:`dask.array.map_overlap` or :func:`dask.array.map_blocks`.
         Ignored if `blockwise` is set to `False`.
@@ -131,6 +142,9 @@ def map_raster(
             assert isinstance(d, dict)
         transformations = d
 
+    if "c" not in dims and relabel:
+        arr = _relabel(arr)
+
     model_kwargs = {
         "chunks": arr.chunksize,
         "c_coords": c_coords,
@@ -139,3 +153,56 @@ def map_raster(
     }
     model = get_raster_model_from_data_dims(dims)
     return model.parse(arr, **model_kwargs)
+
+
+def _relabel(arr: da.Array) -> da.Array:
+    num_blocks = arr.numblocks
+
+    shift = (math.prod(num_blocks) - 1).bit_length()
+
+    meta = np.empty((0,) * arr.ndim, dtype=arr.dtype)
+
+    def _relabel_block(
+        block: ArrayLike, block_id: tuple[int, ...], num_blocks: tuple[int, ...], shift: int
+    ) -> ArrayLike:
+        def _calculate_block_num(block_id: tuple[int, ...], num_blocks: tuple[int, ...]) -> int:
+            if len(num_blocks) != len(block_id):
+                raise ValueError("num_blocks and block_id must have the same length")
+            block_num = 0
+            for i in range(len(num_blocks)):
+                multiplier = reduce(operator.mul, num_blocks[i + 1 :], 1)
+                block_num += block_id[i] * multiplier
+            return block_num
+
+        available_bits = np.iinfo(block.dtype).max.bit_length()
+        max_bits_block = int(block.max()).bit_length()
+
+        if max_bits_block + shift > available_bits:
+            raise ValueError(
+                f"Relabel was set to True, but "
+                f"max bits required to represent the labels in the block ({max_bits_block}) "
+                f"+ required shift ({shift}) > "
+                f"available_bits ({available_bits}). "
+                "To solve this issue, please consider rechunking using a larger chunk size, "
+                "resulting in a fewer number of blocks and thus a lower value for the required shift; "
+                f"cast to a data type (current data type is {block.dtype}) with a higher maximum value "
+                "(resulting in more available bits for the bit shift); "
+                "or consider a sequential relabeling of the dask array, "
+                "which could result in a lower maximum value of the labels in the block."
+            )
+
+        block_num = _calculate_block_num(block_id=block_id, num_blocks=num_blocks)
+
+        mask = block > 0
+        block[mask] = (block[mask] << shift) | block_num
+
+        return block
+
+    return da.map_blocks(
+        _relabel_block,
+        arr,
+        dtype=arr.dtype,
+        num_blocks=num_blocks,
+        shift=shift,
+        meta=meta,
+    )
--- a/src/spatialdata/_core/operations/map.py
+++ b/src/spatialdata/_core/operations/map.py
@@ -12,7 +12,7 @@ import numpy as np
 from dask.array.overlap import coerce_depth
 from xarray import DataArray, DataTree
 
-from spatialdata._types import IntArray
+from spatialdata._types import IntArrayLike
 from spatialdata.models._utils import get_axes_names, get_channel_names, get_raster_model_from_data_dims
 from spatialdata.transformations import get_transformation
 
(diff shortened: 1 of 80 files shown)


ArneDefauw:
Now also added helper function` _relabel_sequential`, see https://github.com/ArneDefauw/spatialdata/blob/633a132ddfb6353925bc099114b893df43ad3f2e/src/spatialdata/_core/operations/map.py#L211, as discussed here https://github.com/scverse/spatialdata/issues/597#issuecomment-2225678023

I think the PR is ready for review now.

giovp (approved):
hi @ArneDefauw , thanks a lot, super useful. I went back to this function in squidpy, and remember that we also did an additional step to harmonize the labels with dask image

```
from dask_image.ndmeasure._utils._label import (
            connected_components_delayed,
            label_adjacency_graph,
            relabel_blocks,
        )

# max because labels are not continuous (and won't be continuous)
label_groups = label_adjacency_graph(img, None, img.max())
new_labeling = connected_components_delayed(label_groups)
relabeled = relabel_blocks(img, new_labeling)
```
any reason for leaving this out? Otherwise for me it is good to go!

ArneDefauw:
> hi @ArneDefauw , thanks a lot, super useful. I went back to this function in squidpy, and remember that we also did an additional step to harmonize the labels with dask image
> 
> ```python
> from dask_image.ndmeasure._utils._label import (
>             connected_components_delayed,
>             label_adjacency_graph,
>             relabel_blocks,
>         )
> 
> # max because labels are not continuous (and won't be continuous)
> label_groups = label_adjacency_graph(img, None, img.max())
> new_labeling = connected_components_delayed(label_groups)
> relabeled = relabel_blocks(img, new_labeling)
> ```
> 
> any reason for leaving this out? Otherwise for me it is good to go!

Hi @giovp ,

thanks for the review!

Yes, I did not include the step to harmonize the labels across chunks due to some issues I observed when cell density is high, see also my previous comments. 
https://github.com/scverse/spatialdata/issues/597#issuecomment-2213297632
https://github.com/scverse/spatialdata/issues/597#issuecomment-2224762535

Maybe a little bit more clear with example. 
So this is map_raster on a labels layer without harmonizing labels:

```
#I use the harpy package (https://github.com/saeyslab/harpy) for downloading of some test data
from sparrow.datasets import mibi_example
sdata = mibi_example()
def copy_labels( arr ):
    return arr

results=map_raster( sdata[ "masks_whole" ].chunk( 212 ), func=copy_labels, blockwise=True, relabel=True, )

sdata[ "masks_whole_copied_relabeled_no_merging" ] = results
sdata.write_element( "masks_whole_copied_relabeled_no_merging" )
sdata = read_zarr( sdata.path )
```

Then without using the harmonizing of labels I get:
Screenshot 2024-08-26 at 11 35 57


and with harmonizing of the labels using


```
from dask_image.ndmeasure._utils._label import (
            connected_components_delayed,
            label_adjacency_graph,
            relabel_blocks,
        )

# max because labels are not continuous (and won't be continuous)
label_groups = label_adjacency_graph(img, None, img.max())
new_labeling = connected_components_delayed(label_groups)
relabeled = relabel_blocks(img, new_labeling)
```

I get:

Screenshot 2024-08-26 at 11 30 33


Which is not great, cells on the boundary of the chunks are merged together. 

A solution for this, is using this approach:
 https://github.com/gletort/NeubiasPasteur2023_AdvancedCellPose/blob/main/distributed_segmentation.py?rgh-link-date=2024-07-08T08%3A01%3A09Z for merging chunks back together. 
Also implemented here https://github.com/saeyslab/harpy/blob/e4966aaf12ff0c36e301ff27df675c2c96f87ed2/src/sparrow/image/segmentation/_segmentation.py#L494, 
which fixes much of these issues. 

But I do not know if you want this to be included in this PR, see also comment of @LucaMarconato  https://github.com/scverse/spatialdata/issues/597#issuecomment-2225678023. 
I would be happy to prepare a follow up PR.

giovp:
hi @ArneDefauw , sorry for late reply. 

I took a look at 

> [gletort/NeubiasPasteur2023_AdvancedCellPose@main/distributed_segmentation.py?rgh-link-date=2024-07-08T08%3A01%3A09Z](https://github.com/gletort/NeubiasPasteur2023_AdvancedCellPose/blob/main/distributed_segmentation.py?rgh-link-date=2024-08-26T09%3A53%3A28Z)

it's a nice approach I think, and I wonder if it should be included already here. Meaning, I am not sure how useful it is the output of relabelling without the across-block label harmonization. Yes, relabelling fix one step of the chunked segmentation, but the result is not really actionable/useful right? Or am I missing something?

also @LucaMarconato wdyt? you can take a look at the output of the excellent post from @ArneDefauw before to understand visually what is the problem

ArneDefauw:
> hi @ArneDefauw , sorry for late reply.
> 
> I took a look at
> 
> > [gletort/NeubiasPasteur2023_AdvancedCellPose@main/distributed_segmentation.py?rgh-link-date=2024-07-08T08%3A01%3A09Z](https://github.com/gletort/NeubiasPasteur2023_AdvancedCellPose/blob/main/distributed_segmentation.py?rgh-link-date=2024-08-26T09%3A53%3A28Z)
> 
> it's a nice approach I think, and I wonder if it should be included already here. Meaning, I am not sure how useful it is the output of relabelling without the across-block label harmonization. Yes, relabelling fix one step of the chunked segmentation, but the result is not really actionable/useful right? Or am I missing something?
> 
> also @LucaMarconato wdyt? you can take a look at the output of the excellent post from @ArneDefauw before to understand visually what is the problem

Yes, indeed, if you do not know how to do the across-block label harmonization, the result will not be very useful. On the other hand, there are probably many ways to do it. 
Maybe spatialdata should at least implement one way to do this label harmonization. 
But I think it is up to you guys to decide :)

melonora:
@ArneDefauw thanks for the PR and apologies for taking some time to get to it. After discussing with @LucaMarconato and @berombau we decided to put the label block harmonization in `squidpy` as this library is for having user friendly operations like this. We do not necessarily want the user to have to think about `blocks` and the implications of that, rather the average user just wants a function `relabel`.

src/spatialdata/_core/operations/map.py
```diff
@@ -68,6 +74,11 @@ def map_raster(
     transformations
         The transformations of the output data. If not provided, the transformations of the input data are copied to the
         output data. It should be specified if the callable changes the data transformations.
+    relabel
+        Whether to relabel the blocks of the output data.
+        This option is ignored when the output data is not a labels layer (i.e., when `dims` does not contain `c`).
+        It is recommended to enable relabeling if `func` returns labels that are not unique across chunks.
+        Relabeling will be done by performing a bit shift.
```
melonora, line 81:
```suggestion
    relabel
        Whether to relabel the blocks of the output data.
        This option is ignored when the output data is not a labels layer (i.e., when `dims` does not contain `c`).
        It is recommended to enable relabeling if `func` returns labels that are not unique across chunks.
        Relabeling will be done by performing a bit shift. Current implementation does not harmonize 
        labels across blocks, see https://github.com/scverse/spatialdata/pull/664 for discussion.
```

ArneDefauw:
> @ArneDefauw thanks for the PR and apologies for taking some time to get to it. After discussing with @LucaMarconato and @berombau we decided to put the label block harmonization in `squidpy` as this library is for having user friendly operations like this. We do not necessarily want the user to have to think about `blocks` and the implications of that, rather the average user just wants a function `relabel`.

Hi @melonora , that makes sense, thanks for having a look at the PR!

melonora:
I adjusted the error message a tiny bit. While the function is private, it is called by public API. A less experienced programmer might not know about bit shift.

melonora:
Also made `relabel_sequential` public given that @ArneDefauw put it as suggestion in the error message which in itself is in a private function, but the error message can be shown by calling a public function.

src/spatialdata/_types.py
```diff
@@ -7,6 +7,8 @@
     from numpy.typing import DTypeLike, NDArray
 
     ArrayLike = NDArray[np.float64]
+    IntArray = NDArray[np.int64]  # or any np.integer
```
LucaMarconato, line 10: rename?

tests/core/operations/test_map.py
```diff
@@ -28,6 +30,11 @@ def _multiply_to_labels(arr, parameter=10):
     return arr[0].astype(np.int32)
 
 
+def _to_constant(arr, constant=5):
```
LucaMarconato, line 33: why 5?

LucaMarconato, line 33: remove default

LucaMarconato, line 33: what's the purpose of the default here? I'd remove it to be explicit.

src/spatialdata/_core/operations/map.py
```diff
@@ -138,3 +154,93 @@ def map_raster(
...
+        meta=meta,
+    )
+
+
+def relabel_sequential(arr: da.Array) -> da.Array:
+    """
+    Relabels integers in a Dask array sequentially.
+
+    This function assigns sequential labels to the integers in a Dask array starting from 1.
+    For example, if the unique values in the input array are [0, 5, 9],
```
LucaMarconato, line 219: I'd use [0, 9, 5] as example, and add this is the tests.

src/spatialdata/_core/operations/map.py
```diff
@@ -138,3 +154,93 @@ def map_raster(
...
+    arr
+        input array.
+
+    Returns
+    -------
+    The relabeled array.
+    """
+    if not np.issubdtype(arr.dtype, np.integer):
+        raise ValueError(f"Sequential relabeling is only supported for arrays of type {np.integer}.")
+    unique_labels = da.unique(arr).compute()
```
LucaMarconato, line 235: I'd add a comment to remind that the unique values that are returned are sorted.

src/spatialdata/_types.py
LucaMarconato: to ArrayLikeInt? Also, in case of ImportError/TypeError ArrayLikeInt is not defined.

src/spatialdata/_core/operations/map.py
```diff
@@ -138,3 +154,93 @@ def map_raster(
...
+                f"Relabel was set to True, but "
+                f"the number of bits required to represent the labels in the block ({max_bits_block}) "
+                f"+ required shift ({shift}) exceeds the available_bits ({available_bits}). In other words"
+                f"the number of labels exceeds the number of integers that can be represented by the dtype"
+                "of the individual blocks."
+                "To solve this issue, please consider the following solutions:"
+                "   1. Rechunking using a larger chunk size, lowering the number of blocks and thereby"
+                "      lowering the value of required shift."
+                "   2. Cast to a data type with a higher maximum value  "
+                "   3. Perform sequential relabeling of the dask array, potentially lowering the total "
```
LucaMarconato, line 193: I'd explain this better:
1) that we provide the relabel_sequential function
2) that the number of labels remain the same, but the max label is lower

src/spatialdata/_types.py
melonora: check for later PR. Changing this throws a bunch of mypy errors

LucaMarconato:
Thanks @ArneDefauw for the PR and @melonora for addressing the review comments. We are good to merge!

LucaMarconato (approved):

PR #672, opened 2024-08-12
Add sort parameter to points model

LucaMarconato:
In `spatialdata-io` (e.g. in `xenium()`) we generally want the `sort=True`, here I enable this via a parameter.


LucaMarconato:
The test that fails is unrelated to this PR.

PR #622, opened 2024-07-08
improves dataloader performance

giovp:
replaces #565 , should be merged after #617

PR #687, opened 2024-08-21
improves dataloader performance

giovp:
replaces #565 #622

iteration over 20 batches, single worker

new implementation
image

main
image

one annoying thing is that the "apply" method of the dataframe to get the bounding box selection is quite slow.


giovp:
quick push to try #699 where tiling is vectorized, removed the need to the `pandas.DataFrame.apply`. Quite big speedup

image

LucaMarconato:
I picked this up, performance indeed improves significantly with the new vectorized bounding box approach. Thanks @giovp 

`asv` benchmarks result:

```
| Change   | Before [accf496c] <main>   | After [e87d3183] <giovp/dataloader3>   |   Ratio | Benchmark (Parameter)                          |
|----------|----------------------------|----------------------------------------|---------|------------------------------------------------|
| -        | 618±20ms                   | 123±1ms                                |    0.2  | benchmark_dataloader.TimeDataloader.time_fetch |
| -        | 169±20ms                   | 18.7±0.1ms                             |    0.11 | benchmark_dataloader.TimeDataloader.time_init  |
```

PR #697, opened 2024-08-31
Fix labels multiscales method

### LINKED ISSUE #696: Generating multiscales for labels causes artifacts
When parsing a labels element with scale factors, I unconsciously obtained artifacts.

Given a single-scale labels image with unsigned integer dtype and passing `scale_factors`, I assumed SpatialData chooses from the `Labels2DModel` the corresponding scaling method that preserves unique labels (does not introduce new labels as interpolation result from two neighboring labels).

However, it leaves the decision to `to_multiscale`, which defaults to `XARRAY_COARSEN`, same as with normal images.

That means, when passing `scale_factors` to labels, `method` must also be passed explicitly. This seems unintuitive and more users will fall into this trap. I suggest when the class is Image2DLabels / Image3DLabels, we override the default `method`.

```python
import numpy as np
from spatialdata.models import Labels2DModel
from multiscale_spatial_image.to_multiscale.to_multiscale import Methods
import pytest


@pytest.mark.parametrize(
    # "method" not passed, passed as None and passed with "nearest" interpolation
    "method_kwarg", [{}, {"method": None}, {"method": Methods.DASK_IMAGE_NEAREST}]
)
def test_parse_labels_multiscale(method_kwarg):
    # A labels image with one label value 4, that partially covers 2×2 blocks.
    # Downsampling with interpolation would produce values 1, 2, 3, 4.
    array = np.array(
        [[0, 0, 0, 0], [0, 4, 4, 4], [4, 4, 4, 4], [0, 4, 4, 4]], dtype=np.uint16
    )
    actual = Labels2DModel.parse(array, scale_factors=(2,), **method_kwarg)
    assert actual.scale0.image.dtype == array.dtype
    assert actual.scale1.image.dtype == array.dtype
    assert set(np.unique(array)) == set(
        np.unique(actual.scale0.image)
    ), "Scale0 should be preserved"
    assert set(np.unique(array)) >= set(
        np.unique(actual.scale1.image)
    ), "Subsequent scales should not have interpolation artifacts"
```

issue_labels_multiscales

I also was sure the exact same method call was once generating multiscales without artifacts, but I don't know when, and the above example reproduces that there is a problem.

Versions:
- multiscale_spatial_image==1.0.1
- spatial_image==1.1.0
- spatialdata==0.2.2

aeisenbarth:
Closes #696 

When passing `scale_factors` to a RasterSchema, and it is for labels instead of images, the default for `to_multiscale`'s `method` argument was `xarray_coarsen` and caused artifacts on labels edges. It is now overriden to `dask_image_nearest`. This avoids artifacts if the user does not explicitly pass a method.

Questions:
- Is there any other scaling method to consider, or can `dask_image_nearest` handle all array types?
- Pydocstyle complains about missing docstring. It is a static analyzer and cannot detect inheritance of docstrings. I didn't want to copy the complete docstring and cause duplication. Currently, I disabled the check. How do you prefer to handle this?

### DIFF
--- a/src/spatialdata/models/models.py
+++ b/src/spatialdata/models/models.py
@@ -255,6 +255,17 @@ class Labels2DModel(RasterSchema):
             **kwargs,
         )
 
+    @classmethod
+    def parse(  # noqa: D102
+        self,
+        *args: Any,
+        **kwargs: Any,
+    ) -> DataArray | DataTree:
+        if kwargs.get("scale_factors") is not None and kwargs.get("method") is None:
+            # Override default scaling method to preserve labels
+            kwargs["method"] = Methods.DASK_IMAGE_NEAREST
+        return super().parse(*args, **kwargs)
+
 
 class Labels3DModel(RasterSchema):
     dims = DimsSchema((Z, Y, X))
@@ -269,6 +280,13 @@ class Labels3DModel(RasterSchema):
             *args,
             **kwargs,
         )
+
+    @classmethod
+    def parse(self, *args: Any, **kwargs: Any) -> DataArray | DataTree:  # noqa: D102
+        if kwargs.get("scale_factors") is not None and kwargs.get("method") is None:
+            # Override default scaling method to preserve labels
+            kwargs["method"] = Methods.DASK_IMAGE_NEAREST
+        return super().parse(*args, **kwargs)
 
 
 class Image2DModel(RasterSchema):
--- a/tests/models/test_models.py
+++ b/tests/models/test_models.py
@@ -195,6 +195,27 @@ class TestModels:
             with pytest.raises(ValueError):
                 model.parse(image, **kwargs)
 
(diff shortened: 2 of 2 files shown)


giovp (approved):
this is great @aeisenbarth , thanks! LGTM!

> Is there any other scaling method to consider, or can `dask_image_nearest` handle all array types?

I think this is most sensible for labels, as it breaks the labeling with default bilinear I believe. However, I wouldn't change the default for images IMHO.

> Pydocstyle complains about missing docstring. It is a static analyzer and cannot detect inheritance of docstrings. I didn't want to copy the complete docstring and cause duplication. Currently, I disabled the check. How do you prefer to handle this?

Yes, this is a pain, in a separate project we've ended up using https://github.com/AntoineD/docstring-inheritance it seems maintainer but not very adopted. Let me ask on scverse zulip what people think. For now, I agree to disable it.

PR #699, opened 2024-09-02
vectorize bounding box query

giovp:
Trying to vectorize the tiling in the dataloader PR, realized some improvements should be added separately.

With this PR, I enable the vectorization of `bounding_box_query`  for all elements. 

This means that it is now possible to pass an array of bounding boxes (and not just one). 

TODO:
- [x] tests for multiple bounding box queries for raster data
- [x] tests for multiple bounding box queries for shapes
- [x] tests for multiple bounding box queries for points

@LucaMarconato @melonora do you have any suggestion of when this could be used to replace current implementations across the projects? A clear use case (and the motivation for this contribution) is the dataloader, see #687 for the speedup, but I wonder if there are other places where this is useful. 

This also is the groundwork for eventual update to the dataloader, being able to return batches of all elements.

considering that this PR is already getting too large, I would postpone the vectorization of polygon query.

### DIFF
--- a/src/spatialdata/_core/query/_utils.py
+++ b/src/spatialdata/_core/query/_utils.py
@@ -2,14 +2,22 @@ from __future__ import annotations
 
 from typing import Any
 
+import numba as nb
 import numpy as np
 from anndata import AnnData
+from datatree import DataTree
 from xarray import DataArray
 
 from spatialdata._core._elements import Tables
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata._types import ArrayLike
 from spatialdata._utils import Number, _parse_list_into_array
+from spatialdata.transformations._utils import compute_coordinates
+from spatialdata.transformations.transformations import (
+    BaseTransformation,
+    Sequence,
+    Translation,
+)
 
 
 def get_bounding_box_corners(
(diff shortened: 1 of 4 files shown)


giovp:
ready for a first pass on review

giovp:
ready for review

LucaMarconato:
Thank you @giovp for the PR, I will review now.

> A clear use case (and the motivation for this contribution) is the dataloader, see https://github.com/scverse/spatialdata/pull/687 for the speedup, but I wonder if there are other places where this is useful.

I can't think of other places right now, so I think we are good with the current improvement.

> considering that this PR is already getting too large, I would postpone the vectorization of polygon query.

I agree.

src/spatialdata/_core/query/_utils.py
```diff
@@ -36,37 +45,146 @@ def get_bounding_box_corners(
...
+        },
     )
+    if num_boxes > 1:
+        return output
+    return output.squeeze().drop_vars("box")
+
+
+@nb.njit(parallel=False, nopython=True)
+def _create_slices_and_translation(
+    min_values: nb.types.Array[nb.float64, nb.float64],
```
LucaMarconato, line 101: I think `nb.types.Array[nb.float64, np.float64]` is incorrect and the correct version is `nb.types.Array(nb.float64, 2, 'C')`. But I am not sure.

LucaMarconato: Mine is wrong. Are you sure about your typing? What looks strange to me is that types like `nb.types.Array[nb.float64, nb.int64]` would not have a meaning. I tried using just `nb.types.Array` and pre-commit works, maybe this is the way to go.

giovp: I will look into it!

LucaMarconato:
@giovp I finished reviewing; I applied some minor changes like simplified the docs and added an extra test.

PR #701, opened 2024-09-05
refactor `filter_table_by_elements`

giovp:
I would like to refactor this function in order to support multiple tables.

### Features

Let's say I have a region element `shapes1`, I would like to filter the table or tables that contain instances from that region. This filtering should have the following features:
- [ ] do not copy, but return a view
- [ ] optionally return a mapping `table_name -> instances` or the same list of tables

@melonora @LucaMarconato I have looked back at the multiple table in-memory design, but I couldn't find a clear description of any constraint or expected behaviour in the conversation or in the design document. In particular, from our previous conversations, I understood that:
- a table can map to multiple elements (e.g. diff views of the same annotation)
- a table can have rows that map to an element, and rows that don't map to any element
- a table can also not map to any element

can two tables map to the same element?
Thanks for the clarification

### DIFF
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -163,40 +163,59 @@ def _filter_table_by_elements(
     -------
     The filtered table (eventually with reordered rows), or None if the input table was None.
     """
-    assert set(elements_dict.keys()).issubset({"images", "labels", "shapes", "points"})
-    assert len(elements_dict) > 0, "elements_dict must not be empty"
-    assert any(
-        len(elements) > 0 for elements in elements_dict.values()
-    ), "elements_dict must contain at least one dict which contains at least one element"
-    to_keep = np.zeros(len(table), dtype=bool)
-    region_key = table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY_KEY]
-    instance_key = table.uns[TableModel.ATTRS_KEY][TableModel.INSTANCE_KEY]
-    instances = None
-    for _, elements in elements_dict.items():
-        for name, element in elements.items():
-            if get_model(element) == Labels2DModel or get_model(element) == Labels3DModel:
-                if isinstance(element, DataArray):
-                    # get unique labels value (including 0 if present)
-                    instances = da.unique(element.data).compute()
-                else:
-                    assert isinstance(element, DataTree)
-                    v = element["scale0"].values()
-                    assert len(v) == 1
-                    xdata = next(iter(v))
-                    # can be slow
-                    instances = da.unique(xdata.data).compute()
-                instances = np.sort(instances)
-            elif get_model(element) == ShapesModel:
-                instances = element.index.to_numpy()
-            elif get_model(element) == PointsModel:
-                instances = element.compute().index.to_numpy()
+
+    def _validate_elements_dict(elements_dict: dict[str, dict[str, Any]]) -> None:
+        assert set(elements_dict.keys()).issubset({"images", "labels", "shapes", "points"})
+        assert len(elements_dict) > 0, "elements_dict must not be empty"
+        assert any(
+            len(elements) > 0 for elements in elements_dict.values()
+        ), "elements_dict must contain at least one dict which contains at least one element"
+
+    def _get_table_keys(table: AnnData) -> tuple[str, str]:
+        return (
+            table.uns[TableModel.ATTRS_KEY][TableModel.REGION_KEY_KEY],
+            table.uns[TableModel.ATTRS_KEY][TableModel.INSTANCE_KEY],
+        )
+
+    def _get_element_instances(element: SpatialElement) -> ArrayLike | None:
+        if get_model(element) in [Labels2DModel, Labels3DModel]:
+            if isinstance(element, DataArray):
+                instances = da.unique(element.data).compute()
             else:
-                continue
-            indices = ((table.obs[region_key] == name) & (table.obs[instance_key].isin(instances))).to_numpy()
-            to_keep = to_keep | indices
+                assert isinstance(element, DataTree)
+                v = element["scale0"].values()
+                assert len(v) == 1
+                xdata = next(iter(v))
+                instances = da.unique(xdata.data).compute()
+            return np.sort(instances)
+        if get_model(element) == ShapesModel:
+            return element.index.to_numpy()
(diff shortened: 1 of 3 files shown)


melonora:
Hi @giovp, yes 2 tables can map to the same element, that is no problem.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -163,42 +163,60 @@ def _filter_table_by_elements(
...
-    if table is None:
+
+    def _validate_elements_dict(elements_dict: dict[str, dict[str, Any]]) -> None:
+        assert set(elements_dict.keys()).issubset({"images", "labels", "shapes", "points"})
+        assert len(elements_dict) > 0, "elements_dict must not be empty"
+        assert any(
+            len(elements) > 0 for elements in elements_dict.values()
+        ), "elements_dict must contain at least one dict which contains at least one element"
+
+    def _get_table_keys(table: AnnData) -> tuple[str, str]:
```
melonora, line 174: What is the reason for not using `get_table_keys` in `models`?

src/spatialdata/_core/query/relational_query.py
```diff
@@ -163,42 +163,60 @@ def _filter_table_by_elements(
...
+                assert isinstance(element, DataTree)
+                v = element["scale0"].values()
+                assert len(v) == 1
+                xdata = next(iter(v))
+                instances = da.unique(xdata.data).compute()
+            return np.sort(instances)
+        if get_model(element) == ShapesModel:
+            return element.index.to_numpy()
+        if get_model(element) == PointsModel:
+            return element.compute().index.to_numpy()
```
melonora, line 194: Same here but for `get_element_instances` in `relational_query`.

melonora (changes requested):
Thanks for the PR @giovp. I would like a little bit more clarification why certain functions seem to have been recreated here as I think it duplicates some code.

Also, I think it would be good to be a bit more explicit in the docstrings of how this would differ from the `join` functions as in both we now have the terminology of filtering without specifying how it differs. I know it is a private function here, but good to have it stated I think.

src/spatialdata/_core/query/relational_query.py
giovp: done

src/spatialdata/_core/query/relational_query.py
giovp: done

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -135,9 +135,15 @@ def test_filter_by_coordinate_system(full_sdata: SpatialData) -> None:
...
-    adata = full_sdata["table"]
+    adata = full_sdata["table"].copy()
+
+    circles_instances = full_sdata["circles"].index.values
+    poly_instances = full_sdata["poly"].index.values
+
+    adata = adata[: len(circles_instances) + len(poly_instances), :].copy()
+    adata.obs["annotated_shapes"] = ["circles"] * len(circles_instances) + ["poly"] * len(poly_instances)
+    adata.obs["instance_id"] = np.concatenate([circles_instances, poly_instances])
+
```
giovp, line 146: this test had quite a big bug. Basically, the table natively from conftest annotates labels, but here it was re used to annotate shapes and circles. Now, both shapes and circles have 5 instances only, and so the table was being filtered only by coordinate system, but this meant that the table had the first five instances mapping to the poly/circles, but then all the other instances also present, which did not map to anything. This is because the filtering was happening with the now removed `filter_table_by_element_names` which wasn't checking for that.

I will add test so that the filter_table function always return correct tables.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -58,31 +58,6 @@ def get_element_annotators(sdata: SpatialData, element_name: str) -> set[str]:
     return table_names
 
 
-def _filter_table_by_element_names(table: AnnData | None, element_names: str | list[str]) -> AnnData | None:
```
giovp, line 61: this function is buggy, and it doesn't make sense with the new multiple table design, as it can return wrong tables. I removed it and refactored it in the other filter method. I will add test for this. See other comment

tests/core/operations/test_spatialdata_operations.py
LucaMarconato: Thanks! Addressed in https://github.com/scverse/spatialdata/pull/1131

src/spatialdata/_core/query/relational_query.py
```diff
@@ -145,8 +122,8 @@ def _(
 
 # TODO: replace function use throughout repo by `join_sdata_spatialelement_table`
 def _filter_table_by_elements(
-    table: AnnData | None, elements_dict: dict[str, dict[str, Any]], match_rows: bool = False
-) -> AnnData | None:
+    table: AnnData | list[AnnData], elements_dict: dict[str, dict[str, Any]], match_rows: bool = False
```
LucaMarconato, line 125: The new PR https://github.com/scverse/spatialdata/pull/1131 replaces the old implementation of `_filter_table_by_elements()` with a much simpler implementation based on `join_spatialelement_table()`.

src/spatialdata/_core/spatialdata.py
```diff
@@ -735,10 +735,17 @@ def _filter_tables(
...
                     assert element_names is not None
-                    table = _filter_table_by_element_names(table, element_names)
+                    elements_dict = {}
+                    for element_type in ["images", "labels", "shapes", "points"]:
+                        elements = getattr(self, element_type)
+                        if elements:  # Check if the dictionary is not empty
+                            elements_dict[element_type] = {
+                                name: elements[name] for name in element_names if name in elements
+                            }
+                    table = _filter_table_by_elements(table, elements_dict=elements_dict)
```
LucaMarconato, line 748: Looks good! https://github.com/scverse/spatialdata/pull/1131 implements this change now.

src/spatialdata/_core/query/relational_query.py
LucaMarconato: Good point, https://github.com/scverse/spatialdata/pull/1131 removes the function as you suggested.

LucaMarconato:
Fully implemented in https://github.com/scverse/spatialdata/pull/1131 😊 Closing this PR, thanks again @giovp.

PR #702, opened 2024-09-08
3 and 4 channels default to rgb

LucaMarconato:
The default parameter `rgb=False` in `to_spatial_image()` ([code here](https://github.com/spatial-image/spatial-image/blob/8bc3c54d6d8595f6ba217e97cf5e35f4716ab5bc/spatial_image.py#L1149)) implies that if we have an RGB image with no channels specified, the image will be parsed by calling its channels `'0', '1', ...`. This means that napari-spatialdata will show the image as separating channels and using grayscale coloring.

This PR sets `rgb=None`, which will set the default channel names to `'r', 'g', 'b'` (`'a'`) when `c_coords` is not set, leading to correct visualization in `napari-spatialdata`.

### DIFF
--- a/src/spatialdata/models/models.py
+++ b/src/spatialdata/models/models.py
@@ -186,6 +186,8 @@ class RasterSchema(DataArraySchema):
                 ) from e
 
         # finally convert to spatial image
+        if "rgb" not in kwargs:
+            kwargs["rgb"] = None
         data = to_spatial_image(array_like=data, dims=cls.dims.dims, **kwargs)
         # parse transformations
         _parse_transformations(data, transformations)


LucaMarconato:
Small PR, merging.

LucaMarconato:
the failing tests make it seems like that `r`, `g`, `b` is being set even when a 3-channel image had channel names already set.

melonora (approved):
Preapproving but please add a warning for the user. Typically with cyclic imaging you are dealing with 4 markers per imaging cycle which in that case should not be displayed as rgb. With this PR you think more of the people using visium. Either is fine for me but the behaviour should be clear.

melonora (changes requested):
nvm tests will have to be fixed as well

LucaMarconato:
@melonora I think it should not be a problem because (after the tests are fixed), the RGB(A) will be used only if `c_coords` is not passed. In the case of 4-channel marker images the user is expected to pass the channel name.

LucaMarconato:
In the end since we are now passing `rgb=None` manually in `spatialdata-io`, I prefer (at least for now) to close this PR. I made a new PR instead that improves the docs: https://github.com/scverse/spatialdata/pull/716.

If user asks for automatically interpreting images as RGB(A) (for instance when adding external images), we can re introduce the approach of this PR and fix the tests.

Issue #404, opened 2023-11-23
Consider adding `.uns` or `.attrs` to `SpatialData`


LucaMarconato:
Consider adding metadata to the `SpatialData` level. This has been asked by @quentinblampey and will also be required for storing coordinate transformations for this refactoring: https://github.com/scverse/spatialdata/milestone/2

PR #711, opened 2024-09-23
Adding `attrs` at the `SpatialData` object level

quentinblampey:
As described in issue #404, it is useful to have attributes at the `SpatialData` object level. This will also be useful in Sopa, to simplify the API.

### Example Usage
```python
import spatialdata

sdata = blobs()

sdata.attrs # empty dictionnary
sdata.attrs["test"] = 12

sdata.write("blobs.zarr") # this saves the attributes on disk

sdata = spatialdata.read_zarr("blobs.zarr") # the attributes are read normally

sdata.attrs["test"] # returns 12
```

What do you think @LucaMarconato?

### Check-list

- [x] Add `attrs` attribute to the `SpatialData` object
- [x] Can write the `attrs` on disk when the data is not backed
- [x] Can write the attrs on disk when the data is already backed
- [x] Can read the `attrs` from disk
- [x] Pass the `attrs` to new SpatialData objects
- [x] Merge conflict keys in `attrs` during concatenation
- [x] Update the format version (see Luca's comment)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning][].
 
 ## [0.x.x] - 2024-xx-xx
 
+### Major
+
+-   Added attributes at the SpatialData object level (`.attrs`)
+
 ### Minor
 
 -   Added `clip: bool = False` parameter to `polygon_query()` #670
--- a/src/spatialdata/_core/_deepcopy.py
+++ b/src/spatialdata/_core/_deepcopy.py
@@ -47,7 +47,8 @@ def _(sdata: SpatialData) -> SpatialData:
     elements_dict = {}
     for _, element_name, element in sdata.gen_elements():
         elements_dict[element_name] = deepcopy(element)
-    return SpatialData.from_elements_dict(elements_dict)
+    deepcopied_attrs = _deepcopy(sdata.attrs)
+    return SpatialData.from_elements_dict(elements_dict, attrs=deepcopied_attrs)
 
 
 @deepcopy.register(DataArray)
--- a/src/spatialdata/_core/concatenate.py
+++ b/src/spatialdata/_core/concatenate.py
@@ -4,11 +4,12 @@ from collections import defaultdict
 from collections.abc import Iterable
 from copy import copy  # Should probably go up at the top
 from itertools import chain
-from typing import Any
+from typing import Any, Callable
 from warnings import warn
 
 import numpy as np
 from anndata import AnnData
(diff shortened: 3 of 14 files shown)


LucaMarconato:
- [ ] before merging we need to update the format version; we probably should add a format for the `SpatialData` container; something asked also by several users (we already have a format for the single elements, which would be independent from the format of the container).

quentinblampey:
Hi @LucaMarconato, have you started working on your last comment? If not, do you want me to have a look (although I'm not exactly sure to understand what you want to do)?

quentinblampey:
I noticed we are missing a feature: after we save a SpatialData object, if we add a new attr, there is currently no way to save it on disk. Something like `sdata.write_element` for the attrs

Marius1311:
Just a random note on this: I agree that having SpatialData - level attributes would be super helpful. Beyond that, I think it would be great to be able to annotate coordinate systems - these often correspond to samples, conditions, time-points etc. I can of course annotate all elements within a coordinate system through tables, but I think it would be more direct to somehow directly provide metadata for an entire coordinate system.

quentinblampey:
I also noticed that we want to pass the `attrs` inside certain functions, e.g. when performing a query (the `attrs` should be kept). I made a commit for this

The only concern is when we concatenate multiple SpatialData objects, how should we merge the duplicate keys (if any)?

LucaMarconato:
@quentinblampey thanks for the update. I haven't updated the format yet, we can work on it this week at the hackathon. Regarding saving to disk, I suggest to extend the `write_metadata()` method. Finally, regarding `.attrs`, we could proceed as `AnnData` [does](https://anndata.readthedocs.io/en/latest/concatenation.html#merging) for `.uns` when calling `ad.concat()`.

timtreis:
Extremely interested in this, thanks @quentinblampey !!

LucaMarconato:
@quentinblampey I have finalized this PR, I'm going to merge it now. 

Two comments.
1. **root-level versioning**: On disk, I am now adding a version number for the root level metadata, and also I am adding the string of the spatialdata software version that was used to create the metadata (this helps reproducibility). When reading the `.zattrs` from disk into the `SpatialData.attrs` slow I am not parsing the two version strings not to clutter the object.
2. **`.attrs` passed by reference**: I have removed the code `dict(sdata.attrs)` and passing the attrs always by reference. The only exceptions are when I call `spatialdata.deepcopy()`, where I call `copy.deepcopy()`, and when the user passes a `Mapping` that is not a `dict`. In such case the casting to dict performs a shallow copy.

Issue #713, opened 2024-09-26
Add default argument `shortest_path: bool=True` in `get_transformation_between_coordinate_systems()`


LucaMarconato:
CC @quentinblampey 

We should add the argument mentioned above since otherwise the user faces an error (prompting to choose between different transformation paths) also in cases when this is not needed. `shortest_path=False` is for rarer use cases.

PR #714, opened 2024-09-26
Added `shortest_path` parameter to `get_transformation_between_coordinate_systems`

quentinblampey:
Related to issue #713, CC @LucaMarconato

Note that I changed the structure of the code, which was highly nested. The logic should be the same though
I also added some tests!


LucaMarconato (approved):
LGTM thanks!

PR #720, opened 2024-10-01
Improved `concatenate()` for non-unique element names

### LINKED ISSUE #541: Feedback on concatenate()
While I in the end was able to concatenate the data the way I like, the user experience wasn't as great as I had hoped, so wanted to drop some feedback. As I'm not that familiar with spatialdata yet, it might be that there are already better solutions -- please let me know if there are. 

### Starting situation
I have ~20 Visium Cytassist samples from a clinical trial processed with nf-core/spatialtranscriptomics (using the https://github.com/nf-core/spatialtranscriptomics/pull/67 branch that already uses spatialdata). The pipeline generates a single `.zarr` folder for each sample. 

### Desired outcome
I would like to have all samples in a single SpatialData object. The AnnData table should contain the gene expression from all samples. 

### Pain points

 * `sd.concatenate` enforces that the input is a list. Is there a reason this can't accept any `Sequence` type (e.g. `dict_values`)?
 * Usually, I pass a dictionary `sample_id -> AnnData` to `anndata.concat`, which nicely makes unique obs_names in combination with `concat(..., index_unique="_")`. This doesn't work with spatialdata.concatenate, which leaves me with either manipulating the `obs_names` for each object before concatenation, or ugly obs names with numeric sufficies (e.g. `AACTCAACCTTGACCA-1_0_0_0_0_0_0_0_0_0_0_0_0_0_0_0`). IMO it would be great to support a dict as input to spatialdata.concatenate, too. 
 * The per-sample SpatialData objects all have the same names for images, shapes and coordinate systems. I currently rename them like this: 
   ```python
   sdatas_vis = {}
  
   for _, row in tqdm(samplesheet.iterrows(), total=samplesheet.shape[0]):
      sample = row["sample"]
      tmp_sd = sd.read_zarr(sample_path / sample / "data" / "sdata_processed.zarr")
      tmp_sd.tables["table"].obs = tmp_sd.tables["table"].obs.assign(**row)
      tmp_sd.tables["table"].obs["region"] = sample
      tmp_sd.tables["table"].uns["spatialdata_attrs"]["region"] = sample
      # rename images
      tmp_sd.images[f"{sample}_hires"] = tmp_sd.images["visium_hires_image"]
      tmp_sd.images[f"{sample}_lowres"] = tmp_sd.images["visium_lowres_image"]
      del tmp_sd.images["visium_hires_image"]
      del tmp_sd.images["visium_lowres_image"]
      # rename shapes
      tmp_sd.shapes[f"{sample}"] = tmp_sd.shapes["visium"]
      del tmp_sd.shapes["visium"]
  
      sdatas_vis[sample] = tmp_sd
   ```
    
    which seems a bit cumbersome. I'm wondering if there's a better solution or what's the intended way of handling such cases. It could also be worth adding a process to the nf-core/spatialtranscriptomics pipeline that already does the concatenation step.

LucaMarconato:
closes #541

### DIFF
--- a/tests/core/operations/test_spatialdata_operations.py
+++ b/tests/core/operations/test_spatialdata_operations.py
@@ -11,7 +11,14 @@ from spatialdata._core.data_extent import are_extents_equal, get_extent
 from spatialdata._core.operations._utils import transform_to_data_extent
 from spatialdata._core.spatialdata import SpatialData
 from spatialdata.datasets import blobs
-from spatialdata.models import Image2DModel, Labels2DModel, PointsModel, ShapesModel, TableModel, get_table_keys
+from spatialdata.models import (
+    Image2DModel,
+    Labels2DModel,
+    PointsModel,
+    ShapesModel,
+    TableModel,
+    get_table_keys,
+)
 from spatialdata.testing import assert_elements_dict_are_identical, assert_spatial_data_objects_are_identical
 from spatialdata.transformations.operations import get_transformation, set_transformation
 from spatialdata.transformations.transformations import (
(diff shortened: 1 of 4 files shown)


tests/core/operations/test_spatialdata_operations.py
```diff
@@ -284,6 +291,32 @@ def test_concatenate_sdatas(full_sdata: SpatialData) -> None:
     assert len(list(concatenated.gen_elements())) == 3
 
 
+@pytest.mark.parametrize("concatenate_tables", [True, False])
+@pytest.mark.parametrize("obs_names_make_unique", [True, False])
+def test_concatenate_sdatas_from_iterable(concatenate_tables: bool, obs_names_make_unique: bool) -> None:
```
grst, line 296: I'd probably add a test case that concatenating an iterable of length 1 still works and returns the identity... it's a commonly overlooked edge case.

grst (approved):
Looks pretty good to me, I just suggest one additional test case. 
I don't yet know when I'll have a chance to try it as I'll be travelling most of next week.

PR #726, opened 2024-10-05
Improve performance of writing pyramids

### LINKED ISSUE #577: Optimize writing performance for `MultiscaleSpatialImage`
As observed by @ArneDefauw, unnecessary loading operations are performed when calling `write_multiscale()` on a list of lazy tensors derived from `to_multiscale()`.

Optimizing the order in which the data is computed and written to disk, so to avoid the loading of the same chunks 2+ times, would probably lead to a drastic performance improvement, up to 10-fold.

aeisenbarth:
Closes #577 

- When doing multiple computations in Dask, it is better to submit them at once so that Dask can optimize the computational graph, re-use intermediate results and avoid recomputations.
- This capability was added to ome-zarr-py in [#257](https://github.com/ome/ome-zarr-py/pull/257) as a flag `compute=True` that can optionally be disabled to control the computation on your own. However, SpatialData kept the default enabled.
- As a consequence, every pyramid level was computed separately, and had to read again the whole input image from disk (if disk-backed). Due to this, the performance decreased with the number of scale factors:
  $$n := \textrm{number of pixels of scale level 0}$$
  $$f := \textrm{number of scale factors}$$
  $$\mathcal{O}\biggl(n \cdot (f+1)\biggr)$$

By disabling immediate computation, this PR collects all delayed objects (of pyramid levels) and lets Dask compute them at once. Input images now only need to be read once. The number of pixel read operations has now computational complexity of $$\mathcal{O}(n)$$

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning][].
 -   Added `shortest_path` parameter to `get_transformation_between_coordinate_systems`
 -   Added `get_pyramid_levels()` utils API
 -   Improved ergonomics of `concatenate()` when element names are non-unique #720
+-   Improved performance of writing images with multiscales #577
 
 ## [0.2.3] - 2024-09-25
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ docs = [
 test = [
     "pytest",
     "pytest-cov",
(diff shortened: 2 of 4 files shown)


LucaMarconato (approved):
Thanks for the PR, it looks great! I really like the test comparing the number of read-write operations.

PR #756, opened 2024-11-05
fix for to_polygons when using processes instead of threads in dask

ArneDefauw:
I noticed that when configuring dask to use 'processes' instead of 'threads' , `spatialdata.to_polygons` fails due to this line:

https://github.com/scverse/spatialdata/blob/27bb4a7579d8ff7cc8f6dd9b782226cb984ceb20/src/spatialdata/_core/operations/vectorize.py#L212

result of `dask.compute() `is lost when using processes, which is expected behaviour.

Note that using 'processes' instead of 'threads' considerably speeds up vectorizing labels for large masks, because the function we try to parallelize does not release the GIL.

Example:

```
import os
from pathlib import Path

import pooch
from pooch import Pooch

BASE_URL = "https://objectstor.vib.be/spatial-hackathon-public/sparrow/public_datasets"

def _get_registry(path: str | Path | None = None) -> Pooch:
    return pooch.create(
        path=pooch.os_cache("sparrow") if path is None else path,
        base_url=BASE_URL,
        version="0.0.1",
        registry={
            "transcriptomics/vizgen/mouse/_sdata_2D.zarr.zip": "e1f36061e97e74ad131eb709ca678658829dc4385a444923ef74835e783d63bc",
        },
    )

registry=_get_registry( path = None ) # set path if you want to download data to somewhere else
unzip_path = registry.fetch("transcriptomics/vizgen/mouse/_sdata_2D.zarr.zip", processor=pooch.Unzip())
sdata = read_zarr(os.path.commonpath(unzip_path))
sdata.path = None
```

```
import dask
from spatialdata import to_polygons

dask.config.set(scheduler="processes")
gdf=to_polygons( sdata[ "segmentation_mask_full" ] )
# finishes in around 3m locally on a mac m2

dask.config.set(scheduler="threads")
gdf=to_polygons( sdata[ "segmentation_mask_full" ] )
# finishes in around 8m locally on a mac m2
```

"segmentation_mask_full' contains the masks from a merscope experiment, around 300k labels.

### DIFF
--- a/src/spatialdata/_core/operations/vectorize.py
+++ b/src/spatialdata/_core/operations/vectorize.py
@@ -161,6 +161,13 @@ def to_polygons(data: SpatialElement, buffer_resolution: int | None = None) -> G
     """
     Convert a set of geometries (2D labels, 2D shapes) to approximated 2D polygons/multypolygons.
 
+    For optimal performance when converting rasters (:class:`xarray.DataArray` or :class:`datatree.DataTree`)
+    to polygons, it is recommended to configure `Dask` to use 'processes' rather than 'threads'.
+    For example, you can set this configuration with:
+
+    >>> import dask
+    >>> dask.config.set(scheduler='processes')
+
     Parameters
     ----------
     data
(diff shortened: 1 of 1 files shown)


LucaMarconato:
@ArneDefauw I am trying this, it looks great thank you!

Only one question. Why do you set `sdata.path = None` in your code example?

ArneDefauw:
> @ArneDefauw I am trying this, it looks great thank you!
> 
> Only one question. Why do you set `sdata.path = None` in your code example?

Hi @LucaMarconato , thanks for having a look!

I typically set `sdata.path` to None if I do unit tests, in order not to overwrite the .zarr store in the registry, but it is unrelated to this fix.

LucaMarconato (approved):

Issue #298, opened 2023-06-14
New multiple table in-memory design


LucaMarconato:
# Scope of the proposal
After receiving feedback from `@EliHei2 @lopollar @melonora @LLehner @AdemSaglamRB`, hearing users asking for multiple table support as an important requirement for the adoption of `SpatialData` in their tools (`Maks, Louis`) and having personally examined the development implications of the current table specification, I propose a new **in-memory** design for supporting multiple tables in `SpatialData`, that I have been maturing in the past months. Importantly, **the on-disk table specification will not change**.

## Timeline and feedback
I propose to implement the new in-memory representation after the following work: aggregation and query refactoring (me), io refactoring (@giovp), new pipeline for daily tasting all the notebook (me), synthetic datasets for R/JS interoperability (me), napari bugfix (me and others). So ideally we would start in 1-1.5 month, enough time for receiving feedback and improving the design. In particular I kindly ask for feedback to @kevinyamauchi @ivirshup @gtca @ilia-kats. I put some usernames in `raw` not to spam them now, but I will tag everybody after the first draft is refined, and additionally `@timtreis @berombau @joshmoore`.

## Disk specification
The disk storage will remain the one described in https://github.com/ome/ngff/pull/64. What I propose to change is the in-memory representation of the data: instead of directly representing in-memory the metadata that we have on-disk, I propose to put the data in a form that is more ergonomic and that follows the `(geo)pandas`, `anndata` and `muon` indexing common practices.

## In-memory representation
I propose the following.
* To stop having one single table shared for the `SpatialData` object and instead allowing, for each object, to have either an `AnnData` table, either a `MuData` table associated (or also having no table associated to it). For keeping the implementation smaller, I would start just with one `AnnData` table, and allow to use also a `MuData` table in a follow up PR. On disk, the `MuData` table would be saved as a list of tables, so that the current specs are still valid.
* To not represent `region`, `region_key`, `instance_key` in-memory, and instead relying on the same shared indexing between the table and the (single) corresponding shapes/labels element. Code examples below. This may require allowing `AnnData` to use custom indices and not just string (or at least integers), or would force the (geo)dataframes to have string indices. We will find a way around this.
[...]

PR #765, opened 2024-11-06
Partial reading

### LINKED ISSUE #457: Tolerance when reading corrupted data

aeisenbarth:
This PR implements modes for `read_zarr` that allow to continue reading SpatialData elements even if some elements caused read errors, for example due to **corrupted files** (crash during write, a race condition or copy errors). This is useful, because with growing datasets (by number of elements), a corruption of a single file can make all other elements inaccessible. The ability to read at least the unaffected data helps for data recovery.

This is implemented by wrapping element-specific reader calls into a **context manager** `handle_read_errors`. If enabled with `on_bad_files="warn"`, it converts **exceptions into warnings**, and continues execution. Since execution is not aborted, all subsequent code within an iteration must be wrapped in the context manager to avoid that it is executed after a handled error. In order to identify which element failed to be read, the warning message is prepended with its location, e.g.:
```text
shapes/blobs_polygons: JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
At the end, a SpatialData object is always returned, containing only the successfully read elements.

The current implementation is at the granularity level of whole elements, that means if anything in an element causes an error, the whole element is skipped. In principle it could be made more fine-granular, e.g. returning an element with just invalid attributes skipped, or returning a table with unreadable columns skipped. Especially for [large single tables](https://github.com/scverse/spatialdata/issues/298) with annotations for all elements, this would be very beneficial (although they are column-based, so that a corrupted file would destroy a column's values for all annotations). However, the relevant code is not contained in this repository, but requires changes in [`anndata.read_zarr`](https://github.com/scverse/anndata/blob/e63cb0a40d4da26794e93ae9dd910d8919287987/src/anndata/_io/zarr.py#L53). To avoid losing annotations due to file corruption, I would recommend not to use a single table, but one table per SpatialData element.

For tests, see `test_read_zarr_with_error` and `test_read_zarr_with_warnings`.

Closes #457 

# Release notes
```md
-   Added option `on_bad_files="warn"` to `read_zarr` for partially reading corrupted datasets #457
```


LucaMarconato:
Thanks @aeisenbarth, great PR and very well structured tests! I made some [very minor adjustments here](https://github.com/scverse/spatialdata/pull/765/commits/56583db10ee7738d0bcad4003f6ac075d937bb58): I made a test more strict and replaced `shutil.rmtree()` with another approach; with the other devs we opted to minimize/avoid calls to `rmtree()`, for possible safety concerns.

PR #777, opened 2024-11-12
add warning when running bounding box query if points is present

BioinfoTongLI:


src/spatialdata/_core/query/spatial_query.py
```diff
@@ -514,6 +514,11 @@ def _(
     min_coordinate = _parse_list_into_array(min_coordinate)
```
berombau, line 514: Add remark on performance with points to function documentation in [Notes](https://numpydoc.readthedocs.io/en/latest/format.html#notes) header

LucaMarconato (approved):

Issue #750, opened 2024-11-01
Improving ergonomics for channel names handling


LucaMarconato:
Brainstormed together with @minhtien-trinh. 

Working with channel names in multi-channel images in not straightforward. In particular:
- [x] the `c_coords` argument in `Image2DModel.parse()` is too hidden; we should move it out of `kwargs`
- [ ] we could consider adding an helper function `set_channels()` (to complement `get_channels()`)
- [ ] `SpatialData.write_metadata()` currently doesn't handle channel names, so it is not possible to easily overwrite the channel names metadata without rewriting the heavy raster data.

melonora:
just waiting for review @minhtien-trinh

minhtien-trinh:
Looks great to me, thank you for handling this @melonora

PR #779, opened 2024-11-13
Move c_coords out of kwargs

melonora:
This PR implements one of the tasks in #750 to improve the handling of channel names. It moves the `c_coords` parameter out of kwargs and also adds a validation for a mismatch in length between `c_coords` and the length of the `c` dimension.

### DIFF
--- a/src/spatialdata/datasets.py
+++ b/src/spatialdata/datasets.py
@@ -40,7 +40,7 @@ def blobs(
     n_shapes: int = 5,
     extra_coord_system: str | None = None,
     n_channels: int = 3,
-    c_coords: ArrayLike | None = None,
+    c_coords: str | list[str] | None = None,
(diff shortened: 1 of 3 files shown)


melonora:
still need to see how we can have `c_coords` only in docstring when dealing with images and not labels.

LucaMarconato:
> still need to see how we can have c_coords only in docstring when dealing with images and not labels.

Does this work now?

Btw, looks great, pre-approving. Only missing thing, please add some tests (testing that it works for images and testing that it gives an error for labels, and for images when there is a mismatch). Then feel free to merge!

LucaMarconato (approved):

melonora:
Main problem is that the docstring must be a literal so you can't use f strings or anything.

PR #786, opened 2024-11-13
change channel names

### LINKED ISSUE #750: Improving ergonomics for channel names handling

melonora:
Closes #750 

This PR allows for either changing the channel_names by means of a static method on a image `SpatialElement` or in case of an element in the `Spatialdata` object a method taking as parameter the element name.

Currently, still need to implement writing purely the metadata.

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -23,7 +23,12 @@ from xarray import DataArray, DataTree
 from spatialdata._core._elements import Images, Labels, Points, Shapes, Tables
 from spatialdata._logging import logger
 from spatialdata._types import ArrayLike, Raster_T
-from spatialdata._utils import _deprecation_alias, _error_message_add_element
+from spatialdata._utils import (
+    _assign_multiscale_coords,
+    _check_match_length_channels_c_dim,
+    _deprecation_alias,
+    _error_message_add_element,
+)
 from spatialdata.models import (
     Image2DModel,
     Image3DModel,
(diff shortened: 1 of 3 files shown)


src/spatialdata/_core/spatialdata.py
```diff
@@ -315,6 +320,52 @@ def get_instance_key_column(table: AnnData) -> pd.Series:
             return table.obs[instance_key]
         raise KeyError(f"{instance_key} is set as instance key column. However the column is not found in table.obs.")
 
+    @staticmethod
```
LucaMarconato, line 323: Three comments:
- [ ] I would move this function to `spatialdata.models._utils`, because `get_channels()` is there
- [ ] I would rename this function to `set_channels()` to be consistent with the old naming
- [ ] Please also update the `api.md` file and the `changelog` file

src/spatialdata/_core/spatialdata.py
```diff
@@ -315,6 +320,52 @@ def get_instance_key_column(table: AnnData) -> pd.Series:
...
+            if isinstance(element, DataArray):
+                element = element.assign_coords(c=channel_names)
+            else:
+                element = element.map_over_datasets(_assign_multiscale_coords, {"c": channel_names})
+        else:
+            raise TypeError(f"Model `{model}` does not support setting channel names.")
+
+        return element
+
+    def set_sdata_image_channel_names(self, element_name: str, channel_names: str | list[str]) -> None:
```
LucaMarconato, line 352: I would remove this method because we started following the philosophy of separating the modification of data in-memory and the on-disk write, as with transformations. 

So now the workflow would be to call `set_channels()` (or `set_channel_names()` depending on how we call it) and then manually call `sdata.write_channel_names()` or `sdata.write_metadata()`.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Actually I like the name `set_channel_names()` more than `set_channels()`, so 
- [ ] I would rename `get_channels()` to `get_channel_names()`.

src/spatialdata/_utils.py
```diff
@@ -311,3 +313,43 @@ def _error_message_add_element() -> None:
...
+    cls_dims
+        The dimensions of the particular `ImageModel`
+
+    Returns
+    -------
+    c_coords
+        The channel names as list
+    """
+    c_index = cls_dims.index("c")
+    c_length = (
```
LucaMarconato, line 339: I'd add a quick safety check that all the scales of the data-tree have the same channel dim.

tests/io/test_metadata.py
```diff
@@ -111,11 +111,22 @@ def test_save_transformations_incremental(element_name, full_sdata, caplog):
...
-@pytest.mark.skip(reason="Not implemented yet")
 def test_save_channel_names_incremental(images: SpatialData) -> None:
-    # note: the non-incremental IO for channel names is already covered in TestReadWrite.test_images(), so here we
-    # only test the incremental IO
-    pass
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        f0 = os.path.join(tmp_dir, "sdata.zarr")
+        images.write(f0)
+
+        new_channels = ["first", "second", "third"]
```
LucaMarconato, line 119: after removing the method `set_sdata_image_channel_names()`, this test needs to use `set_channel_names()` followed by `sdata.write_channel_names()`.

LucaMarconato:
Looks great! Three further minor comments:
- [ ] after renaming `get_channels()` to `get_channel_names()`, we need to:
	- [ ] update the docs
	- [ ] call automatically `get_channel_names()` from `get_channels()` and give a deprecation warning
- [ ] we need to call `write_channel_names()` from `write_metadata()` (this method needs to have an extra argument `channel_names: bool=True`
- [ ] we need to update the test for `write_metadata()` to consider the channel names

LucaMarconato (approved):

src/spatialdata/_core/spatialdata.py
melonora: I would still have this as a method, but then indeed separating the in memory modification and the write on disk and adding an argument to this function write which is a bool

src/spatialdata/_utils.py
melonora: not relevant anymore as I had to switch to `get_axes_names`, which has the dispatch.

Issue #165, opened 2023-03-01
Offset when rasterizing data


LucaMarconato:
When rasterizing images or labels (basically resampling), there is an offset (see picture) that I think is due to the fact that somewhere in the code for rasterization the coordinates of the pixels are considered to be associated with the top left corner, and in other parts of the code for rasterization the coordinates refer to the center of the pixels.

This half pixel difference is not negligible since when we align together images of different resolutions, one pixel for one image can be many pixels for another.

The screenshot is from this example: https://github.com/scverse/spatialdata/blob/eb99aebbdb4f7b7ee69d9663c5c996f5fa8f0440/examples/dev-examples/spatial_query_and_rasterization.py#L15

![image](https://user-images.githubusercontent.com/2664412/222229276-5d0d4ba0-151d-4ed1-9a7c-bd9801cdd2bd.png)

LucaMarconato:
Another important point, the rasterization functions could be used for computing aggregation functions. Until this offset problem is addressed, the aggregation would be imprecise.

LucaMarconato:
The function `transform_to_data_extent()` (which is coming soon in a new PR) is affected by this bug. I had to create a test with permissive tolerance, which reads as `assert np.allclose(a, first_a, rtol=0.005)`.

- [ ] When this bug is addressed we should reduce the tolerance.

LucaMarconato:
The bug is not just of rasterize, but multiple factors are involved. First, napari considers the coordinates as the center of pixels. Here is for instance the effect of rotating an image by -45 degrees, and showing also the (0, 0) coordinate before and after rotation. Note that the rotation doesn't preserve the location of the (0, 0)-pixel top-left corner. 

I need to check, but I suspect that `ndinterp()` used in `transform()` and `rasterize()` may not operate as in napari, leading to some misalignment.

![image](https://github.com/scverse/spatialdata/assets/2664412/56083682-fc08-44f3-a911-feff1be07a7f)

LucaMarconato:
One of the reasons for the bug is the fact that (0, 0) is not the top-left corner but the center of the top-left pixel, as shown in the image above. I am fixing this in #453. Nevertheless, there is also another problem, apparently due to `ndinterp.affine_transform()`: https://github.com/scipy/scipy/issues/20038.

LucaMarconato:
After digging deeper into this I found out the causes of the bug, some of which can be fixed (as I did partially did in #453), but one of them requires a design choice to be made. I will explain this here.

[...]

PR #787, opened 2024-11-13
bug with 3d data test in transform_to_data_extent

LucaMarconato:
A bug with points is still present; it is a consequence of this bug https://github.com/scverse/spatialdata/issues/165. We should really close it.


vjcitn:
This PR solved my problem, so I advocate its adoption.  Thanks!

LucaMarconato:
Thanks @vjcitn, I finalized the tests (one of the cases required some extra checks). Ready to merge now!

PR #791, opened 2024-11-19
Fix `transform_to_data_extent` converting labels to images

### LINKED ISSUE #783: `rasterize` converts labels to images
**Describe the bug**

When applying `transform_to_data_extent` on a SpatialData with labels, they are merged into images.

**To Reproduce**

```python
sdata = blobs()
sdata_rasterized = transform_to_data_extent(sdata, coordinate_system="global", target_unit_to_pixels=1)
assert len(sdata_rasterized.labels) == len(sdata.labels)
```
```pycon
In [61]: sdata_rasterized
Out[61]: 
SpatialData object
├── Images
│     ├── 'blobs_image': DataArray[cyx] (3, 512, 512)
│     ├── 'blobs_labels': DataArray[cyx] (1, 512, 512)
│     ├── 'blobs_multiscale_image': DataArray[cyx] (3, 512, 512)
│     └── 'blobs_multiscale_labels': DataArray[cyx] (1, 512, 512)
├── Points
│     └── 'blobs_points': DataFrame with shape: (<Delayed>, 4) (2D points)
├── Shapes
│     ├── 'blobs_circles': GeoDataFrame shape: (5, 2) (2D shapes)
│     ├── 'blobs_multipolygons': GeoDataFrame shape: (2, 1) (2D shapes)
│     └── 'blobs_polygons': GeoDataFrame shape: (5, 1) (2D shapes)
└── Tables
      └── 'table': AnnData (26, 3)
with coordinate systems:
    ▸ 'global', with elements:
        blobs_image (Images), blobs_labels (Images), blobs_multiscale_image (Images), blobs_multiscale_labels (Images), blobs_points (Points), blobs_circles (Shapes), blobs_multipolygons (Shapes), blobs_polygons (Shapes)

```

**Expected behavior**
All element types should be preserved. Labels should stay in labels.

**Additional context**
This is caused by the following line:
https://github.com/scverse/spatialdata/blob/6694c91d6b23cdcc8a6e57b21bcbe25d9423763f/src/spatialdata/_core/operations/rasterize.py#L330-L331

The behavior is undocumented and not intuitive. As it was intentionally introduced, there seems to exist a use case, but in other use cases it is a potential source of bugs. Some control over it is already provided through the flag `return_regions_as_labels=False`.
- We should **document** it.
- We should rethink whether it is safer to change the **default** of `return_regions_as_labels`, especially if it is for internal use.
- We must either make sure all functions calling `rasterize` **pass the correct value** for `return_regions_as_labels` that is appropriate in each case, or further expose the flag to their callers.
  A possible fix would be to make `transform_to_data_extent` call `rasterize(…, return_regions_as_labels=True)` https://github.com/scverse/spatialdata/blob/6694c91d6b23cdcc8a6e57b21bcbe25d9423763f/src/spatialdata/_core/operations/_utils.py#L118-L128

aeisenbarth:
This PR ensures `transform_to_data_extent` preserves the data type of labels and does not convert them to images. Here, I chose the most minimal solution to pass the (missing) parameter `return_regions_as_labels=True` and override its default of `False`.

Meanwhile, @quentinblampey has already documented the parameter `return_regions_as_labels` in `rasterize`.

Closes #783

### DIFF
--- a/src/spatialdata/_core/operations/_utils.py
+++ b/src/spatialdata/_core/operations/_utils.py
@@ -125,6 +125,7 @@ def transform_to_data_extent(
                 target_width=target_width,
                 target_height=None,
                 target_depth=None,
+                return_regions_as_labels=True,
             )
             sdata_to_return_elements[element_name] = rasterized
         else:
--- a/tests/core/operations/test_spatialdata_operations.py
+++ b/tests/core/operations/test_spatialdata_operations.py
@@ -17,6 +17,7 @@ from spatialdata.models import (
     PointsModel,
     ShapesModel,
     TableModel,
+    get_model,
     get_table_keys,
 )
 from spatialdata.testing import assert_elements_dict_are_identical, assert_spatial_data_objects_are_identical
@@ -468,6 +469,7 @@ def test_transform_to_data_extent(full_sdata: SpatialData, maintain_positioning:
         for element in elements:
             before = full_sdata[element]
             after = sdata[element]
+            assert get_model(after) == get_model(before)
             data_extent_before = get_extent(before, coordinate_system="global")
             data_extent_after = get_extent(after, coordinate_system="global")
             # huge tolerance because of the bug with pixel perfectness


aeisenbarth:
CI is failing, probably due to a mismatch of Zarr and AnnData versions ("cannot import name 'Array' from 'zarr.core'").

LucaMarconato (approved):

PR #794, opened 2024-11-25
Preserve points feature_key in queries

quentinblampey:
In bounding_box/polygon queries, the `"feature_key"` of points was not preserved.

- [x] preserve feature_key in bounding_box
- [x] preserve feature_key in polygon
- [x] tests added

While fixing that, I also found (and fixed) a minor issue in the `PointsModel.parse` method when passing `feature_key`. Indeed, if corresponding column is already a category, it is dropped from the dataframe, see the line below:
```python
if feature_key is not None and data[feature_key].dtype.name != "category":
    table[feature_key] = data[feature_key].astype(str).astype("category")
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning][].
 ### Fixed
 
 -   Updated deprecated default stages of `pre-commit` #771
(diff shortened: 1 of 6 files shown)


LucaMarconato:
Looks great to me thanks! Fixing the docs and then merging.

PR #818, opened 2025-01-05
add support for layer selection of get_values method

Sonja-Stockhaus:
We need the method in `spatialdata-plot` and I'm currently adding the possibility to color shapes/points/labels by values stored in a different layer of a table (https://github.com/scverse/spatialdata-plot/issues/326). For that, we would need this functionality here as well.


LucaMarconato (approved):

LucaMarconato:
Looks great thanks! I added a small test, now auto-merging.

PR #819, opened 2025-01-05
Prettier mismatch between local and remote

LucaMarconato:
The local prettier makes changes that are rejected by the prettier online.


LucaMarconato:
May be due to this: https://github.com/prettier/prettier/issues/8782

LucaMarconato:
I could not find a solution so I am temporarily ignoring markdown files. Do you also experience this problem @timtreis @melonora @Sonja-Stockhaus? Discussion also started here: https://scverse.zulipchat.com/#narrow/channel/316218-repo-management/topic/prettier.20local.20vs.20remote.20mismatch/near/491979762.

LucaMarconato:
I switched to the mirror suggested here https://github.com/scverse/cookiecutter-scverse/issues/283#issuecomment-2368879513, and it fixes the issue.

PR #841, opened 2025-01-22
Docker automatic build

LucaMarconato:
This PR:
- improves the `Dockerfile`
- Replaces the approach drafted here https://github.com/scverse/spatialdata/pull/326
- introduces a workflow for automatically building a Docker daily, as follows:
	- the container includes the latest available versions of `spatialdata` (with `torch`), `spatialdata-io`, `spatialdata-plot`. Not `napari-spatialdata`
	- to trigger the build of the container you need to manually run the GitHub action. If things work as expected (let's test it for a few releases), we can uncomment the part that runs the workflow everyday
	- the docker container will contain all the 3 spatialdata-xx versions in the tag, and available here: https://github.com/scverse/spatialdata/pkgs/container/spatialdata/versions
	- running the workflow everyday will ensure that the docker container is using up-to-date versions and will not cause unnecessary computation. In fact, the workflow will first check if the container that is going to be built is using a combination of versions for which the container already exists, in that case the build will be skipped. Example of a [workflow run that skips the docker build steps](https://github.com/scverse/spatialdata/actions/runs/12909422601/job/35997051434?pr=841).

To build the Docker container locally:
```bash
# this is for Apple Silicon machines, if you are not using such machine you can omit the --build-arg
docker build --build-arg TARGETPLATFORM=linux/arm64 --tag spatialdata .
docker run -it spatialdata
```

To run one container that was built by the GitHub Actions:
```bash
docker pull ghcr.io/scverse/spatialdata:spatialdata0.3.0_spatialdata-io0.1.7_spatialdata-plot0.2.9
docker run -it ghcr.io/scverse/spatialdata:spatialdata0.3.0_spatialdata-io0.1.7_spatialdata-plot0.2.9
```

CC @FloWuenne @migueLib @melonora

### DIFF
--- a/.github/workflows/build_image.yml
+++ b/.github/workflows/build_image.yml
@@ -85,7 +85,8 @@ jobs:
 
             - uses: docker/build-push-action@v5
               if: ${{ env.skip_build == 'false' }}
-              IMAGE_TAG: ${{ env.REGISTRY }}/scverse/spatialdata:${{ env.IMAGE_TAG_SUFFIX }}
+              env:
+                  IMAGE_TAG: ${{ env.REGISTRY }}/scverse/spatialdata:${{ env.IMAGE_TAG_SUFFIX }}
               with:
                   context: .
                   file: ./Dockerfile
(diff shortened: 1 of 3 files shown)


LucaMarconato:
I'll update the docs and then we can merge.

Issue #805, opened 2024-12-09
How to open the current Zarr based image data as OME-Zarr


tischi:
As discussed in https://github.com/scverse/spatialdata/issues/803, the zarr based image data that is currently stored with spatialdata is not compatible with the current OME-Zarr spec and can thus not readily be opened with the [current community eco-system](https://link.springer.com/article/10.1007/s00418-023-02209-1) for supporting OME-Zarr.

For simplicity I will call the currently saved image data SD-Zarrr (short for SpatialData-Zarr). 
I would like to explore whether we could somehow make it possible to open SD-Zarr as OME-Zarr.

One idea that I have is to add another folder next to the `images` folder and call this extra folder `ome-zarr`. In that folder there would be sub-folders for the images stored in the `images` folder. These subfolders would only contain a `.zgroup` and `.zattrs` with a valid OME-Zarr `multiscales` where the paths to the resolution levels in the `multiscales` would then be, e.g. `../../images/img0/0`. I don't know whether this would work? @bogovicj @joshmoore @bugraoezdemir

LucaMarconato:
# Intro

> As discussed in https://github.com/scverse/spatialdata/issues/803, the zarr based image data that is currently stored with spatialdata is not compatible with the current OME-Zarr spec and can thus not readily be opened with the [current community eco-system](https://link.springer.com/article/10.1007/s00418-023-02209-1) for supporting OME-Zarr.

In the following discussion I want to keep the context forefront to the additional tagged people, so I will repeat some points from the linked discussion. Also below I report what we wrote in the OME-Zarr paper linked above, with regards to the SpatialData framework:

> SpatialData [...] uses the ome-zarr-py readers and writers for raster data (images and labels) and implements features of the **upcoming version of the NGFF specification**, including tables and transformations. Furthermore, it experimentally supports the representation of additional modalities commonly found in spatial omics data, such as points (e.g. transcripts locations) and shapes (e.g. spatial transcriptomics circular array capture locations and generic polygonal ROIs). 

In SpatialData by design we chose not to be compatible with OME-Zarr 0.4, but what should have been "OME-Zarr 0.5" (=transformations)[^1]. The hope was that we would have never needed to be compatible with OME-Zarr 0.4 because the ecosystem would have embraced the transformations specs. Unfortunately this is not yet the case, so below I explore some possibilities for this.

[...]

PR #849, opened 2025-01-28
Specify NGFF 0.6-dev in output data

LucaMarconato:
This PR monkeypatches `ome_zarr.format` to make it possible to use the read/write APIs of `ome-zarr-py`, while clarifying which OME-Zarr data version we are building on top.

Precisely, we implement the [NGFF transformations specification](https://github.com/ome/ngff/pull/138), which, as of today, it is not merged to NGFF main yet. The timeline is to have the specification included in NGFF v0.6, hence from this PR we will write `0.6-dev` instead of `0.4` (which was kept for compatibility with `ome-zarr-py`).

All the tests pass, the code for reproducing the documentation and paper notebooks and the data conversions from `spatialdata-sandbox` are currently running.

Practically, the PR will not change APIs or usage of the function of the `spatialdata` package and will only modify, on-disk, the version string for images and labels, as shown here (in this example I used Seqfish v2 data).

```
(ome311) macbook@Lucas-MacBook-Pro-2021 seqfish_v2_io % diff -r data.zarr.v4 data.zarr.v6
diff -r data.zarr.v4/images/Roi1_DAPI/.zattrs data.zarr.v6/images/Roi1_DAPI/.zattrs
91c91
<             "version": "0.4"
---
>             "version": "0.6-dev"
diff -r data.zarr.v4/labels/Roi1_Segmentation/.zattrs data.zarr.v6/labels/Roi1_Segmentation/.zattrs
3c3
<         "version": "0.4"
---
>         "version": "0.6-dev"
71c71
<             "version": "0.4"
---
>             "version": "0.6-dev"
diff -r data.zarr.v4/zmetadata data.zarr.v6/zmetadata
105c105
<                     "version": "0.4"
---
>                     "version": "0.6-dev"
157c157
<                 "version": "0.4"
---
>                 "version": "0.6-dev"
225c225
<                     "version": "0.4"
---
>                     "version": "0.6-dev"
```

CC @joshmoore

### DIFF
--- a/src/spatialdata/_io/format.py
+++ b/src/spatialdata/_io/format.py
@@ -47,9 +47,7 @@ def _parse_version(group: zarr.Group, expect_attrs_key: bool) -> str | None:
 
 
 class SpatialDataFormat(CurrentFormat):
-    @property
-    def version(self) -> str:
-        return "0.6-dev"
+    pass
 
 
 def format_implementations() -> Iterator[Format]:
@@ -138,6 +136,22 @@ class RasterFormatV01(SpatialDataFormat):
     @property
     def spatialdata_format_version(self) -> str:
         return "0.1"
+
+    @property
+    def version(self) -> str:
+        return "0.4"
+
+
+class RasterFormatV02(RasterFormatV01):
+    @property
+    def spatialdata_format_version(self) -> str:
+        return "0.2"
+
+    @property
+    def version(self) -> str:
+        # 0.1 -> 0.2 changed the version string for the NGFF format, from 0.4 to 0.6-dev-spatialdata as discussed here
+        # https://github.com/scverse/spatialdata/pull/849
+        return "0.6-dev-spatialdata"
 
 
 class ShapesFormatV01(SpatialDataFormat):
@@ -230,7 +244,7 @@ class TablesFormatV01(SpatialDataFormat):
             raise ValueError("`table.obs[instance_key]` must not contain null values, but it does.")
 
 
-CurrentRasterFormat = RasterFormatV01
+CurrentRasterFormat = RasterFormatV02
 CurrentShapesFormat = ShapesFormatV02
 CurrentPointsFormat = PointsFormatV01
 CurrentTablesFormat = TablesFormatV01
@@ -248,6 +262,7 @@ TablesFormats = {
 }
 RasterFormats = {
     "0.1": RasterFormatV01(),
+    "0.2": RasterFormatV02(),
 }
 SpatialDataContainerFormats = {
     "0.1": SpatialDataContainerFormatV01(),
--- a/src/spatialdata/_io/format.py
+++ b/src/spatialdata/_io/format.py
@@ -50,21 +50,6 @@ class SpatialDataFormat(CurrentFormat):
     pass
 
 
-def format_implementations() -> Iterator[Format]:
-    """Return an instance of each format implementation, newest to oldest."""
-    yield SpatialDataFormat()
-    yield FormatV04()
-    yield FormatV03()
-    yield FormatV02()
-    yield FormatV01()
-
-
-# monkeypatch the ome_zarr.format module to include the SpatialDataFormat (we want to use the APIs from ome_zarr to
-# read, but signal that the format we are using is a dev version of NGFF, since it builds on some open PR that are
-# not released yet)
-ome_zarr.format.format_implementations = format_implementations
-
-
 class SpatialDataContainerFormatV01(SpatialDataFormat):
     @property
     def spatialdata_format_version(self) -> str:
@@ -151,7 +136,7 @@ class RasterFormatV02(RasterFormatV01):
     def version(self) -> str:
         # 0.1 -> 0.2 changed the version string for the NGFF format, from 0.4 to 0.6-dev-spatialdata as discussed here
         # https://github.com/scverse/spatialdata/pull/849
-        return "0.6-dev-spatialdata"
+        return "0.4-dev-spatialdata"
 
 
 class ShapesFormatV01(SpatialDataFormat):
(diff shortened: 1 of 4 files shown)


src/spatialdata/_io/format.py
```diff
@@ -45,7 +47,24 @@ def _parse_version(group: zarr.Group, expect_attrs_key: bool) -> str | None:
 
 
 class SpatialDataFormat(CurrentFormat):
-    pass
+    @property
+    def version(self) -> str:
+        return "0.6-dev"
```
joshmoore, line 52: cc @bogovicj in case defining a specific pre-release version (i.e. dev1 or beta1) would make detecting differences in these upcoming versions more reliable.

See also [ongoing conversation around 0.5.2](https://github.com/ome/ngff/pull/293)

joshmoore: Actually, going back and looking at the diff output in the description, I'd suggest we *not* use 0.6 unless this also implements the Zarr v3 work of 0.5.

LucaMarconato: This is a good point. Any suggestion on the string to use? Maybe something more verbose like `0.4-dev-transformations`?

LucaMarconato: Or `0.4-dev-spatialdata	`.

joshmoore: * :+1: for starting with `0.4-`
 * :+1: for having "spatialdata" or similar in it
 * you might think about whether or not you have a version field from SD itself
 * I defer to you if you want "dev" in it or not (I don't think the validator, e.g., has special behavior for that. @will-moore?)

will-moore: No, the validator currently just uses the version string to construct URLs for loading the schemas. So it'll need a little patching to parse the "0.4" out of a longer string if it's going to validate those.

LucaMarconato: Thanks @joshmoore. 
> you might think about whether or not you have a version field from SD itself
The SpatialData is versioned here: https://github.com/scverse/spatialdata/blob/a7e9e991c31474c7495b17a5eb8544d7d2eba108/src/spatialdata/_io/format.py#L233

@will-moore I think your "no" answers to "(I don't think the validator, e.g., has special behavior for that. @will-moore?)". Do you have any preference for the `0.4-` string?

will-moore: I can handle `0.4-*` anything is fine.

I tried looking at some spatialdata data in the validator with...

```
$ ome_zarr view mouse_liver_data.zarr
```

 ...looking at the `/labels/segmentation_mask/` the validator is complaining that `/multiscales/0/coordinateTransformations/0` doesn't have `scales` or `translation` objects.
Instead you have `"input":{"axes":[...]}, "output": {"axes":[...]}, "type": "identity"}` which is heading towards the CoordinateSystems RFC-5 https://ngff.openmicroscopy.org/rfc/5/index.html but not quite the same.

LucaMarconato: Yes, I will adjust the coordinate transformations metadata together with @melonora in `ome-zarr-models-py`.

LucaMarconato: If you agree, I'd go then with `0.4-dev-spatialdata` and merge this PR.

LucaMarconato: Thanks Will, but as Josh said here:
> Except this version explicitly does not validate, so perhaps it's enough that it's just not the known "0.4"?
Not passing the validation is expected because the coordinate transformation specification that we implemented was based on an earlier draft and our plan is to finalize it in `ome-zarr-models-py`.

In this PR I made a patch to `ome-zarr-py` to be able to read the tensors even with partially incorrect metadata. We indeed handle the read/write of metadata in `spatialdata`, and use `ome-zarr-py` to read the tensors.

What would be needed here is to agree on an appropriate version number, such as `0.4-dev-spatialdata`, that doesn't imply that validation is passed.

LucaMarconato: Please in this comment you can see the full context of why we want to fix the version string now and we defer passing the validation later: https://github.com/scverse/spatialdata/issues/805#issuecomment-2531706080.

will-moore:
As I understand it from #805, the eventual goals are:
 - SpatialData writes valid OME-NGFF data according to the CoordinateSystems spec (RFC-5) - likely to be v0.6
 - There is some way to convert this into OME-NGFF v0.4 so it can be read by more tools (that don't yet support v0.6)

The main issues are:
 - The RFC-5 is not yet approved, so we don't yet have v0.6 spec. What to do in the meantime?
 - The OME-NGFF v0.5 spec uses Zarr v3, and we still don't have support for that in `ome-zarr-py`. See https://github.com/ome/ome-zarr-py/pull/404. However, https://github.com/thewtex/ngff-zarr/ does support v0.5
 

Re: your patch for `ome-zarr-py` that allows it to handle partially incorrect metadata - Are you proposing to open a PR with that change (and for it to be able to recognise custom version numbers and be more lenient in validation)?

Is this what you're asking for agreement on? I don't see a problem with that cc @joshmoore ?
But maybe easier to discuss on an actual `ome-zarr-py` PR?

LucaMarconato:
Correct @will-moore, thanks for the clear summary.

> The RFC-5 is not yet approved, so we don't yet have v0.6 spec. What to do in the meantime?

I prefer to direct my efforts in implementing the NGFF RFC-5 in `ome-zarr-models-py`, so that we align with that (i.e. import it and remove the transformations code from `spatialdata`) as soon as possible. In the meantime I would leave `ome-zarr-py` as it is, meaning that if someone tries to read the metadata an error is given, since `0.4-dev-spatialdata` is not supported by it. At the same time, thanks to the current PR we will be able to continue using `ome-zarr-py` internally. And hopefully we will be able to replace the `0.4-dev-spatialdata` with `0.6` soon (ideally just some months). When this happens, I think we could contribute to making `ome-zarr-py` handle RFC-5 (which should be easy by then since one could then depend on `ome-zarr-models.py`).

joshmoore:
Sorry for the wait. I have not come up with anything better than `0.4-dev-spatialdata`. It's obviously not a great situation, but rather than spending in more of our time on it, :+1: for trying to get a stable NGFF-0.6-based version that we can all build on. (Transformation scripts, etc. may be something to consider a bit down the road.)

LucaMarconato:
Thanks @joshmoore. I updated the version string and bumped the spatialdata version for the raster elements to reflect the change in the underlying NGFF format written. This allows the user to write in the old version if needed for backward compatibility (tests added for this).

The work on ome-zarr-models-py continues!

Issue #821, opened 2025-01-07
Unable to save subsetted Xenium `sdata` to `.zarr`


christinedien:
The bug can be reproduced using one of the spatialdata example datasets: I had used the Xenium Aligned Breast Cancer [dataset](https://s3.embl.de/spatialdata/spatialdata-sandbox/xenium_rep1_io_aligned.zip)

**Describe the bug**
After subsetting a spatialdata object using `sdata.query.bounding_box(...)`, attempting to save the new subset as a `.zarr` sometimes results in an irregular chunking error. This occurs after a `sdata.write(...)` command

**To Reproduce**
Code to reproduce the behavior:

```
import spatialdata as sd
import spatialdata_io
import squidpy as sq
import spatialdata_plot
```
```
sdata = sd.read_zarr("/path/to/data_aligned.zarr/")
```

The following subset calls **DO NOT** result in an error
```
subset = sdata.query.bounding_box(
    axes=["x", "y"],
    min_coordinate=[0,0],
    max_coordinate=[15000,13000],
    target_coordinate_system="global",
)

subset.write(f'../data/xenium/subset_test/subset.zarr', overwrite=True)
```

```
subset3 = sdata.query.bounding_box(
    axes=["x", "y"],
    min_coordinate=[0,0],
    max_coordinate=[20000,13000],
    target_coordinate_system="global",
)

subset3.write(f'../data/xenium/subset_test/subset3.zarr', overwrite=True)
```

The following subsetting examples **DO** result in an error:
```
subset2 = sdata.query.bounding_box(
    axes=["x", "y"],
    min_coordinate=[15000,0],
    max_coordinate=[35000,13000],
    target_coordinate_system="global",
)

subset2.write(f'../data/xenium/subset_test/subset2.zarr', overwrite=True)
```

```
subset4 = sdata.query.bounding_box(
    axes=["x", "y"],
    min_coordinate=[10000,0],
    max_coordinate=[30000,13000],
    target_coordinate_system="global",
)
subset4.write(f'../data/xenium/subset_test/subset4.zarr', overwrite=True)
```

I initially thought maybe the subset size may have something to do with the irregular chunking. In this case the query size (20,000 x 13,000) in `subset2` and `subset4` resulted in an error but subset3 has the same query size and was able to be saved.

In my private dataset, only 1/12 of my queries is savable

**Expected behavior**
I would have expected the subsetted `spatialdata` object to be savable to a `.zarr`

**Screenshots**

[...]

PR #861, opened 2025-02-04
Fix rechunk bug after spatial query

LucaMarconato:
Closes https://github.com/scverse/spatialdata/issues/821 until the root of the problem (https://github.com/dask/dask/issues/11713) is addressed, and until we can update to the latest dask (https://github.com/dask/dask/issues/11146).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,7 +127,7 @@ exclude = [
     "docs/_build",
     "dist",
     "setup.py",
-    
+
 ]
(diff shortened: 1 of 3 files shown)


LucaMarconato:
This PR also checks for the regularity of the chunks after the query, so a potential bug is found before writing to disk. We could consider adding this check into the validation functions. This last idea is tracked here: https://github.com/scverse/spatialdata/issues/862.

Issue #812, opened 2024-12-17
Error (Codec does not support buffers > 2GB) when trying to save zarr directory


kotr98:
Hi,

i have experienced an error when trying to load the output of 10x xenium into a spatialdata object, attaching a HE image, and trying to save it as an zarr files. I iterated over many samples with the same loop and it works for most of them, so the code per se works. However, for some samples i get the error 

ValueError: Codec does not support buffers of > 2147483647 bytes

when i try to save the sdata object.

Has anyone else experienced this problem before? How can i fix it so i get the zarr files for all my samples?

Thanks in adance,
Korbinian

I used this script to produce the error:

```
import spatialdata_io as sdio

sdata = sdio.xenium("spatial_data/outs")

he_image_datatree = sdio.xenium_aligned_image(
    image_path="HE.tif",
    alignment_file=None,
    dims=None  
)

sdata.images['he_image'] = he_image_datatree

sdata.write("test.zarr")
```

ckmah:
The HE image is larger than the default maximum. Try this:

```
from PIL import Image
Image.MAX_IMAGE_PIXELS = None
```

kotr98:
Thanks for the fast reply. Unfortunately it did not resolve the error:

I can reproduce it with this data sample: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM7990532 
I downloaded the GSM7990532_warped_THD0008-TMA3_HE.tif.gz HE image and GSM7990532_output-XETG00048__0003392__THD0008__20230313__191400.tar.gz and subsequently unzipped them. 

```
import spatialdata_io as sdio
from PIL import Image
Image.MAX_IMAGE_PIXELS = None


sdata = sdio.xenium("relabel_output-XETG00048__0003392__THD0008__20230313__191400/outs")

he_image_datatree = sdio.xenium_aligned_image(
    image_path="GSM7990532_warped_THD0008-TMA3_HE.tif",
    alignment_file=None,
    dims=None  
)

sdata.images['he_image'] = he_image_datatree

sdata.write("test.zarr")
```

ckmah:
I would try to manually set the chunk sizes, assuming your H&E images is shaped [3, y, x], something like:

```python
...

he_image_datatree = sdio.xenium_aligned_image(
    image_path="GSM7990532_warped_THD0008-TMA3_HE.tif",
    alignment_file=None,
    dims=["c", "y", "x"] ,
[...]

PR #874, opened 2025-02-11
Add warning (raster models) when large chunks are detected

LucaMarconato:
Closes: #812 

FYI: @timtreis


LucaMarconato:
There is a mismatch between the local precommits and the ones in the CI likely introduced with https://github.com/scverse/spatialdata/pull/851. We need to fix it. CC @Zethson

PR #877, opened 2025-02-12
Replace pip with uv for test workflow

LucaMarconato:
Giving a try to using `uv` instead of `pip` for the test workflow. I did't edit the release and docker workflows since they are run rarely and trying to make them faster would not pay off.

I took inspiration from https://github.com/vitessce/vitessce-python/pull/390/files. Note that I removed the `pip` caching mechanism in this commit fac784813019f16119da7bbd733b2736d77f33c7. It could be that having `uv` + caching leads to better results.


LucaMarconato:
`uv` without caching is already faster! 
- with `uv`: (around 30s) https://github.com/scverse/spatialdata/actions/runs/13289302024/job/37105692929?pr=877
- with `pip` + caching: (around 2m) https://github.com/scverse/spatialdata/actions/runs/13270423075/job/37048312124

I'll give a try to `uv` + caching now.

LucaMarconato:
I'm having problems with optional dependencies. It seems that they are not correctly cached.

LucaMarconato:
Ok I made it work. Caching doesn't help with pre-built wheels, re-downloading every time is as fast (also explained here https://github.com/astral-sh/setup-uv#disable-cache-pruning).

I will disable the cache. This PR will still bring a 4-fold decrease in the time required to setup the env.

LucaMarconato:
I tried using `uv` instead of `pip` in `readthedocs` using the approach described here: https://github.com/astral-sh/uv/issues/10074, but it didn't seem to speed up things (strange tbh, but ok, 2 min is already fast compared to the time to run tests).
- `pip` run: https://readthedocs.org/projects/scverse-spatialdata/builds/27168628/
- `uv` run: https://readthedocs.org/projects/scverse-spatialdata/builds/27168689/

I therefore went back to `pip` for readthedocs.

LucaMarconato:
CC @timtreis (worth moving to `uv` also in `spatialdata-plot` etc). It's very easy to move to `uv` (I did many experiments with caching here, but in the end what I recommend is what shown in the PR diff).

quentinblampey:
Hi @LucaMarconato, do you think it would also be worth it to move to `uv` in `pyproject.toml` (not just in the CI)?
I also saw that two dependencies may be unused: `pooch` and `xarray_spatial`, is that correct?

I can have a look if you want

quentinblampey:
Oh it's because `xarray-spatial` is named `xrspatial`, sorry I missed that

melonora:
can look into it, as for uv, I would actually like to experiment with automatically creating the docker container using pixi. Pixi seems to support this.

PR #883, opened 2025-02-24
pin numba to earlierst version supporting python 3.10

giovp:
installing with `uv pip` I got errors if I didn't explicitly pin numba


LucaMarconato (approved):
LGTM thanks!

Issue #739, opened 2024-10-31
Benchmark spatial partitioning in dask geopandas


LucaMarconato:

berombau:
Spatial partitioning in Shapely https://shapely.readthedocs.io/en/2.0.6/strtree.html

Issue #233, opened 2023-04-27
Geopandas vs pandas for points


LucaMarconato:
Points now use lazy dataframes (Dask `DataFrame`). We talked about allowing having in-memory both as dataframes and lazy dataframes. https://github.com/scverse/spatialdata/issues/153

What about using `GeoDataFrame` and Dask `GeoDataFrame` instead? This will allow for:
- lazy loading
- unify points and circles (circles are in reality points in which we add a radius column via the schema) https://github.com/scverse/spatialdata/issues/46
- spatial index
- the user has to convert to `GeoDataFrame` anyway to exploit the `geopandas` functions, these functions are cumbersome (see below), but even worst, the user may be tempted or may expect to save points in the `SpatialData` object as `GeoDataFrame`.

Drawbacks:
- performance could not be good enough because it seems that `GeoDataFrame` is creating one Python object per row, we should aim at 500M-1B points.
- more complex type for the user
- 3D only partially supported by `geopandas`, we need to understand the implication. Maybe it's fine to have the user only being able to do queries with `geopandas` for the 2D component of the data and use our APIs for 3D queries. In the end, if we use Dask dataframes we still need to implement these queries, so at worst we can extra queries APIs for free from `geopandas`.

Functions to convert back and forth points represented as Dask dataframes and `geopandas` dataframes.

```
from dask.dataframe.core import DataFrame as DaskDataFrame
from geopandas import GeoDataFrame


def points_dask_dataframe_to_geopandas(points: DaskDataFrame) -> GeoDataFrame:
    # let's ignore the z component here
    points_gdf = GeoDataFrame(geometry=geopandas.points_from_xy(points["x"], points["y"]))
    for c in points.columns:
        points_gdf[c] = points[c]
    return points_gdf


def points_geopandas_to_dask_dataframe(gdf: GeoDataFrame) -> DaskDataFrame:
    # convert the GeoDataFrame to a Dask DataFrame
    ddf = dd.from_pandas(gdf[gdf.columns.drop("geometry")], npartitions=1)
    ddf["x"] = gdf.geometry.x
    ddf["y"] = gdf.geometry.y
    # parse
    ddf = PointsModel.parse(ddf, coordinates={"x": "x", "y": "y"})
    return ddf
```

LucaMarconato:
### Comment for the Basel hackathon
[...]

PR #913, opened 2025-03-25
Geopandas benchmarks

aeisenbarth:
This is a work-in-progress for testing performance of [`geopandas`](https://geopandas.org/en/stable/) compared to [`dask-geopandas`](https://dask-geopandas.readthedocs.io/en/stable/).
I used the `asv` framework to add benchmarks for geopandas [nearest](https://geopandas.org/en/stable/docs/reference/api/geopandas.sindex.SpatialIndex.nearest.html) and [query](https://geopandas.org/en/stable/docs/reference/api/geopandas.sindex.SpatialIndex.query.html).

Unfortunately, dask-geopandas doesn't implement these functions yet. However, this code can be used as a starting point to continue performance tests.

Note also that the latest version of dask-geopandas is only compatible with newer dask dependencies.

Refers to #233, #739

### DIFF
--- a/benchmarks/spatialdata_benchmark.py
+++ b/benchmarks/spatialdata_benchmark.py
@@ -72,3 +72,61 @@ class TimeQueries:
             target_coordinate_system="global",
             filter_table=filter_table,
         )
+
+
+class TimeGeopandasQuery:
+    params = (
+        [100, 1000, 10000],  # TODO: test for larger number of points
+        ["geopandas", "dask_geopandas"],
+    )
+    param_names = ["num_objects", "lib"]
+    query_size = 100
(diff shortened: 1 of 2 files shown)


LucaMarconato:
Thanks @aeisenbarth for starting this work. 

> Note also that the latest version of dask-geopandas is only compatible with newer dask dependencies.

Now the conflicts with dask are resolved thanks to @melonora's PR.

Regarding this PR, I labeled as `backlog` because currently this PR is not prioritized. When working on lazy loading for shapes it can be picked up again.

LucaMarconato:
I'll actually close: we can follow-up on these experiments in `spatialdata-sandbox`.

PR #921, opened 2025-04-02
Account for the padding shift during vectorization + use multipolygon

quentinblampey:
In `_vectorize_mask`, we pad the mask before using `find_contours`, which is needed to have correct geometries. Yet, it create a shift of one pixel in the coordinates.

This PR shifts the geometry of one pixel to recover the original coordinates. In the test that I added, when doing rasterize + vectorize, we now have an IoU of 0.98 instead of 0.92.

When one label has multiple geometries, I now use a MultiPolygon instead of multiple Polygon. This is more convenient and less ambiguous, as we'll have only one row per label from the mask.

### DIFF
--- a/src/spatialdata/_core/operations/vectorize.py
+++ b/src/spatialdata/_core/operations/vectorize.py
@@ -227,30 +227,30 @@ def _(
     return ShapesModel.parse(gdf, transformations=transformations.copy())
 
 
-def _region_props_to_polygons(region_props: RegionProperties) -> list[Polygon]:
+def _region_props_to_polygon(region_props: RegionProperties) -> MultiPolygon:
     mask = np.pad(region_props.image, 1)
     contours = skimage.measure.find_contours(mask, 0.5)
 
     # shapes with <= 3 vertices, i.e. lines, can't be converted into a polygon
     polygons = [Polygon(contour[:, [1, 0]]) for contour in contours if contour.shape[0] >= 4]
+    polygon = MultiPolygon(polygons) if len(polygons) > 1 else polygons[0]
 
     yoff, xoff, *_ = region_props.bbox
-    return [shapely.affinity.translate(poly, xoff, yoff) for poly in polygons]
+    return shapely.affinity.translate(polygon, xoff - 1, yoff - 1)  # account for the padding
 
 
 def _vectorize_mask(
     mask: np.ndarray,  # type: ignore[type-arg]
 ) -> GeoDataFrame:
     if mask.max() == 0:
-        return GeoDataFrame(geometry=[])
+        return GeoDataFrame({"label": []}, geometry=[])
 
     regions = skimage.measure.regionprops(mask)
 
-    polygons_list = [_region_props_to_polygons(region) for region in regions]
-    geoms = [poly for polygons in polygons_list for poly in polygons]
-    labels = [region.label for i, region in enumerate(regions) for _ in range(len(polygons_list[i]))]
-
-    return GeoDataFrame({"label": labels}, geometry=geoms)
+    return GeoDataFrame(
+        {"label": [region.label for region in regions]},
+        geometry=[_region_props_to_polygon(region) for region in regions],
+    )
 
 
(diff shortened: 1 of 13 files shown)


LucaMarconato:
Thanks for the PR! I leave my review.

src/spatialdata/_core/operations/vectorize.py
```diff
@@ -227,12 +227,15 @@ def _vectorize_chunk(chunk: np.ndarray, yoff: int, xoff: int) -> GeoDataFrame:
...
 
-def _region_props_to_polygon(region_props: RegionProperties) -> MultiPolygon:
+def _region_props_to_polygon(region_props: RegionProperties) -> MultiPolygon | Polygon:
     mask = np.pad(region_props.image, 1)
     contours = skimage.measure.find_contours(mask, 0.5)
 
     # shapes with <= 3 vertices, i.e. lines, can't be converted into a polygon
     polygons = [Polygon(contour[:, [1, 0]]) for contour in contours if contour.shape[0] >= 4]
+
+    assert len(polygons) > 0, "No valid polygon found in the region properties."
```
LucaMarconato: I could not reproduce this assertion because even when trying to with labels consisting of single pixels, they get converted to 4-sided polygons, for example this is with `_vectorize_mask(np.array([[0, 1], [1, 2]]))`:
image

LucaMarconato: To show the plot I used this simple code:
```python
import matplotlib.pyplot as plt
from matplotlib.patches import Polygon as MplPolygon

plt.figure()
plt.imshow(mask)
for region in regions:
    multi_poly = _region_props_to_polygon(region)
    if isinstance(multi_poly, Polygon):
        x, y = multi_poly.exterior.coords.xy
        coords = np.array(list(zip(x, y)))
        polygon = MplPolygon(coords, closed=True, fill=True, edgecolor='black',
                            facecolor='lightblue', alpha=0.5)
        ax = plt.gca()
        ax.add_patch(polygon)
        ax.set_xlim(-1, mask.shape[1] + 1)
        ax.set_ylim(-1, mask.shape[0] + 1)
    else:
        for polygon in multi_poly.geoms:
            x, y = polygon.exterior.coords.xy
            coords = np.array(list(zip(x, y)))
            polygon = MplPolygon(coords, closed=True, fill=True, edgecolor='black',
                                facecolor='lightblue', alpha=0.5)
            ax = plt.gca()
            ax.add_patch(polygon)
            ax.set_xlim(-1, mask.shape[1] + 1)
            ax.set_ylim(-1, mask.shape[0] + 1)
plt.show()
```

LucaMarconato: I modified the list comprehension by removing the `if ... shape[0] >= 4` and added a stricter `if ... raise` before. Also I removed the assertion. What do you think? If you know of an example leading to invalid polygons we care revert my change.

LucaMarconato: Last small comment, here I used a `if ... raise` instead of an `assert` because running the script with `-O` (optimized mode) removes all the assertions. Also `assert` lines, when not tested, escape from the codecov report.

LucaMarconato (approved):

LucaMarconato:
If you agree on my changes we can merge.

src/spatialdata/_core/operations/vectorize.py
quentinblampey: I also couldn't reproduce the `assert`: actually I also believe it may never happen, but I wanted to keep it for safety

quentinblampey:
Thanks @LucaMarconato, your change looks good!

src/spatialdata/_core/operations/vectorize.py
quentinblampey: Thanks also for the information about the `-O` mode, I didn't know that, it means I should use less `assert`

melonora: Typically never good to use assert except for tests or debug mode:)

PR #939, opened 2025-05-24
Add type hint when get a SpatialElement item

quentinblampey:
Currently, when we get an item by name, e.g. `geo_df = sdata.shapes["boundaries"]`, then the type of `geo_df` is not inferred and we lose type hints and auto-completion.

This PR aims to fix this, via the use of `TypeVar`, to precise the type of each element.

I also added `AnnData` to the return types of `SpatialData.__getitem__`


LucaMarconato:
LGTM thanks!

PR #946, opened 2025-06-30
Support modifying/filtering labels elements via join operations

selmanozleyen:
Hi @timtreis @ilan-gold ,

I am working on this [PR](https://github.com/scverse/squidpy/pull/1011) also and thought maybe it will be smart to have the code I use there supported here because they seem useful at first glance. So I wrote and tested [subset_sdata_by_table_mask](https://github.com/selmanozleyen/spatialdata/blob/feature/filter_operations_on_label/src/spatialdata/_core/query/relational_query.py#L1114) in this PR.

@LucaMarconato I can't ask for reviews other than @ilan-gold do you know why?

---

Some notes
- For `PointsModel` (`GeoDataFrame`) it assumes the index is `instance_id` always
- For `Label2DModel` the image is assumed to have the `instance_id`s as values themselves
- For `Label2DModel` when the element is a `xr.DataTree` it assumes the keys are the different scales
- I didn't use the relational operations to get the shapes and points for this reason because they assume the merge is on the element indices themselves
- I didn't add support for AnnData because we would have to also document the return types and stuff in the case of AnnData input which doesn't seem reasonable to me if its going to make it 1-1 same as `scanpy.pp.filter_cells`

Code excerpt from tests to demonstrate the usage:
```python
sdata = concatenate(
        {
            "labels": blobs_annotating_element("blobs_labels"),
            "shapes": blobs_annotating_element("blobs_circles"),
            "points": blobs_annotating_element("blobs_points"),
            "multiscale_labels": blobs_annotating_element("blobs_multiscale_labels"),
        },
        concatenate_tables=True,
    )
    third_elems = sdata.tables["table"].obs["instance_id"] == 3
    subset_sdata = subset_sdata_by_table_mask(sdata, "table", third_elems)

    labels_remaining_ids = set(np.unique(subset_sdata.labels["blobs_labels-labels"].data.compute())) - {0}
    assert labels_remaining_ids == {3}

    for scale in subset_sdata.labels["blobs_multiscale_labels-multiscale_labels"]:
        ms_labels_remaining_ids = set(
            np.unique(subset_sdata.labels["blobs_multiscale_labels-multiscale_labels"][scale].image.compute())
        ) - {0}
        assert ms_labels_remaining_ids == {3}

    points_remaining_ids = set(np.unique(subset_sdata.points["blobs_points-points"]["instance_id"].compute())) - {0}
    assert points_remaining_ids == {3}

    shapes_remaining_ids = set(np.unique(subset_sdata.shapes["blobs_circles-shapes"].index)) - {0}
    assert shapes_remaining_ids == {3}
```


src/spatialdata/_core/query/masking.py
```diff
@@ -0,0 +1,114 @@
...
+        x_size_next = labels_element[scales[i + 1]].image.shape[1]
+        y_factor = y_size_current / y_size_next
+        x_factor = x_size_current / x_size_next
+
+        scale_factors.append((y_factor, x_factor))
+
+    return scale_factors
+
+
+def filter_shapesmodel_by_instance_ids(element: ShapesModel, ids_to_remove: list[str]) -> GeoDataFrame:
```
ilan-gold: If this is meant to be public, should go into an `__init__.py` file and be documented but it's up to your judgement!

src/spatialdata/_core/query/relational_query.py
```diff
@@ -1019,3 +1021,132 @@ def get_values(
...
+        output_core_dims=[["y", "x"]],
+        vectorize=True,
+        dask="parallelized",
+        output_dtypes=[image.dtype],
+        dataset_fill_value=0,
+        dask_gufunc_kwargs={"allow_rechunk": True},
+    )
+
+    # Force computation to ensure the changes are materialized
+    computed_result = processed.compute()
```
timtreis: I know it's from my original PR but I'm not fully sure if it's needed - should do test because this can be very costly.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -1019,3 +1021,132 @@ def get_values(
...
+
+
+@_filter_by_instance_ids.register(DataArray)
+def _(element: DataArray, ids_to_remove: list[int], instance_key: str) -> DataArray:
+    del instance_key
+    return Labels2DModel.parse(_set_instance_ids_in_labels_to_zero(element, ids_to_remove))
+
+
+@_filter_by_instance_ids.register(DataTree)
+def _(element: DataArray | DataTree, ids_to_remove: list[int], instance_key: str) -> xr.DataArray | xr.DataTree:
```
timtreis: incorrect return type

tests/core/query/test_relational_query_subset_sdata_by_table_mask.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+import pytest
+
+from spatialdata import concatenate, subset_sdata_by_table_mask
+from spatialdata._core.query.relational_query import _filter_by_instance_ids
+from spatialdata.datasets import blobs_annotating_element
+
+
+def test_filter_labels2dmodel_by_instance_ids():
```
timtreis: Parametrise, don't loop over inputs

tests/core/query/test_relational_query_subset_sdata_by_table_mask.py
```diff
@@ -0,0 +1,80 @@
...
+    filtered_labels_element = _filter_by_instance_ids(labels_element, [2, 3], "instance_id")
+
+    for scale in labels_element:
+        filtered_ids = set(np.unique(filtered_labels_element[scale].image.compute())) - {
+            0,
+        }
+        preserved_ids = np.unique(labels_element[scale].image.compute())
+        assert filtered_ids == (set(all_instance_ids) - {2, 3})
+        # check if there is modification of the original labels
+        assert set(preserved_ids) == set(all_instance_ids) | {0}
```
timtreis: Use np.testing instead

src/spatialdata/_core/query/relational_query.py
```diff
@@ -1019,3 +1021,132 @@ def get_values(
...
+
+
+@_filter_by_instance_ids.register(DaskDataFrame)
+def _(element: DaskDataFrame, ids_to_remove: list[str], instance_key: str) -> DaskDataFrame:
+    return element[~element[instance_key].isin(ids_to_remove)]
+
+
+@_filter_by_instance_ids.register(DataArray)
+def _(element: DataArray, ids_to_remove: list[int], instance_key: str) -> DataArray:
+    del instance_key
```
timtreis: Why this?

timtreis (changes requested):
Can this PR make use of https://github.com/scverse/spatialdata/blob/7604a3d2325079293ff523c5dff4483dffc890cb/src/spatialdata/_core/query/relational_query.py#L787 and https://github.com/scverse/spatialdata/blob/7604a3d2325079293ff523c5dff4483dffc890cb/src/spatialdata/_core/query/relational_query.py#L757?

src/spatialdata/_core/query/relational_query.py
selmanozleyen: you mean why del instance_key? It is to explicitly clarify that I won't be using it for this dispatch

src/spatialdata/_core/query/relational_query.py
selmanozleyen: ok fixed

tests/core/query/test_relational_query_subset_sdata_by_table_mask.py
selmanozleyen: the loop here is for the multiple scales. Or did you mean something else?

src/spatialdata/_core/query/relational_query.py
selmanozleyen: ok done

selmanozleyen:
About using `def match_sdata_to_table(` and the other. I noticed for  shapes the index is assumed to be the `instance_id` but this doesn't match how the `blobs` are filled and would fail in the tests. It was documented that way so I assumed this was intentional but `match_sdata_to_table` wouldn't (and didn't) pass the current tests because it assumed the element index was the instance_id.

tests/core/query/test_relational_query_subset_sdata_by_table_mask.py
selmanozleyen: but these are all simple set comparisons. would have to order then compare the matrices which seems convoluted. But if you think the same still I can do it

src/spatialdata/_core/query/relational_query.py
```diff
@@ -1099,3 +1101,124 @@ def get_values(
...
+    # Create a new DataArray to ensure persistence
+    return xr.DataArray(
+        data=processed.data,
+        coords=image.coords,
+        dims=image.dims,
+        attrs=image.attrs.copy(),  # Preserve all attributes
+    )
+
+
+def _get_scale_factors(labels_element: DataTree) -> list[tuple[float, float]]:
```
LucaMarconato: partially redundant with `_get_scale()` and `_set_transformation()` defined in `spatialdata/transformations/_utils.py`. I will remove from here as now I removed `_filter_by_instance_ids` (replaced by `filter_by_table_query()`.

LucaMarconato:
I picked this up today. 

[The PR before today's rework](https://github.com/scverse/spatialdata/pull/946/changes/BASE..631fe2ad10bd2f5fca78b7beccbe7ed272376406) introduced a series of functions, some of them new, other redundant with other parts of the codebase. I will comment on them one by one.

- `_mask_block()` new, I kept it.
- `_set_instance_ids_in_labels_to_zero()` new. I kept it and added a check to detect the 3D case and throw a `NotImplementedError`
- `_get_scale_factors()`, redundant, removed. See https://github.com/scverse/spatialdata/pull/946/changes/BASE..631fe2ad10bd2f5fca78b7beccbe7ed272376406#r3219789350
- `_filter_by_instance_ids()` already available for points and shapes, new for labels. I removed the dispatch mechanism and called this function (after rework), `_filter_labels_element()`.
- `subset_sdata_by_table_mask()` the only public function of the PR. I removed this function as it was redundant with `match_sdata_to_table` and `filter_by_table_query` with the assumption of supporting right joins for labels. I now implemented right joins (calling `_set_instance_ids_in_labels_to_zero()` internally.

The initial purpose of the PR is preserved (you can see the tests). Therefore squidpy, which was blocked by merging this work, can build on this. The new syntax is:

```python
filter_by_table_query(
	sdata, "table", obs_expr=an.col("instance_id").is_in([1, 3, 4]), filter_label_pixels=True
)
```

Here below is a self-contained example (showing also how to use `match_sdata_to_table()`.
```python
# /// script                                                                                                                                                                                                                                                                   
# requires-python = ">=3.11"
# dependencies = [
#   "spatialdata @ git+https://github.com/scverse/spatialdata.git@8d6552ecf24a5b4e46b0c81267b18ab95c7e5660",
#   "annsel",
# ]
# ///

import annsel as an
import numpy as np

from spatialdata import match_sdata_to_table
from spatialdata._core.query.relational_query import filter_by_table_query
from spatialdata.datasets import blobs_annotating_element

sdata = blobs_annotating_element("blobs_labels")
keep_ids = [1, 3, 4]

# match_sdata_to_table: pre-filter the table to the desired instances
table = sdata.tables["table"]
subset_table = table[table.obs["instance_id"].isin(keep_ids)]
result_match = match_sdata_to_table(sdata, "table", table=subset_table, filter_label_pixels=True)

# filter_by_table_query: express the same filter as an annsel predicate
result_query = filter_by_table_query(
    sdata, "table", obs_expr=an.col("instance_id").is_in(keep_ids), filter_label_pixels=True
)

for name, result in [("match_sdata_to_table", result_match), ("filter_by_table_query", result_query)]:
    remaining = set(np.unique(result["blobs_labels"].data.compute())) - {0}
    print(f"{name}: remaining instance ids = {sorted(remaining)}")
```

Issue #799, opened 2024-12-03
Option to output GeoArrow-encoded parquet from geopandas


keller-mark:
**Is your feature request related to a problem? Please describe.**
Currently, spatialdata stores Shape elements on-disk using WKB-encoded parquet files, since this is the GeoPandas default of [geopandas.to_parquet](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_parquet.html) .

WKB encoding has a high overhead for decoding. For background:
- https://geoarrow.org/format.html#motivation
- https://observablehq.com/@kylebarron/geoparquet-on-the-web
- In Vitessce in JS, we have to iterate over each row of the parquet column and perform WKB decoding ([code](https://github.com/vitessce/vitessce/blob/bca77bd434001d6b765404ae4abf6871537741a7/packages/file-types/zarr/src/SpatialDataShapesSource.js#L246))

**Describe the solution you'd like**
Would it be possible to either:
- use `to_parquet(geometry_encoding='geoarrow')` as the default
- allow users to opt-in to using GeoArrow encoding (as opposed to WKB(

**Describe alternatives you've considered**

**Additional context**

LucaMarconato:
Thanks for reporting this. Do you have a estimate of the performance improvement that you would observed? We should also perform some benchmarks on our side.

PR #951, opened 2025-07-03
Geometry encoding parameter for shapes

keller-mark:
Fixes https://github.com/scverse/spatialdata/issues/799
The [to_parquet](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_parquet.html) function supports a `geometry_encoding` parameter. When [geoarrow](https://geoarrow.org/), it will be more efficient to read/parse the geometries, as the data can stay in its parquet/arrow memory layout during downstream usage. Visualization applications will benefit from this (and other applications such as data processing pipelines should too).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ node_modules/
 
 .mypy_cache
 .ruff_cache
+uv.lock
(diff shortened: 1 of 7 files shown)


LucaMarconato:
Hi @keller-mark, this is ready for review, correct?

keller-mark:
Yes, thanks for the reminder, I have just updated the status

PR #967, opened 2025-08-11
constrain xarray

LucaMarconato:
This PR removes the `xarray-dataclasses` dependency. The reason is that this library is not maintained anymore and both `spatial-image` and `multiscale-spatial-image` now use `xarray-dataclass` instead. This PR makes spatialdata follow the constraint that both these libraries have for `xarray-dataclass`. 

In addition, due to recent work on `spatial-image` and `multiscale-spatial-image` we can now support python version 3.13.


quentinblampey:
Thanks @melonora for the quick fix!

I have two small comments:
- As far as I understand, Python 3.13 can be used only from `multiscale_spatial_image==2.0.3`, correct? If so, since you updated the Python requirements in `spatialdata`, maybe we should force `multiscale_spatial_image>=2.0.3` instead of `multiscale_spatial_image>=2.0.2`?
- If we make the above update, then we could even simply remove the `xarray-dataclass` dependency from spatialdata, as it's already in the dependencies of `multiscale_spatial_image` (and also because it's not used directly in spatialdata anyway)

melonora:
@quentinblampey so for resolving yes, so I will change that. I had to make a few changes and thus release 2.0.3 but then found out that both already supported python 3.13 but we can constrain the version not to cause problems.

PR #968, opened 2025-08-14
use python>=3.10 typing conventions

quentinblampey:
Since `spatialdata` requires `python>=3.10`, we can remove most of the `from __future__ import annotations`, and also remove all the old typings such as `Union` or `Optional`

This PR is low priority, as it's just a cleanup


melonora (approved):
Thanks for the cleanup!

PR #969, opened 2025-08-18
Zarr v3 support

### LINKED ISSUE #979: deprecations and refactors in zarr v3 work.
We had quite a list of deprecations and also code that could be refactored. These mostly have been completed now in #969 . This issue just tracks what has been done.

- [x] Removal of all add_<element> methods
- [x] Added groupby `observed=False` to silence warning that users would have no control over.
- [x] Changed deprecated `resolution` parameter to `quad_segs`  in `to_polygons` to silence the warning for users.
- [x] Use `.init_from_elements` instead of deprecated `from_elements_dict` in `join_spatialelement_table` and `deepcopy`. Removed deprecated `from_elements_dict`.
- [x] In `match_table_to_element` removed the deprecated assumption of table with name `table` being present.
- [x] Removed deprecated kwargs of `polygon_query`.
- [x] Throughout code base remove deprecated `tables` argument in `init_from_elements`.    
- [x] Remove deprecated `element` argument from `transform_element_to_coordinate_system`.
- [x] Remove deprecated `table` property, setter and deleter.  
- [x] Remove deprecated `save_transformations`.
- [x] Remove deprecated `get_channels`.
- [x] When parsing raster data, that has to be transposed, we don't give a warning anymore. Instead the description of the behaviour has been moved to the docstring.

In addition to this, overall we had 3600~ warnings when running all tests. This was getting much for even just opening the CI page. With all addressed deprecation removals and updates to the tests, the warnings have been reduced to 758.

LucaMarconato:
closes #979 
Resources:
- [Zarr migration guide](https://zarr.readthedocs.io/en/latest/user-guide/v3_migration.html).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "anndata>=0.9.1",
     "click",
     "dask-image",
-    "dask>=2024.4.1,<=2024.11.2",
+    "dask>=2024.10.0,<=2024.11.2",
     "datashader",
     "fsspec[s3,http]",
     "geopandas>=0.14",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,7 +98,7 @@ xfail_strict = true
 addopts = [
 #    "-Werror",  # if 3rd party libs raise DeprecationWarnings, just use filterwarnings below
     "--import-mode=importlib",  # allow using test files with same name
-    "-s"  # print output from tests
+    "-s",  # print output from tests
 ]
 # These are all markers coming from xarray, dask or anndata. Added here to silence warnings.
 markers = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     "bump2version",
+    "sentry-prevent-cli",
 ]
 test = [
     "pytest",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,12 @@ addopts = [
     "--import-mode=importlib",  # allow using test files with same name
     "-s"  # print output from tests
 ]
+# These are all markers coming from xarray, dask or anndata. Added here to silence warnings.
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+    "gpu: run test on GPU using CuPY.",
+    "skip_with_pyarrow_strings: skipwhen pyarrow string conversion is turned on",
+]
 # info on how to use this https://stackoverflow.com/questions/57925071/how-do-i-avoid-getting-deprecationwarning-from-inside-dependencies-with-pytest
 filterwarnings = [
     #     "ignore:.*U.*mode is deprecated:DeprecationWarning",
--- a/src/spatialdata/_core/_elements.py
+++ b/src/spatialdata/_core/_elements.py
@@ -42,7 +42,13 @@ class Elements(UserDict[str, T]):
     def _check_key(key: str, element_keys: Iterable[str], shared_keys: set[str | None]) -> None:
         check_valid_name(key)
         if key in element_keys:
-            warn(f"Key `{key}` already exists. Overwriting it in-memory.", UserWarning, stacklevel=2)
+            warn(
+                f"Key `{key}` already exists. Overwriting it in-memory. If you want to silence this warning "
+                f"either delete it from memory 'del sdata[{key}]' (does not make changes on disk) or filter the "
+                f"warning.",
+                UserWarning,
+                stacklevel=2,
+            )
         else:
             try:
                 check_key_is_case_insensitively_unique(key, shared_keys)
--- a/src/spatialdata/_core/_elements.py
+++ b/src/spatialdata/_core/_elements.py
@@ -3,7 +3,6 @@
 from collections import UserDict
 from collections.abc import Iterable, KeysView, ValuesView
 from typing import TypeVar
-from warnings import warn
 
 from anndata import AnnData
 from dask.dataframe import DataFrame as DaskDataFrame
@@ -41,15 +40,7 @@ class Elements(UserDict[str, T]):
     @staticmethod
     def _check_key(key: str, element_keys: Iterable[str], shared_keys: set[str | None]) -> None:
         check_valid_name(key)
-        if key in element_keys:
-            warn(
-                f"Key `{key}` already exists. Overwriting it in-memory. If you want to silence this warning "
-                f"either delete it from memory 'del sdata[{key}]' (does not make changes on disk) or filter the "
-                f"warning.",
-                UserWarning,
-                stacklevel=2,
-            )
-        else:
+        if key not in element_keys:
             try:
                 check_key_is_case_insensitively_unique(key, shared_keys)
             except ValueError as e:
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -1115,7 +1115,7 @@ class SpatialData:
         overwrite: bool = False,
         consolidate_metadata: bool = True,
         sdata_formats: SpatialDataFormatType | list[SpatialDataFormatType] | None = None,
-    ) -> None:
+    ) -> tuple[Path | None, Path | None]:
         """
         Write the `SpatialData` object to a Zarr store.
 
@@ -1140,6 +1140,10 @@ class SpatialData:
             By default, the latest format is used for all elements, i.e.
             :class:`~spatialdata._io.format.CurrentRasterFormat`, :class:`~spatialdata._io.format.CurrentShapesFormat`,
             :class:`~spatialdata._io.format.CurrentPointsFormat`, :class:`~spatialdata._io.format.CurrentTablesFormat`.
+
+        Returns
+        -------
+        The old path and the new path of the SpatialData zarr store.
         """
         from spatialdata._io.format import _parse_formats
 
@@ -1166,14 +1170,20 @@ class SpatialData:
                 parsed_formats=parsed,
             )
 
+        old_path = self.path
+        if self.path != file_path:
+            self.path = file_path
         if parse_url(file_path):
-            if self.path != file_path:
-                old_path = self.path
-                self.path = file_path
-                logger.info(f"The Zarr backing store has been changed from {old_path} the new file path: {file_path}")
-
             if consolidate_metadata:
                 self.write_consolidated_metadata()
+        else:
+            warnings.warn(
+                "The SpatialData object is empty. Only the directory has been written, but it is nozarr store.",
+                UserWarning,
+                stacklevel=2,
+            )
+
+        return old_path, self.path
 
     def _write_element(
         self,
@@ -1272,15 +1282,18 @@ class SpatialData:
         If you pass a list of names, the elements will be written one by one. If an error occurs during the writing of
         an element, the writing of the remaining elements will not be attempted.
         """
+        from spatialdata._io.format import _parse_formats
+
+        parsed_formats = _parse_formats(formats=sdata_formats)
+        if parse_url(self.path) is None:
+            store = parse_url(self.path, mode="w", fmt=parsed_formats["SpatialData"]).store
+            store.close()
         if isinstance(element_name, list):
             for name in element_name:
                 assert isinstance(name, str)
                 self.write_element(name, overwrite=overwrite)
             return
 
-        from spatialdata._io.format import _parse_formats
-
-        parsed_formats = _parse_formats(formats=sdata_formats)
         check_valid_name(element_name)
         self._validate_element_names_are_unique()
         element = self.get(element_name)
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -1256,27 +1256,26 @@ class SpatialData:
                 image=element,
                 group=element_group,
                 name=element_name,
-                format=parsed_formats["raster"],
+                element_format=parsed_formats["raster"],
             )
         elif element_type == "labels":
             write_labels(
                 labels=element,
                 group=root_group,
                 name=element_name,
-                format=parsed_formats["raster"],
+                element_format=parsed_formats["raster"],
             )
         elif element_type == "points":
             write_points(
                 points=element,
                 group=element_group,
-                format=parsed_formats["points"],
+                element_format=parsed_formats["points"],
             )
         elif element_type == "shapes":
             write_shapes(
                 shapes=element,
                 group=element_group,
-                name=element_name,
-                format=parsed_formats["shapes"],
+                element_format=parsed_formats["shapes"],
             )
         elif element_type == "tables":
             write_table(
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -1043,7 +1043,7 @@ class SpatialData:
         overwrite: bool = False,
         saving_an_element: bool = False,
     ) -> None:
-        from spatialdata._io._utils import _backed_elements_contained_in_path, _is_subfolder, _open_zarr_store
+        from spatialdata._io._utils import _backed_elements_contained_in_path, _is_subfolder
 
         if isinstance(file_path, str):
             file_path = Path(file_path)
@@ -1052,7 +1052,6 @@ class SpatialData:
             raise ValueError(f"file_path must be a string or a Path object, type(file_path) = {type(file_path)}.")
 
         if os.path.exists(file_path):
-            _open_zarr_store(file_path, mode="r")
             if parse_url(file_path, mode="r", fmt=FormatV05()) is None:
                 raise ValueError(
                     "The target file path specified already exists, and it has been detected to not be a Zarr store. "
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -18,7 +18,6 @@ from dask.delayed import Delayed
 from geopandas import GeoDataFrame
 from ome_zarr.format import FormatV05
 from ome_zarr.io import parse_url
-from ome_zarr.types import JSONDict
 from shapely import MultiPolygon, Polygon
 from xarray import DataArray, DataTree
 
@@ -32,9 +31,6 @@ from spatialdata._core.validation import (
 )
 from spatialdata._logging import logger
 from spatialdata._types import ArrayLike, Raster_T
-from spatialdata._utils import (
-    _error_message_add_element,
-)
 from spatialdata.models import (
     Image2DModel,
     Image3DModel,
@@ -1805,16 +1801,6 @@ class SpatialData:
 
         return read_zarr(file_path, selection=selection)
 
-    def add_image(
-        self,
-        name: str,
-        image: DataArray | DataTree,
-        storage_options: JSONDict | list[JSONDict] | None = None,
-        overwrite: bool = False,
-    ) -> None:
-        """Deprecated. Use `sdata[name] = image` instead."""  # noqa: D401
-        _error_message_add_element()
-
     @property
     def images(self) -> Images:
         """Return images as a Dict of name to image data."""
@@ -2210,26 +2196,6 @@ class SpatialData:
                 assert model == ShapesModel
                 element_type = "shapes"
             elements_dict.setdefault(element_type, {})[name] = element
-        # when the "tables" argument is removed, we can remove all this if block
-        if tables is not None:
-            warnings.warn(
-                'The "tables" argument is deprecated and will be removed in a future version. Please '
-                "specifies the tables in the `elements` argument. Until the removal occurs, the `elements` "
-                "variable will be automatically populated with the tables if the `tables` argument is not None.",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-            if "tables" in elements_dict:
-                raise ValueError(
-                    "The tables key is already present in the elements dictionary. Please do not specify "
-                    "the `tables` argument."
-                )
-            elements_dict["tables"] = {}
-            if isinstance(tables, AnnData):
-                elements_dict["tables"]["table"] = tables
-            else:
-                for name, table in tables.items():
-                    elements_dict["tables"][name] = table
         return cls(**elements_dict, attrs=attrs)
 
     def subset(
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -16,7 +16,6 @@ from dask.dataframe import DataFrame as DaskDataFrame
 from dask.dataframe import read_parquet
 from dask.delayed import Delayed
 from geopandas import GeoDataFrame
-from ome_zarr.io import parse_url
 from shapely import MultiPolygon, Polygon
 from xarray import DataArray, DataTree
 
@@ -1141,6 +1140,7 @@ class SpatialData:
             :class:`~spatialdata._io.format.CurrentRasterFormat`, :class:`~spatialdata._io.format.CurrentShapesFormat`,
             :class:`~spatialdata._io.format.CurrentPointsFormat`, :class:`~spatialdata._io.format.CurrentTablesFormat`.
         """
+        from spatialdata._io._utils import _resolve_zarr_store
         from spatialdata._io.format import _parse_formats
 
         parsed = _parse_formats(sdata_formats)
@@ -1150,9 +1150,9 @@ class SpatialData:
         self._validate_can_safely_write_to_path(file_path, overwrite=overwrite)
         self._validate_all_elements()
 
-        # parse_url cannot be replaced here as it actually also initialized an ome-zarr store.
-        store = parse_url(file_path, mode="w", fmt=parsed["SpatialData"]).store
-        zarr_group = zarr.open_group(store=store, mode="r+")
+        store = _resolve_zarr_store(file_path)
+        zarr_format = parsed["SpatialData"].zarr_format
+        zarr_group = zarr.create_group(store=store, overwrite=overwrite, zarr_format=zarr_format)
         self.write_attrs(zarr_group=zarr_group, sdata_format=parsed["SpatialData"])
         store.close()
 
@@ -1162,7 +1162,7 @@ class SpatialData:
                 zarr_container_path=file_path,
                 element_type=element_type,
                 element_name=element_name,
-                overwrite=overwrite,
+                overwrite=False,
                 parsed_formats=parsed,
             )
 
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -33,7 +33,6 @@ from spatialdata._core.validation import (
 from spatialdata._logging import logger
 from spatialdata._types import ArrayLike, Raster_T
 from spatialdata._utils import (
-    _deprecation_alias,
     _error_message_add_element,
 )
 from spatialdata.models import (
@@ -233,34 +232,6 @@ class SpatialData:
                             f"({table.obs[instance_key].dtype}) that does not match the dtype of the indices of "
                             f"the annotated element ({dtype})."
                         )
-
-    @staticmethod
-    def from_elements_dict(
-        elements_dict: dict[str, SpatialElement | AnnData],
-        attrs: Mapping[Any, Any] | None = None,
-    ) -> SpatialData:
-        """
-        Create a SpatialData object from a dict of elements.
-
-        Parameters
-        ----------
-        elements_dict
-            Dict of elements. The keys are the names of the elements and the values are the elements.
-            A table can be present in the dict, but only at most one; its name is not used and can be anything.
-        attrs
-            Additional attributes to store in the SpatialData object.
-
-        Returns
-        -------
-        The SpatialData object.
-        """
-        warnings.warn(
-            'This method is deprecated and will be removed in a future release. Use "SpatialData.init_from_elements('
-            ')" instead. For the moment, such methods will be automatically called.',
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        return SpatialData.init_from_elements(elements=elements_dict, attrs=attrs)
 
     @staticmethod
     def get_annotated_regions(table: AnnData) -> list[str]:
@@ -816,7 +787,6 @@ class SpatialData:
             # set the new transformations
             set_transformation(element=element, transformation=new_transformations, set_all=True)
 
-    @_deprecation_alias(element="element_name", version="0.3.0")
     def transform_element_to_coordinate_system(
         self,
         element_name: str,
@@ -1282,7 +1252,7 @@ class SpatialData:
                 table=element,
                 group=element_type_group,
                 name=element_name,
-                format=parsed_formats["tables"],
+                element_format=parsed_formats["tables"],
             )
         else:
             raise ValueError(f"Unknown element type: {element_type}")
@@ -1304,7 +1274,7 @@ class SpatialData:
             The name(s) of the element(s) to write.
         overwrite
             If True, overwrite the element if it already exists.
-        format
+        sdata_formats
             It is recommended to leave this parameter equal to `None`. See more details in the documentation of
              `SpatialData.write()`.
 
@@ -1823,51 +1793,6 @@ class SpatialData:
             TableModel().validate(v)
             self._tables[k] = v
 
-    @property
-    def table(self) -> None | AnnData:
-        """
-        Return table with name table from tables if it exists.
-
-        Returns
-        -------
-        The table.
-        """
-        warnings.warn(
-            "Table accessor will be deprecated with SpatialData version 0.1, use sdata.tables instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        # Isinstance will still return table if anndata has 0 rows.
-        if isinstance(self.tables.get("table"), AnnData):
-            return self.tables["table"]
-        return None
-
-    @table.setter
-    def table(self, table: AnnData) -> None:
-        warnings.warn(
-            "Table setter will be deprecated with SpatialData version 0.1, use tables instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        TableModel().validate(table)
-        if self.tables.get("table") is not None:
-            raise ValueError("The table already exists. Use del sdata.tables['table'] to remove it first.")
-        self.tables["table"] = table
-
-    @table.deleter
-    def table(self) -> None:
-        """Delete the table."""
-        warnings.warn(
-            "del sdata.table will be deprecated with SpatialData version 0.1, use del sdata.tables['table'] instead.",
-            DeprecationWarning,
-            stacklevel=2,
-        )
-        if self.tables.get("table"):
-            del self.tables["table"]
-        else:
-            # More informative than the error in the zarr library.
-            raise KeyError("table with name 'table' not present in the SpatialData object.")
-
     @staticmethod
     def read(file_path: Path | str, selection: tuple[str] | None = None) -> SpatialData:
         """
@@ -2185,14 +2110,14 @@ class SpatialData:
             yield from d.values()
 
     def _gen_elements(
-        self, include_table: bool = False
+        self, include_tables: bool = False
     ) -> Generator[tuple[str, str, SpatialElement | AnnData], None, None]:
         """
         Generate elements contained in the SpatialData instance.
 
         Parameters
         ----------
-        include_table
+        include_tables
             Whether to also generate table elements.
 
         Returns
@@ -2201,7 +2126,7 @@ class SpatialData:
         itself.
         """
         element_types = ["images", "labels", "points", "shapes"]
-        if include_table:
+        if include_tables:
             element_types.append("tables")
         for element_type in element_types:
             d = getattr(SpatialData, element_type).fget(self)
@@ -2235,7 +2160,7 @@ class SpatialData:
         -------
         A generator that yields tuples containing the name, description, and element objects themselves.
         """
-        return self._gen_elements(include_table=True)
+        return self._gen_elements(include_tables=True)
 
     def _validate_element_names_are_unique(self) -> None:
         """
@@ -2370,7 +2295,7 @@ class SpatialData:
         """
         elements_dict: dict[str, SpatialElement] = {}
         names_tables_to_keep: set[str] = set()
-        for element_type, element_name, element in self._gen_elements(include_table=True):
+        for element_type, element_name, element in self._gen_elements(include_tables=True):
             if element_name in element_names:
                 if element_type != "tables":
                     elements_dict.setdefault(element_type, {})[element_name] = element
@@ -2403,7 +2328,7 @@ class SpatialData:
 
     def __contains__(self, key: str) -> bool:
         element_dict = {
-            element_name: element_value for _, element_name, element_value in self._gen_elements(include_table=True)
+            element_name: element_value for _, element_name, element_value in self._gen_elements(include_tables=True)
         }
         return key in element_dict
 
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -18,6 +18,7 @@ from dask.delayed import Delayed
 from geopandas import GeoDataFrame
 from shapely import MultiPolygon, Polygon
 from xarray import DataArray, DataTree
+from zarr.errors import GroupNotFoundError
 
 from spatialdata._core._elements import Images, Labels, Points, Shapes, Tables
 from spatialdata._core.validation import (
@@ -563,13 +564,6 @@ class SpatialData:
         else:
             raise TypeError("Path must be `None`, a `str` or a `Path` object.")
 
-        if not self.is_self_contained():
-            logger.info(
-                "The SpatialData object is not self-contained (i.e. it contains some elements that are Dask-backed from"
-                f" locations outside {self.path}). Please see the documentation of `is_self_contained()` to understand"
-                f" the implications of working with SpatialData objects that are not self-contained."
-            )
-
     def locate_element(self, element: SpatialElement) -> list[str]:
         """
         Locate a SpatialElement within the SpatialData object and returns its Zarr paths relative to the root.
@@ -1047,7 +1041,7 @@ class SpatialData:
         overwrite: bool = False,
         saving_an_element: bool = False,
     ) -> None:
-        from spatialdata._io._utils import _backed_elements_contained_in_path, _is_subfolder
+        from spatialdata._io._utils import _backed_elements_contained_in_path, _is_subfolder, _resolve_zarr_store
 
         if isinstance(file_path, str):
             file_path = Path(file_path)
@@ -1057,11 +1051,14 @@ class SpatialData:
 
         # TODO: add test for this
         if os.path.exists(file_path):
-            if parse_url(file_path, mode="r") is None:
+            store = _resolve_zarr_store(file_path)
+            try:
+                zarr.open(store, mode="r")
+            except GroupNotFoundError as err:
                 raise ValueError(
                     "The target file path specified already exists, and it has been detected to not be a Zarr store. "
                     "Overwriting non-Zarr stores is not supported to prevent accidental data loss."
-                )
+                ) from err
             if not overwrite:
                 raise ValueError(
                     "The Zarr store already exists. Use `overwrite=True` to try overwriting the store. "
@@ -1113,6 +1110,7 @@ class SpatialData:
         file_path: str | Path,
         overwrite: bool = False,
         consolidate_metadata: bool = True,
+        update_sdata_path: bool = True,
         sdata_formats: SpatialDataFormatType | list[SpatialDataFormatType] | None = None,
     ) -> None:
         """
@@ -1129,16 +1127,35 @@ class SpatialData:
             If `True`, triggers :func:`zarr.convenience.consolidate_metadata`, which writes all the metadata in a single
             file at the root directory of the store. This makes the data cloud accessible, which is required for certain
             cloud stores (such as S3).
-        format
+        update_sdata_path
+            Whether to update the `path` attribute of the `SpatialData` object to `file_path` after a successful write
+            (default yes). Here are the implications.
+
+                - If `True`, and if the `SpatialData` object has dask-backed elements, the object will become "not
+                 self-contained" because the dask-backed element will have a path that is different from the new
+                 `sdata.path` attribute (now equal to `file_path`). By re-reading the object from disk, the object will
+                 become self-contained again.
+                - If `False`, the `SpatialData` object will keep its current `path` attribute, meaning that calling
+                `sdata.write_element()`, `sdata.write_attrs()`, ... will write the data to the current `sdata.path`
+                location, not to the `file_path` location.
+
+            Please consult :func:`spatialdata.SpatialData.is_self_contained` for more information on the implications of
+            working with self-contained and non-self-contained SpatialData objects.
+        sdata_formats
             The format to use for writing the elements of the `SpatialData` object. It is recommended to leave this
             parameter equal to `None` (default to latest format for all the elements). If not `None`, it must be
-            either a format for an element, or a list of formats.
-            For example it can be a subset of the following list `[RasterFormatVXX(), ShapesFormatVXX(),
-            PointsFormatVXX(), TablesFormatVXX()]`. (XX denote the version number, and should be replaced with the
-            respective format; the version numbers can differ across elements).
-            By default, the latest format is used for all elements, i.e.
-            :class:`~spatialdata._io.format.CurrentRasterFormat`, :class:`~spatialdata._io.format.CurrentShapesFormat`,
-            :class:`~spatialdata._io.format.CurrentPointsFormat`, :class:`~spatialdata._io.format.CurrentTablesFormat`.
+            either a format for an element, or a list of formats. For example it can be a subset of the following
+            list `[RasterFormatVXX(), ShapesFormatVXX(), PointsFormatVXX(), TablesFormatVXX()]`. (XX denote the
+            version number, and should be replaced with the respective format; the version numbers can differ across
+            elements). By default, the latest format is used for all elements,
+            i.e. :class:`~spatialdata._io.format.CurrentRasterFormat`,
+            :class:`~spatialdata._io.format.CurrentShapesFormat`,
+            :class:`~spatialdata._io.format.CurrentPointsFormat`,
+            :class:`~spatialdata._io.format.CurrentTablesFormat`. Also, by default, if a format for the SpatialData
+            "container" object (i.e. SpatialDataContainerFormatVXX) is specified, but the format for some elements is
+            unspecified, the element formats will be set to the latest element format compatible with the specified
+            SpatialData container format. All the formats and relationships between them are defined in
+            `spatialdata._io.format.py`.
         """
         from spatialdata._io._utils import _resolve_zarr_store
         from spatialdata._io.format import _parse_formats
@@ -1166,7 +1183,7 @@ class SpatialData:
                 parsed_formats=parsed,
             )
 
-        if self.path != file_path:
+        if self.path != file_path and update_sdata_path:
             self.path = file_path
 
         if consolidate_metadata:
@@ -1640,7 +1657,7 @@ class SpatialData:
         element_name: str | None = None,
         consolidate_metadata: bool | None = None,
         write_attrs: bool = True,
-        format: SpatialDataContainerFormatType | None = None,
+        sdata_format: SpatialDataContainerFormatType | None = None,
     ) -> None:
         """
         Write the metadata of a single element, or of all elements, to the Zarr store, without rewriting the data.
@@ -1661,6 +1678,11 @@ class SpatialData:
         consolidate_metadata
             If True, consolidate the metadata to more easily support remote reading. By default write the metadata
             only if the metadata was already consolidated.
+        write_attrs
+            If True, write the SpatialData.attrs metadata to the root of the Zarr store.
+        sdata_format
+            The format to use for writing the metadata of the `SpatialData` object. See more details in the
+            documentation of `SpatialData.write()`.
 
         Notes
         -----
@@ -1677,7 +1699,7 @@ class SpatialData:
         # TODO: write .attrs['spatialdata_attrs'] metadata for DaskDataFrame.
 
         if write_attrs:
-            self.write_attrs(format=format)
+            self.write_attrs(sdata_format=sdata_format)
 
         # TODO: discuss when has_consolidated_metadata that we should just consolidate it because after a writing
         # operation the consolidated store could otherwise be out of sync.
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -1154,7 +1154,7 @@ class SpatialData:
             if index == 0:
                 # parse_url cannot be replaced here as it actually also initialized an ome-zarr store.
                 store = parse_url(file_path, mode="w", fmt=parsed["SpatialData"]).store
-                zarr_group = zarr.open_group(store=store, mode="w" if overwrite else "a")
+                zarr_group = zarr.open_group(store=store, mode="r+" if overwrite else "a")
                 self.write_attrs(zarr_group=zarr_group, sdata_format=parsed["SpatialData"])
                 store.close()
             self._write_element(
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -122,7 +122,6 @@ class SpatialData:
     annotation directly.
     """
 
-    @_deprecation_alias(table="tables", version="0.1.0")
     def __init__(
         self,
         images: dict[str, Raster_T] | None = None,
@@ -604,94 +603,6 @@ class SpatialData:
                 f" the implications of working with SpatialData objects that are not self-contained."
             )
 
-    def _get_groups_for_element(
-        self, zarr_path: Path, element_type: str, element_name: str, use_consolidated: bool = True
-    ) -> tuple[zarr.Group, zarr.Group, zarr.Group]:
-        """
-        Get the Zarr groups for the root, element_type and element for a specific element.
-
-        The store must exist, but creates the element type group and the element group if they don't exist.
-
-        Parameters
-        ----------
-        zarr_path
-            The path to the Zarr storage.
-        element_type
-            type of the element; must be in ["images", "labels", "points", "polygons", "shapes", "tables"].
-        element_name
-            name of the element
-
-        Returns
-        -------
-        either the existing Zarr subgroup or a new one.
-        """
-        from spatialdata._io._utils import _open_zarr_store
-
-        if not isinstance(zarr_path, Path):
-            raise ValueError("zarr_path should be a Path object")
-
-        if element_type not in [
-            "images",
-            "labels",
-            "points",
-            "polygons",
-            "shapes",
-            "tables",
-        ]:
-            raise ValueError(f"Unknown element type {element_type}")
-
-        store = _open_zarr_store(zarr_path, mode="r+")
-        if element_type != "labels":
-            root = zarr.open_group(store=store, mode="a")
-        else:
-            # This is required as ome-zarr accesses the labels group within root. If data has been consolidated
-            # before it will already look for the labels element just added, but the data has not been reconsolidated
-            # yet. Thus, when writing we open the root store here with use_consolidated == False.
-            root = zarr.open_group(store=store, mode="a", use_consolidated=use_consolidated)
-
-        element_type_group = root.require_group(element_type)
-        # This is required as adata performs a consolidated check before writing anything. If the Tables group was
-        # consolidated before, this prevents anndata from writing. Therefore, we read with use_consolidated == False
-        # when writing.
-        if not use_consolidated and element_type in ["labels", "tables"]:
-            element_type_group = zarr.open_group(
-                element_type_group.store_path, mode="a", use_consolidated=use_consolidated
-            )
-
-        element_name_group = element_type_group.require_group(element_name)
-        return root, element_type_group, element_name_group
-
-    def _group_for_element_exists(self, zarr_path: Path, element_type: str, element_name: str) -> bool:
-        """
-        Check if the group for an element exists.
-
-        Parameters
-        ----------
-        element_type
-            type of the element; must be in ["images", "labels", "points", "polygons", "shapes", "tables"].
-        element_name
-            name of the element
-
-        Returns
-        -------
-        True if the group exists, False otherwise.
-        """
-        from spatialdata._io._utils import _open_zarr_store
-
-        store = _open_zarr_store(zarr_path, mode="r")
-        root = zarr.open_group(store=store, mode="r")
-        assert element_type in [
-            "images",
-            "labels",
-            "points",
-            "polygons",
-            "shapes",
-            "tables",
-        ]
-        exists = element_type in root and element_name in root[element_type]
-        store.close()
-        return exists
-
     def locate_element(self, element: SpatialElement) -> list[str]:
         """
         Locate a SpatialElement within the SpatialData object and returns its Zarr paths relative to the root.
@@ -721,7 +632,6 @@ class SpatialData:
             raise ValueError("Found an element name with a '/' character. This is not allowed.")
         return [f"{found_element_type[i]}/{found_element_name[i]}" for i in range(len(found))]
 
-    @_deprecation_alias(filter_table="filter_tables", version="0.1.0")
     def filter_by_coordinate_system(
         self,
         coordinate_system: str | list[str],
@@ -1315,6 +1225,8 @@ class SpatialData:
         overwrite: bool,
         parsed_formats: dict[str, SpatialDataFormatType] | None = None,
     ) -> None:
+        from spatialdata._io.io_zarr import _get_groups_for_element
+
         if not isinstance(zarr_container_path, Path):
             raise ValueError(
                 f"zarr_container_path must be a Path object, type(zarr_container_path) = {type(zarr_container_path)}."
@@ -1324,7 +1236,7 @@ class SpatialData:
             file_path=file_path_of_element, overwrite=overwrite, saving_an_element=True
         )
 
-        root_group, element_type_group, element_group = self._get_groups_for_element(
+        root_group, element_type_group, element_group = _get_groups_for_element(
             zarr_path=zarr_container_path, element_type=element_type, element_name=element_name, use_consolidated=False
         )
         from spatialdata._io import (
@@ -1567,6 +1479,7 @@ class SpatialData:
     def _validate_can_write_metadata_on_element(self, element_name: str) -> tuple[str, SpatialElement | AnnData] | None:
         """Validate if metadata can be written on an element, returns None if it cannot be written."""
         from spatialdata._io._utils import _is_element_self_contained
+        from spatialdata._io.io_zarr import _group_for_element_exists
 
         # check the element exists in the SpatialData object
         element = self.get(element_name)
@@ -1589,7 +1502,7 @@ class SpatialData:
         self._check_element_not_on_disk_with_different_type(element_type=element_type, element_name=element_name)
 
         # check if the element exists in the Zarr storage
-        if not self._group_for_element_exists(
+        if not _group_for_element_exists(
             zarr_path=Path(self.path),
             element_type=element_type,
             element_name=element_name,
@@ -1623,6 +1536,8 @@ class SpatialData:
             The name of the element to write the channel names of. If None, write the channel names of all image
             elements.
         """
+        from spatialdata._io.io_zarr import _get_groups_for_element
+
         if element_name is not None:
             check_valid_name(element_name)
             if element_name not in self:
@@ -1642,10 +1557,8 @@ class SpatialData:
 
         # Mypy does not understand that path is not None so we have the check in the conditional
         if element_type == "images" and self.path is not None:
-            _, _, element_group = self._get_groups_for_element(
-                zarr_path=Path(self.path),
-                element_type=element_type,
-                element_name=element_name,
+            _, _, element_group = _get_groups_for_element(
+                zarr_path=Path(self.path), element_type=element_type, element_name=element_name, use_consolidated=False
             )
 
             from spatialdata._io._utils import overwrite_channel_names
@@ -1663,6 +1576,8 @@ class SpatialData:
         element_name
             The name of the element to write. If None, write the transformations of all elements.
         """
+        from spatialdata._io.io_zarr import _get_groups_for_element
+
         if element_name is not None:
             check_valid_name(element_name)
             if element_name not in self:
@@ -1686,7 +1601,7 @@ class SpatialData:
 
         # Mypy does not understand that path is not None so we have a conditional
         assert self.path is not None
-        _, _, element_group = self._get_groups_for_element(
+        _, _, element_group = _get_groups_for_element(
             zarr_path=Path(self.path),
             element_type=element_type,
             element_name=element_name,
@@ -1805,6 +1720,8 @@ class SpatialData:
         if write_attrs:
             self.write_attrs()
 
+        # TODO: discuss when has_consolidated_metadata that we should just consolidate it because after a writing
+        # operation the consolidated store could otherwise be out of sync.
         if consolidate_metadata is None and self.has_consolidated_metadata():
             consolidate_metadata = True
         if consolidate_metadata:
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -1701,9 +1701,7 @@ class SpatialData:
         if write_attrs:
             self.write_attrs(sdata_format=sdata_format)
 
-        # TODO: discuss when has_consolidated_metadata that we should just consolidate it because after a writing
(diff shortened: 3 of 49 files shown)


src/spatialdata/_core/spatialdata.py
```diff
@@ -621,10 +621,13 @@ def _get_groups_for_element(
         -------
         either the existing Zarr subgroup or a new one.
         """
+        from spatialdata._io.format import SpatialDataFormat
+
         if not isinstance(zarr_path, Path):
             raise ValueError("zarr_path should be a Path object")
-        store = parse_url(zarr_path, mode="r+").store
-        root = zarr.group(store=store)
+        store = SpatialDataFormat().init_store(str(zarr_path), mode="r+")
```
LucaMarconato: can probably be restored to `parse_url`. I used this while debugging.

LucaMarconato:
In the latest commit I started (still wip) removing the usage of `parse_url()` in favor of `_open_zarr_store()` from @berombau (from the `cloud` branch).

Importantly, [`parse_url()` returns `None`](https://ome-zarr.readthedocs.io/en/stable/api/io.html#ome_zarr.io.parse_url) if the URL doesn't exist or if there is an error opening the path. Instead `_open_zarr_store()` never complains—we need to adjust the logic accordingly.

In particular, when calling `_open_zarr_store()` we should be able to intercept the following cases (for which `parse_url()` returns `None`):
- we are trying to "open" a path that exists but that is not a Zarr store;
- we are trying to "open", in `r` or `r+` a path that doesn't exist, or that cannot be created.

We should not solve this with ad-hoc checks, but by using the right API from `zarr-python`, so that we get a `None` or an exception if we are trying to perform such operations.

melonora:
@falexwolf everything is green now for zarr v3. I will still do a refactor and add some more tests to ensure the backward compatibility (allowing users to update everything they have with previous formats) and add some deprecation warnings etc. But first initial support is there!

We do not have the sharding api yet though, but we will put a social media post out for people to get involved in that so we can garner some feedback.

falexwolf:
That's awesome, @melonora! We'll try this out!

Zethson:
All of our use cases still run through with this PR so I think we're good.

melonora:
currently blocked by https://github.com/ome/ome-zarr-py/pull/491

src/spatialdata/_io/_utils.py
```diff
@@ -363,24 +434,40 @@ def _is_element_self_contained(
...
-    """
-    warnings.warn(
-        "This function is deprecated and should be replaced by `SpatialData.write_transformations()` or "
-        "`SpatialData.write_metadata()`, which gives more control over which metadata to write. This function will call"
-        " `SpatialData.write_transformations()`; please call this function directly.",
-        DeprecationWarning,
-        stacklevel=2,
-    )
-    sdata.write_transformations()
+def _open_zarr_store(path: StoreLike, **kwargs: Any) -> zarr.storage.StoreLike:
```
ilan-gold: I think I would rename this function to something like `_to_fsspec_store` or similar. Or give it a clearer docstring because I'm not clear what it does.  For example, the second if statement returns a `LocalStore` but then later if a `LocalStore` is detected to be backing a `zarr.Group`, an `FsspecStore` is used

src/spatialdata/_core/spatialdata.py
```diff
@@ -1123,7 +1052,8 @@ def _validate_can_safely_write_to_path(
             raise ValueError(f"file_path must be a string or a Path object, type(file_path) = {type(file_path)}.")
 
         if os.path.exists(file_path):
-            if parse_url(file_path, mode="r") is None:
+            _open_zarr_store(file_path, mode="r")
```
ilan-gold: What does this do? This output is not assigned

src/spatialdata/_io/io_zarr.py
```diff
@@ -1,25 +1,32 @@
...
 from spatialdata._io.io_raster import _read_multiscale
 from spatialdata._io.io_shapes import _read_shapes
 from spatialdata._io.io_table import _read_table
 from spatialdata._logging import logger
 
 
-def _open_zarr_store(store: str | Path | zarr.Group) -> tuple[zarr.Group, str]:
+# TODO: remove with incoming remote read / write PR
+# Not removing this now as it requires substantial extra refactor beyond scope of zarrv3 PR.
+def _open_zarr_store(
```
ilan-gold: I think there are two functions with this name, which seems bad

src/spatialdata/_io/_utils.py
```diff
@@ -363,24 +434,40 @@ def _is_element_self_contained(
...
+
+    if isinstance(path, PosixUPath | WindowsUPath):
+        # if the input is a local path, use LocalStore
+        return LocalStore(path.path)
+
+    if isinstance(path, zarr.Group):
+        # if the input is a zarr.Group, wrap it with a store
+        if isinstance(path.store, LocalStore):
+            # create a simple FSStore if the store is a LocalStore with just the path
+            return FsspecStore(os.path.join(path.store.path, path.path), **kwargs)
```
ilan-gold, line 454: Specifically, this could preclude using `zarrs-python` at the moment given that we only check for `LocalStore`, I think.  I'm not sure how much of a hotpath this store is, but if it is used for heavy processing, that is something to consider.  If you're interested in `zarrs-python`, I would recommend running your test suite locally with it

src/spatialdata/_io/_utils.py
ilan-gold: In general, without knowing more about your use-case, I do wonder if having the user manage the store could be a simplifying factor. I understand for local storage, a file path/string is an acceptable object, but there are lots of different stores for different things that people might want to use. 

When I look at `ome-zarr-py` they don't do this, but they do appear to have a primitive of sorts for passing around the store information https://ome-zarr.readthedocs.io/en/stable/api/io.html#ome_zarr.io.ZarrLocation that isn't just strings.  

Maybe this doesn't matter for the use-case here, but I could see this being annoying for cloud access where people use different stores with very specific settings like access levels/keys.

src/spatialdata/_io/io_zarr.py
melonora: jup are aware of that, we would follow up with the remote read / write PR though in which this would be addressed. Discussion point is whether to do the refactor already here in this large PR.

src/spatialdata/_io/_utils.py
melonora: In general it is to allow for remote read and write. Luca already copied over some of the code though, see his comment:

> In the latest commit I started (still wip) removing the usage of `parse_url()` in favor of `_open_zarr_store()` from @berombau (from the `cloud` branch).
> 
> Importantly, [`parse_url()` returns `None`](https://ome-zarr.readthedocs.io/en/stable/api/io.html#ome_zarr.io.parse_url) if the URL doesn't exist or if there is an error opening the path. Instead `_open_zarr_store()` never complains—we need to adjust the logic accordingly.
> 
> In particular, when calling `_open_zarr_store()` we should be able to intercept the following cases (for which `parse_url()` returns `None`):
> 
> * we are trying to "open" a path that exists but that is not a Zarr store;
> * we are trying to "open", in `r` or `r+` a path that doesn't exist, or that cannot be created.
> 
> We should not solve this with ad-hoc checks, but by using the right API from `zarr-python`, so that we get a `None` or an exception if we are trying to perform such operations

ilan-gold: Ok I see.  In that case (I think you had asked previously), I would resolve this branch first with an eye towards that one, `cloud`.  For example https://zarr.readthedocs.io/en/latest/user-guide/storage.html#object-store or https://docs.rapids.ai/api/kvikio/nightly/zarr/ seem like nice use-cases, but judging by this function, they don't really gel well.

ilan-gold: i.e., `FSSpecStore` is not the only option and often not the desired one, although that might have been true for zarr v2

src/spatialdata/_core/spatialdata.py
melonora: Think Luca meant to replace `parse_url` here. I will remove for now and just do the refactor to the new function in 1 commit.

src/spatialdata/_io/io_zarr.py
```diff
@@ -236,3 +216,94 @@ def read_zarr(
...
     return sdata
+
+
+def _get_groups_for_element(
+    zarr_path: Path, element_type: str, element_name: str, use_consolidated: bool = True
+) -> tuple[zarr.Group, zarr.Group, zarr.Group]:
+    """
+    Get the Zarr groups for the root, element_type and element for a specific element.
+
+    The store must exist, but creates the element type group and the element group if they don't exist. In all cases
```
LucaMarconato, line 199: The IO mode "a" is used, which creates the store when it doesn't not exist. Instead, "the store must exist" implies using "r+" (which is what it was used in the previous implementation).

We should either change "a" -> "r+", or update the docstring by remove "the store must exist" and say that the store will be opened it exists, created otherwise.

src/spatialdata/_io/io_zarr.py
```diff
@@ -236,3 +216,94 @@ def read_zarr(
...
+        "tables",
+    ]:
+        raise ValueError(f"Unknown element type {element_type}")
+    resolved_store = _resolve_zarr_store(zarr_path)
+
+    # When writing, use_consolidated must be set to False. Otherwise, the metadata store
+    # can get out of sync with newly added elements (e.g., labels), leading to errors.
+    root_group = zarr.open_group(store=resolved_store, mode="a", use_consolidated=use_consolidated)
+    element_type_group = root_group.require_group(element_type)
+    element_type_group = zarr.open_group(element_type_group.store_path, mode="a", use_consolidated=use_consolidated)
```
LucaMarconato, line 241: I think that this line is redundant and it can be removed.

LucaMarconato: if it's needed (because of the consolidation metadata parameter), we may need to add this line also for the `element_name_group`.

src/spatialdata/_io/io_zarr.py
```diff
@@ -236,3 +217,98 @@ def read_zarr(
...
+    element_type
+        type of the element; must be in ["images", "labels", "points", "polygons", "shapes", "tables"].
+    element_name
+        name of the element
+
+    Returns
+    -------
+    True if the group exists, False otherwise.
+    """
+    store = _resolve_zarr_store(zarr_path, mode="r")
```
LucaMarconato: in `_get_groups_for_element()` we do not pass `mode` to `_resolve_zarr_store()`, while here we do. If it is possible (it should), I suggest to remove the `mode` argument from `_resolve_zarr_store()` and use it only when calling `zarr.open_group` or `zarr.require_group`.

In doing this, I would make sure that `_resolve_zarr_store()` never creates groups that do not exist, and never overwrite Zarr groups (or non-Zarr files/folders) that exist.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1123,7 +1055,7 @@ def _validate_can_safely_write_to_path(
             raise ValueError(f"file_path must be a string or a Path object, type(file_path) = {type(file_path)}.")
 
         if os.path.exists(file_path):
-            if parse_url(file_path, mode="r") is None:
+            if parse_url(file_path, mode="r", fmt=FormatV05()) is None:
```
LucaMarconato: needs to be updated to using `_resolve_zarr_store()`

src/spatialdata/_core/spatialdata.py
```diff
@@ -1205,13 +1137,17 @@ def write(
...
+
         if isinstance(file_path, str):
             file_path = Path(file_path)
         self._validate_can_safely_write_to_path(file_path, overwrite=overwrite)
         self._validate_all_elements()
 
-        store = parse_url(file_path, mode="w").store
-        zarr_group = zarr.group(store=store, overwrite=overwrite)
+        store = parse_url(file_path, mode="w", fmt=parsed["SpatialData"]).store
+        zarr_group = zarr.open_group(store=store, mode="w" if overwrite else "a")
```
LucaMarconato: we should remove `parse_url` and use `_resolve_zarr_store()` instead.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1221,8 +1157,8 @@ def write(
                 zarr_container_path=file_path,
                 element_type=element_type,
                 element_name=element_name,
-                overwrite=False,
-                format=format,
+                overwrite=overwrite,
```
LucaMarconato: I think I changed this to `overwrite` in this PR, but it should be `False`. In fact either we are writing to a fresh zarr store that was not present in the location, either we overwrote the previous store. In both cases, there will be no elements on disk, so there is no need to pass `overwrite`.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1251,32 +1189,62 @@ def _write_element(
...
+            write_table(
+                table=element,
+                group=element_type_group,
+                name=element_name,
+                element_format=parsed_formats["tables"],
+            )
         else:
             raise ValueError(f"Unknown element type: {element_type}")
 
     def write_element(
```
LucaMarconato, line 1263: Same here, I'd add the deprecation decorator to automatically change the old argument `format` into `sdata_formats`.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1240,8 +1176,10 @@ def _write_element(
         element_type: str,
         element_name: str,
         overwrite: bool,
-        format: SpatialDataFormat | list[SpatialDataFormat] | None = None,
+        parsed_formats: dict[str, SpatialDataFormatType] | None = None,
```
LucaMarconato, line 1199: Ok! No need to add the deprecation decorator because it's a private method.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1417,9 +1391,9 @@ def delete_element_from_disk(self, element_name: str | list[str]) -> None:
             )
 
         # delete the element
-        store = parse_url(self.path, mode="r+").store
-        root = zarr.group(store=store)
-        root[element_type].pop(element_name)
+        store = parse_url(self.path, mode="r+", fmt=FormatV05()).store
```
LucaMarconato: `parse_url()` needs to be replaced by the new approach.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1438,23 +1412,23 @@ def _check_element_not_on_disk_with_different_type(self, element_type: str, elem
                 )
 
     def write_consolidated_metadata(self) -> None:
```
LucaMarconato, line 1437: I'd remove `_write_consolidated_metadata()` by moving its code content here.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1438,23 +1412,23 @@ def _check_element_not_on_disk_with_different_type(self, element_type: str, elem
...
-        store.close()
+        from spatialdata._io.io_zarr import _write_consolidated_metadata
+
+        _write_consolidated_metadata(self.path)
 
     def has_consolidated_metadata(self) -> bool:
         return_value = False
-        store = parse_url(self.path, mode="r").store
-        if "zmetadata" in store:
+        store = parse_url(self.path, mode="r", fmt=FormatV05()).store
```
LucaMarconato: Also here, `parse_url()` needs to be replaced by the new method.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1438,23 +1412,23 @@ def _check_element_not_on_disk_with_different_type(self, element_type: str, elem
...
+        group = zarr.open_group(store, mode="r")
+        if getattr(group.metadata, "consolidated_metadata", None):
```
LucaMarconato, line 1448: Does this work for Zarr v2? I'd add a test. For Zarr v2 we were checking `if `zmetadata` in store:`

src/spatialdata/_core/spatialdata.py
```diff
@@ -1613,20 +1596,26 @@ def _element_type_and_name_from_element_path(self, element_path: str) -> tuple[s
...
+        spatialdata_container_format = parsed["SpatialData"]
+        assert isinstance(spatialdata_container_format, SpatialDataContainerFormatType)
 
         store = None
 
         if zarr_group is None:
             assert self.is_backed(), "The SpatialData object must be backed by a Zarr store to write attrs."
-            store = parse_url(self.path, mode="r+").store
-            zarr_group = zarr.group(store=store, overwrite=False)
+            store = parse_url(self.path, mode="r+", fmt=FormatV05()).store
```
LucaMarconato: to replace `parse_url` with the new method.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1680,6 +1669,8 @@ def write_metadata(
         if write_attrs:
             self.write_attrs()
 
+        # TODO: discuss when has_consolidated_metadata that we should just consolidate it because after a writing
+        # operation the consolidated store could otherwise be out of sync.
```
LucaMarconato: Difficult to read. What about:

> TODO: Consider always consolidating metadata after a write operation if has_consolidated_metadata is True,
>      to prevent the consolidated store from becoming out of sync.

src/spatialdata/_core/spatialdata.py
```diff
@@ -2143,14 +2062,14 @@ def _gen_spatial_element_values(self) -> Generator[SpatialElement, None, None]:
             yield from d.values()
 
     def _gen_elements(
```
LucaMarconato, line 2100: if it's a private function, I'd add the deprecation warning not to break existing code in the spatialdata ecosystem (e.g. spatialdata-plot etc)

src/spatialdata/_core/spatialdata.py
LucaMarconato: Not relevant anymore.

src/spatialdata/_io/io_shapes.py
```diff
@@ -77,6 +77,9 @@ def write_shapes(
 ) -> None:
     """Write shapes to spatialdata zarr store.
 
+    Note that the parquet file is not recognized as part of the zarr hierarchy as it is not a valid component of a
```
LucaMarconato, line 83: I'd try to add the `zarr.json` file or `.group` to the `parquet` path (it's a folder, not a file), and see if this fixes the warning. It's fine for the `parquet` folder to be also a Zarr group.

src/spatialdata/_io/io_zarr.py
```diff
@@ -305,5 +306,9 @@ def _group_for_element_exists(zarr_path: Path, element_type: str, element_name:
 def _write_consolidated_metadata(path: Path | str | None) -> None:
     if path is not None:
         f, f_store_path = _open_zarr(path, mode="r+", use_consolidated=False)
-        zarr.consolidate_metadata(f.store)
+        # .parquet files are not recognized as proper zarr and thus throw a warning. This does not affect SpatialData.
```
LucaMarconato, line 280: If we turn the parquet folder into a Zarr group the problem should disappear, and it would make remote read easier, since developers would see the content of the parquet folder even without the need for parquet APIs.

src/spatialdata/_core/spatialdata.py
```diff
@@ -296,9 +296,10 @@ def get_instance_key_column(table: AnnData) -> pd.Series:
     def set_channel_names(self, element_name: str, channel_names: str | list[str], write: bool = False) -> None:
         """Set the channel names for an image `SpatialElement` in the `SpatialData` object.
 
-        This method assumes that the `SpatialData` object and the element are already stored on disk as it will
-        also overwrite the channel names metadata on disk. In case either the `SpatialData` object or the
-        element are not stored on disk, please use `SpatialData.set_image_channel_names` instead.
+        This method will overwrite the element in memory with the same element, but with new channel names.
+        If 'write` is 'True', this method assumes that the `SpatialData` object and the element are already stored on
+        disk as it will also overwrite the channel names metadata on disk. In case either the `SpatialData` object or
```
LucaMarconato: Comment regarding "In case either the `SpatialData` object or the element are not stored on disk, please use `SpatialData.set_image_channel_names` instead.". 

- `set_image_channel_names` does not exist. It is `spatialdata.models.set_channel_names`
- since it should be equivalent to calling this function with `write=False`, I would not confuse the user by telling to use that function instead

src/spatialdata/_core/spatialdata.py
```diff
@@ -307,9 +308,11 @@ def set_channel_names(self, element_name: str, channel_names: str | list[str], w
         channel_names
             The channel names to be assigned to the c dimension of the image `SpatialElement`.
         write
-            Whether to overwrite the channel metadata on disk.
+            Whether to overwrite the channel metadata on disk. This will not rewrite the pixel data itself.
```
LucaMarconato: For extra clarity, I'd change this:
> Whether to overwrite the channel metadata on disk. This will not rewrite the pixel data itself.
into this
> Whether to overwrite the channel metadata on disk (lightweight operation). This will not rewrite the pixel data itself (heavy operation).

src/spatialdata/_core/spatialdata.py
```diff
@@ -307,9 +308,11 @@ def set_channel_names(self, element_name: str, channel_names: str | list[str], w
...
         """
-        self.images[element_name] = set_channel_names(self.images[element_name], channel_names)
+        with warnings.catch_warnings():
```
LucaMarconato: I see the problem with the warning now. I think we should stop throwing a warning when calling `sdata['my_image'] = new_image`, otherwise it can get cumbersome to have to do things like

```python
old_image = sdata['my_image']
del sdata['my_image']
sdata['my_image'] = some_function(old_image)
```

instead of just

```python
sdata['my_image'] = some_function(sdata['my_image'])
```

tests/core/operations/test_rasterize.py
```diff
@@ -158,8 +158,11 @@ def test_rasterize_points_shapes_with_string_index(points, shapes):
     sdata = SpatialData.init_from_elements({"points_0": points["points_0"], "circles": shapes["circles"]})
 
     # make the indices of the points_0 and circles dataframes strings
-    sdata["points_0"]["str_index"] = dd.from_pandas(pd.Series([str(i) for i in sdata["points_0"].index]), npartitions=1)
-    sdata["points_0"] = sdata["points_0"].set_index("str_index")
+    points = sdata["points_0"]
```
LucaMarconato, line 161: yeah, this is another argument for removing the warning on in-place in-memory overwrite of the elements.

tests/io/test_readwrite.py
```diff
@@ -63,7 +63,9 @@ def test_points(self, tmp_path: str, points: SpatialData) -> None:
 
         # check the index is correctly written and then read
         new_index = dd.from_array(np.arange(1, len(points["points_0"]) + 1))
-        points["points_0"] = points["points_0"].set_index(new_index)
+        el_point = points["points_0"].set_index(new_index)
```
LucaMarconato: same here, the syntax get more verbose unnecessarily

tests/io/test_readwrite.py
```diff
@@ -337,6 +339,7 @@ def test_overwrite_fails_when_no_zarr_store_bug_dask_backed_data(self, full_sdat
                 # now we have a sdata with dask-backed elements
                 sdata2 = SpatialData.read(f)
                 p = sdata2[element]
+                del full_sdata[element]
```
LucaMarconato: same here, the syntax get more verbose unnecessarily

tests/utils/test_testing.py
```diff
@@ -46,11 +46,14 @@ def _change_metadata_tables(sdata: SpatialData, element_name: str) -> None:
 def _change_metadata_image(sdata: SpatialData, element_name: str, coords: bool, transformations: bool) -> None:
     if coords:
         if isinstance(sdata[element_name], DataArray):
-            sdata[element_name] = sdata[element_name].assign_coords({"c": np.array(["m", "l", "b"])})
+            element = sdata[element_name].assign_coords({"c": np.array(["m", "l", "b"])})
```
LucaMarconato: same here, the syntax get more verbose unnecessarily

tests/utils/test_testing.py
```diff
@@ -46,11 +46,14 @@ def _change_metadata_tables(sdata: SpatialData, element_name: str) -> None:
...
+            del sdata[element_name]
+            sdata[element_name] = element
         else:
             assert isinstance(sdata[element_name], DataTree)
 
             dt = sdata[element_name].assign_coords({"c": np.array(["m", "l", "b"])})
+            del sdata[element_name]
```
LucaMarconato: same here, the syntax get more verbose unnecessarily

src/spatialdata/_core/spatialdata.py
LucaMarconato: After removing the warning, we should cleanup this code.

tests/core/operations/test_rasterize.py
LucaMarconato: After removing the warning, we should remove `del` from this code. I still like to define `points = sdata["points_0"]`, it makes things a bit clearer.

tests/io/test_readwrite.py
LucaMarconato: After removing the warning, we should restore the old code.

tests/io/test_readwrite.py
LucaMarconato: After removing the warning, we should restore the old code.

tests/utils/test_testing.py
LucaMarconato: After removing the warning, we should restore the old code.

tests/utils/test_testing.py
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -35,8 +35,11 @@ def test_join_using_string_instance_id_and_index(sdata_query_aggregation):
         [f"string_{i}" for i in sdata_query_aggregation["values_polygons"].index]
     )
 
-    sdata_query_aggregation["values_polygons"] = sdata_query_aggregation["values_polygons"][:5]
-    sdata_query_aggregation["values_circles"] = sdata_query_aggregation["values_circles"][:5]
+    values_polygons = sdata_query_aggregation["values_polygons"][:5]
+    values_circles = sdata_query_aggregation["values_circles"][:5]
+    del sdata_query_aggregation["values_polygons"], sdata_query_aggregation["values_circles"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -85,7 +88,9 @@ def test_left_inner_right_exclusive_join(sdata_query_aggregation):
     assert table is None
     assert all(element_dict[key] is None for key in element_dict)
 
-    sdata["values_polygons"] = sdata["values_polygons"].drop([10, 11])
+    values_polygons = sdata["values_polygons"].drop([10, 11])
+    del sdata["values_polygons"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -147,7 +152,9 @@ def test_left_inner_right_exclusive_join(sdata_query_aggregation):
     assert "by_polygons" in element_dict
 
     # check multiple elements joined to table.
-    sdata["values_circles"] = sdata["values_circles"].drop([7, 8])
+    values_circles = sdata["values_circles"].drop([7, 8])
+    del sdata["values_circles"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -289,7 +296,9 @@ def test_left_exclusive_and_right_join(sdata_query_aggregation):
     assert table is None
 
     # Dropped indices correspond to instance ids 7, 8 for 'values_circles' and 10, 11 for 'values_polygons'
-    sdata["table"] = sdata["table"][sdata["table"].obs.index.drop(["7", "8", "19", "20"])]
+    table_update = sdata["table"][sdata["table"].obs.index.drop(["7", "8", "19", "20"])]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -372,7 +381,9 @@ def test_left_exclusive_and_right_join(sdata_query_aggregation):
 
 def test_match_rows_inner_join_non_matching_element(sdata_query_aggregation):
     sdata = sdata_query_aggregation
-    sdata["values_circles"] = sdata["values_circles"][4:]
+    circles = sdata["values_circles"][4:]
+    del sdata["values_circles"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -403,6 +414,7 @@ def test_match_rows_inner_join_non_matching_table(sdata_query_aggregation):
     original_instance_id = table.obs["instance_id"]
     reversed_instance_id = [6, 7, 8, 3, 4, 5] + list(reversed(range(12)))
     table.obs["instance_id"] = reversed_instance_id
+    del sdata["table"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -433,8 +445,11 @@ def test_match_rows_inner_join_non_matching_table(sdata_query_aggregation):
 def test_inner_join_match_rows_duplicate_obs_indices(sdata_query_aggregation: SpatialData, join_type: str) -> None:
     sdata = sdata_query_aggregation
     sdata["table"].obs.index = ["a"] * sdata["table"].n_obs
-    sdata["values_circles"] = sdata_query_aggregation["values_circles"][:4]
-    sdata["values_polygons"] = sdata_query_aggregation["values_polygons"][:5]
+    values_circles = sdata_query_aggregation["values_circles"][:4]
+    values_polygons = sdata_query_aggregation["values_polygons"][:5]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -824,6 +839,7 @@ def test_get_values_df_points(points):
         region_key="region",
         instance_key="instance_id",
     )
+    del points["points_0"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_relational_query.py
```diff
@@ -906,6 +922,7 @@ def test_filter_table_non_annotating(full_sdata):
     obs = pd.DataFrame({"test": ["a", "b", "c"]}, index=list(map(str, range(3))))
     adata = AnnData(obs=obs)
     table = TableModel.parse(adata)
+    del full_sdata["table"]
```
LucaMarconato: After removing the warning, we should restore the old code.

tests/core/query/test_spatial_query.py
```diff
@@ -751,6 +751,7 @@ def test_spatial_query_different_axes(full_sdata, name: str):
         if name == "multipoly":
             new_data = GeoDataFrame({"geometry": [MultiPolygon([Polygon([(3, 1), (4, 1), (3, 0)])])]})
         gdf = pd.concat([gdf, new_data], ignore_index=True)
+        del full_sdata[name]
```
LucaMarconato: After removing the warning, we should restore the old code.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1052,7 +1052,6 @@ def _validate_can_safely_write_to_path(
             raise ValueError(f"file_path must be a string or a Path object, type(file_path) = {type(file_path)}.")
 
         if os.path.exists(file_path):
-            _open_zarr_store(file_path, mode="r")
             if parse_url(file_path, mode="r", fmt=FormatV05()) is None:
                 raise ValueError(
                     "The target file path specified already exists, and it has been detected to not be a Zarr store. "
```
LucaMarconato, line 1059: Now that we modified the implementation of this crucial function, we need to add a test for raising this exception. And for all the other exceptions in this function.

src/spatialdata/_io/io_zarr.py
```diff
@@ -74,7 +70,10 @@ def read_zarr(
     -------
     A SpatialData object.
     """
-    f, f_store_path = _open_zarr_store(store)
+    from spatialdata._io._utils import _resolve_zarr_store as rzs
```
LucaMarconato: you call the function only once, so no need to call it `rzs`

src/spatialdata/_io/io_zarr.py
```diff
@@ -261,19 +260,16 @@ def _get_groups_for_element(
...
-    from spatialdata._io._utils import _open_zarr_store
-
-    store = _open_zarr_store(zarr_path, mode="r+")
+    resolved_store = _resolve_zarr_store(zarr_path)
 
     # When writing, use_consolidated must be set to False. Otherwise, the metadata store
     # can get out of sync with newly added elements (e.g., labels), leading to errors.
-    root = zarr.open_group(store=store, mode="a", use_consolidated=use_consolidated)
-    element_type_group = root.require_group(element_type)
+    root_group = zarr.open_group(store=resolved_store, mode="a", use_consolidated=use_consolidated)
```
LucaMarconato: The first group should be opened with `r+`, not `a`, because we want to fail if the store does not exist (this is what the docstring promises). The second group is correctly opened with `a`, because we want to create it if it doesn't exist (and we do not want to overwrite).

src/spatialdata/_core/spatialdata.py
```diff
@@ -2275,30 +2198,13 @@ def init_from_elements(
                 assert model == ShapesModel
                 element_type = "shapes"
             elements_dict.setdefault(element_type, {})[name] = element
-        # when the "tables" argument is removed, we can remove all this if block
```
LucaMarconato, line 2278: we need to remove `tables` from the arguments and from the docstring. And we need to add a deprecation warnings/decorator to be backward compatible.

LucaMarconato (commented):
Partial review. All easy fixes, in general good work!

src/spatialdata/_core/spatialdata.py
melonora: ok will remove from the docstring, but the deprecation warning was there already for a year

src/spatialdata/_core/_elements.py
```diff
@@ -42,7 +42,13 @@ def _remove_shared_key(self, key: str) -> None:
     def _check_key(key: str, element_keys: Iterable[str], shared_keys: set[str | None]) -> None:
         check_valid_name(key)
         if key in element_keys:
-            warn(f"Key `{key}` already exists. Overwriting it in-memory.", UserWarning, stacklevel=2)
+            warn(
```
LucaMarconato: Commented elsewhere. Let's remove the warning here.

src/spatialdata/_core/spatialdata.py
```diff
@@ -16,8 +16,8 @@
 from dask.dataframe import read_parquet
 from dask.delayed import Delayed
 from geopandas import GeoDataFrame
+from ome_zarr.format import FormatV05
```
LucaMarconato: After all the comments below are addressed, there should be no need for this import.

src/spatialdata/_io/format.py
```diff
@@ -135,7 +195,19 @@ def version(self) -> str:
...
+class RasterFormatV03(FormatV05, CoordinateMixinV01):
+    @property
+    def spatialdata_format_version(self) -> str:
+        return "0.3"
+
+    @property
+    def version(self) -> str:
+        # 0.1 -> 0.2 changed the version string for the NGFF format, from 0.4 to 0.6-dev-spatialdata as discussed here
+        # https://github.com/scverse/spatialdata/pull/849
+        return "0.4-dev-spatialdata"
```
LucaMarconato: This should be 0.5-dev-spatialdata

src/spatialdata/_io/_utils.py
```diff
@@ -59,6 +63,18 @@ def _get_transformations_from_ngff_dict(
 def overwrite_coordinate_transformations_non_raster(
     group: zarr.Group, axes: tuple[ValidAxis_t, ...], transformations: MappingToCoordinateSystem_t
 ) -> None:
+    """Write coordinate transformations of non-raster element to disk.
+
+    Parameters
+    ----------
+    group: zarr.Group
```
LucaMarconato, line 57: sphinx already add types in the docs, so no need to add them here. Same comments for 
- overwrite_coordinate_transformations_raster
- _overwrite_coordinate_transformations_raster_zarrv3
- _overwrite_coordinate_transformations_raster_zarrv2
- _resolve_zarr_store (here the type in the comment is even different from the docstring)

src/spatialdata/_io/io_zarr.py
melonora: addressed

src/spatialdata/_io/_utils.py
```diff
@@ -90,15 +124,58 @@ def overwrite_coordinate_transformations_raster(
...
+        The zarr group containing the raster element for which to write the transformations, e.g. the zarr group
+        containing sdata['image2d'].
+    coordinate_transformations: list[dict[str, BaseTransformation]]
+        List of NGFF transformation representations as dictionaries.
+    """
+    if len(multiscales := group.metadata.attributes["ome"]["multiscales"]) != 1:
```
LucaMarconato: the difference between zarr v2 zarr v3 is in `group.attrs` vs `group.metadata.attribute['ome']`. I think having 2 functions for just this difference is an overhead we can avoid. I would therefore go back to having a single function where we have a single if else.

src/spatialdata/_io/_utils.py
```diff
@@ -110,16 +187,9 @@ def overwrite_channel_names(group: zarr.Group, element: DataArray | DataTree) ->
         channel_names = element["scale0"]["image"].coords["c"].data.tolist()
 
     channel_metadata = [{"label": name} for name in channel_names]
-    omero_meta = group.attrs["omero"]
+    omero_meta = group.attrs["ome"]["omero"]
```
LucaMarconato: This seems to be the version of 0.5, without the 0.4 version. Let's add tests to cover this.

The fix in the code should be an `if else` to check the version, and then do
- omero_meta = group.attrs["omero"]  # for 0.4
- omero_meta = group.attrs["omero"]  # for 0.5

src/spatialdata/_io/_utils.py
melonora: renamed to `_resolve_zarr_store` and then we can update it in a next PR.

src/spatialdata/_io/_utils.py
```diff
@@ -363,24 +434,70 @@ def _is_element_self_contained(
...
-        The SpatialData object
+    This function accepts various forms of input (e.g. filesystem paths,
+    UPath objects, existing Zarr stores, or `zarr.Group`s) and resolves
+    them into a `StoreLike` that can be passed to Zarr APIs. It handles
+    local files, fsspec-backed stores, consolidated metadata stores, and
+    groups with nested paths.
+
+    Parameters
+    ----------
+    path : StoreLike | str | Path | UPath | zarr.Group
```
LucaMarconato: the signature of `_resolve_zarr_store()` is different and should match the types in the docstrings. Then, the types should be removed from the docstring (sphinx infers them).

src/spatialdata/_io/io_zarr.py
melonora: changed to `r+` for now, though it is technically already created prior to access of this function always.

src/spatialdata/_io/_utils.py
LucaMarconato: @ilan-gold can you expand on this please:

> If you're interested in zarrs-python, I would recommend running your test suite locally with it

LucaMarconato: We do not write to Zarr directly, but we delegate this to `ome-zarr-py`. Also, `multiscale-spatial-image` has its own `to_zarr()` function, that we do not use.

Maybe `zarrs-python` could be used in those libraries instead of `zarr-python`?

src/spatialdata/_io/io_zarr.py
melonora: This is not redundant, important here is that the zarr group must be opened with "use_consolidated" being "False" when writing or you will get errors. This is because the zarr group is accessed multiple times while writing one element and then in between partial writes the consolidated metadata store of the group is out of sync. This is not required for `element_name_group` as this is not used for the consolidated metadata check.

src/spatialdata/_io/_utils.py
```diff
@@ -363,24 +434,70 @@ def _is_element_self_contained(
...
+
+    if isinstance(path, zarr.Group):
+        # if the input is a zarr.Group, wrap it with a store
+        if isinstance(path.store, LocalStore):
+            # create a simple FSStore if the store is a LocalStore with just the path
+            return FsspecStore(os.path.join(path.store.path, path.path), **kwargs)
+        if isinstance(path.store, FsspecStore):
+            # if the store within the zarr.Group is an FSStore, return it
+            # but extend the path of the store with that of the zarr.Group
+            return FsspecStore(path.store.path + "/" + path.path, fs=path.store.fs, **kwargs)
```
LucaMarconato, line 458: can we replace this (and hte `os.path.join` above), to UPath APIs?

src/spatialdata/_io/format.py
```diff
@@ -112,6 +100,78 @@ def validate_coordinate_transformations(
...
+        if Points_s.ATTRS_KEY in data:
+            if Points_s.INSTANCE_KEY in data[Points_s.ATTRS_KEY]:
+                d[Points_s.INSTANCE_KEY] = data[Points_s.ATTRS_KEY][Points_s.INSTANCE_KEY]
+            if Points_s.FEATURE_KEY in data[Points_s.ATTRS_KEY]:
+                d[Points_s.FEATURE_KEY] = data[Points_s.ATTRS_KEY][Points_s.FEATURE_KEY]
+        return d
+
+
+class TableValidateMixinV01:
+    def validate_table(
```
LucaMarconato: I think we can remove the TableValidateMixinV01 altogether and just call TableModel.validate() when validate_table is called (its only called in `write_table()`).

It could be that this check is not available in `TableModel.validate()`:
```python
  if region_key is not None and not isinstance(table.obs[region_key].dtype, CategoricalDtype):
            raise ValueError(
                f"`table.obs[region_key]` must be of type `categorical`, not `{type(table.obs[region_key])}`."
            )
```

In such case, we should add it.

src/spatialdata/_io/format.py
```diff
@@ -173,106 +250,139 @@ def attrs_to_dict(self, data: dict[str, Any]) -> dict[str, str | dict[str, Any]]
...
     "0.1": RasterFormatV01(),
     "0.2": RasterFormatV02(),
+    "0.3": RasterFormatV03(),
 }
-SpatialDataContainerFormats = {
+SpatialDataContainerFormats: dict[str, SpatialDataContainerFormatType] = {
     "0.1": SpatialDataContainerFormatV01(),
+    "0.2": SpatialDataContainerFormatV02(),
+}
+ContainerFormatValidElements = {
```
LucaMarconato, line 315: Great!

src/spatialdata/_io/io_raster.py
```diff
@@ -36,40 +33,65 @@
...
+    ----------
+    image_nodes: list[Node]
+        List of nodes returned from the ome-zarr-py Reader.
+    nodes: list[Node]
+        List to append the nodes with the multiscales spec to.
+
+    Returns
+    -------
+    list[Node]
+        List of nodes with the multiscales spec.
```
LucaMarconato: please remove the return, no need for it

src/spatialdata/_io/io_raster.py
```diff
@@ -121,33 +141,41 @@ def _write_raster(
...
-    # _get_group_for_writing_data()
-    if raster_type == "image":
-        assert label_metadata is None
-    else:
+    """Write raster data to disk.
+
+    Parameters
+    ----------
+    raster_type: Literal["image", "labels"]
+        Whether the raster data pertains to a image or labels 'SpatialElement`.
```
LucaMarconato: types to be removed from the docstring

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+def _write_raster_dataarray(
+    raster_type: Literal["image", "labels"],
+    group: zarr.Group,
+    element_name: str,
+    raster_data: DataArray,
+    raster_format: Format,
+    storage_options: JSONDict | list[JSONDict] | None = None,
+    **metadata: str | JSONDict | list[JSONDict],
+) -> None:
+    """Write raster data of type DataArray to disk.
```
LucaMarconato, line 244: types to be removed from the docstring

src/spatialdata/_io/io_raster.py
```diff
@@ -121,33 +141,41 @@ def _write_raster(
     raster_data: DataArray | DataTree,
     group: zarr.Group,
     name: str,
-    format: Format = CurrentRasterFormat(),
+    raster_format: Format,
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
     # triggers the write operation
     group.attrs[ATTRS_KEY] = attrs
 
 
+def _write_raster_dataarray(
+    raster_type: Literal["image", "labels"],
+    group: zarr.Group,
+    element_name: str,
+    raster_data: DataArray,
+    raster_format: Format,
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_zarr.py
melonora: renamed to `resolve_zarr_store` as the name of the function in `_utils.py` did not reflect what the function actually does.

src/spatialdata/_io/_utils.py
melonora: I think this would be work for a future PR.

src/spatialdata/_io/io_zarr.py
melonora: resolve_zarr_store is literally only returning a StoreLike object. It does not create anything, there is a mode that can be passed on for readonly, but given that we still open specific zarr groups after this, there is not much of an effect.

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+        _write_raster_datatree(raster_type, group, name, raster_data, raster_format, storage_options, **metadata)
     else:
         raise ValueError("Not a valid labels object")
 
-    # as explained in a comment in format.py, since coordinate transformations are not part of NGFF yet, we need to have
-    # our spatialdata extension also for raster type (eventually it will be dropped in favor of pure NGFF). Until then,
-    # saving the NGFF version (i.e. 0.4) is not enough, and we need to also record which version of the spatialdata
-    # format we are using for raster types
-    group = _get_group_for_writing_transformations()
+    # Since NGFF does not yet support coordinate transformations, we need a SpatialData extension for rasters. This will
```
LucaMarconato: the code that this comment refers to "Since NGFF does not yet support coordinate transformations, we need a SpatialData extension for rasters. This will be dropped once NGFF supports it. For now, saving the NGFF version (0.4) is not enough", has been moved away from this function. Let's remove this part of the comment.

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+    raster_format: Format,
+    storage_options: JSONDict | list[JSONDict] | None = None,
+    **metadata: str | JSONDict | list[JSONDict],
+) -> None:
+    """Write raster data of type DataArray to disk.
+
+    Parameters
+    ----------
+    raster_type: Literal["image", "labels"]
+        Whether the raster data pertains to a image or labels 'SpatialElement`.
```
LucaMarconato, line 249: to remove the types from the docstring

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+    storage_options: JSONDict | list[JSONDict] | None = None,
+    **metadata: str | JSONDict | list[JSONDict],
+) -> None:
+    """Write raster data of type DataTree to disk.
+
+    Parameters
+    ----------
+    raster_type: Literal["image", "labels"]
+        Whether the raster data pertains to a image or labels 'SpatialElement`.
+    group: zarr.Group
```
LucaMarconato: to remove the types from the docstring

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+    input_axes: tuple[str, ...] = tuple(raster_data.dims)
+    chunks = raster_data.chunks
+    parsed_axes = _get_valid_axes(axes=list(input_axes), fmt=raster_format)
+    if storage_options is not None:
+        if "chunks" not in storage_options and isinstance(storage_options, dict):
+            storage_options["chunks"] = chunks
+    else:
+        storage_options = {"chunks": chunks}
+    # Scaler needs to be None since we are passing the data already downscaled for the multiscale case.
+    metadata[raster_type] = data
```
LucaMarconato, line 280: This line is obscure without the comment that got removed! I will restore it.

> We need this because the argument of write_image_ngff is called image while the argument of
> write_labels_ngff is called label.

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+        group=trans_group, transformations=transformations, axes=tuple(input_axes)
+    )
+
+
 def write_image(
     image: DataArray | DataTree,
     group: zarr.Group,
     name: str,
-    format: Format = CurrentRasterFormat(),
+    element_format: Format = CurrentRasterFormat(),
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_raster.py
```diff
@@ -254,7 +347,7 @@ def write_labels(
     labels: DataArray | DataTree,
     group: zarr.Group,
     name: str,
-    format: Format = CurrentRasterFormat(),
+    element_format: Format = CurrentRasterFormat(),
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_raster.py
```diff
@@ -121,33 +141,41 @@ def _write_raster(
...
+    ----------
+    raster_type: Literal["image", "labels"]
+        Whether the raster data pertains to a image or labels 'SpatialElement`.
+    raster_data: DataArray | DataTree
+        The raster data to write.
+    group: zarr.Group
+        The zarr group in the 'image' or 'labels' zarr group to write the raster data to.
+    name: str
+        The name of the raster element.
+    raster_format: Format
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+    ----------
+    raster_type: Literal["image", "labels"]
+        Whether the raster data pertains to a image or labels 'SpatialElement`.
+    group: zarr.Group
+        The zarr group in the 'image' or 'labels' zarr group to write the raster data to.
+    element_name: str
+        The name of the raster element.
+    raster_data: DataArray
+        The raster data to write.
+    raster_format: Format
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+    trans_group = group["labels"][element_name] if raster_type == "labels" else group
+    overwrite_coordinate_transformations_raster(group=trans_group, transformations=transformations, axes=input_axes)
+
+
+def _write_raster_datatree(
+    raster_type: Literal["image", "labels"],
+    group: zarr.Group,
+    element_name: str,
+    raster_data: DataTree,
+    raster_format: Format,
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_io/io_raster.py
```diff
@@ -156,86 +184,151 @@ def _get_group_for_writing_transformations() -> zarr.Group:
...
+    ----------
+    raster_type: Literal["image", "labels"]
+        Whether the raster data pertains to a image or labels 'SpatialElement`.
+    group: zarr.Group
+        The zarr group in the 'image' or 'labels' zarr group to write the raster data to.
+    element_name: str
+        The name of the raster element.
+    raster_data: DataTree
+        The raster data to write.
+    raster_format: Format
```
LucaMarconato: let's use `RasterFormatType` instead

src/spatialdata/_core/spatialdata.py
melonora: No you cannot do this as the function is called also from a static method that requires passing on the `path` argument.

melonora: sdata.write_consolidated_metadata does not require this.

src/spatialdata/_core/spatialdata.py
melonora: It cannot be replaced here as _resolve_zarr_store does not initialize the zarr store itself, which `parse_url` does, e.g. `parse_url` already creates initial metadata conforming the zarr store to an ome-zarr store.

src/spatialdata/_core/spatialdata.py
melonora: Do you know whether opening the zarr store with mode `w` actually makes sure the zarr store is initially empty?

src/spatialdata/_core/spatialdata.py
melonora: did add deprecation alias for `formats` -> `sdata_formats`

src/spatialdata/_core/spatialdata.py
melonora: no, these methods are all from private modules and not made accessible in public api

src/spatialdata/_core/spatialdata.py
melonora: not certain if we can, because this resolves the format which _resolve_zarr_store does not do.

src/spatialdata/_core/spatialdata.py
melonora: Added checks in `test_container_v1_v2` for both container format versions. Both pass.

src/spatialdata/_io/io_shapes.py
melonora: I would not do this, it goes beyond the zarr library and then we now also need to check which version of zarr is actually being used in order to determine how to write this.

src/spatialdata/_io/io_zarr.py
melonora: lets discuss this also taking zarr versions into account.

tests/core/operations/test_rasterize.py
melonora: warning is now removed and all code is addressed accordingly

tests/core/query/test_relational_query.py
melonora: done

src/spatialdata/_core/spatialdata.py
melonora: well I would still do it for this PR so we can discuss haha and then remove this

src/spatialdata/_core/spatialdata.py
melonora: no I think we should just test all the repos and have the PRs and with that releases ready to go. Users do not have control in those repos over this so we should take care of it.

src/spatialdata/_io/_utils.py
melonora: good catch, before it was more different, but I noticed the coordinate transformations were written to wrong location. Refactored back to 1 function now.

src/spatialdata/_io/io_raster.py
melonora: done

src/spatialdata/_core/spatialdata.py
melonora: not certain, would have to check with opening particular stores.

src/spatialdata/_io/_utils.py
melonora: This was already matching, Storelike in the type hint encompasses the Union in the docstring, so just removed the docstring typehint.

src/spatialdata/_io/_utils.py
melonora: would check this in the remote PR instead as right now the code base would not land here.

src/spatialdata/_io/io_raster.py
melonora: done

src/spatialdata/_io/io_raster.py
melonora: done

src/spatialdata/_io/io_raster.py
melonora: removed

src/spatialdata/_io/io_raster.py
melonora: done

src/spatialdata/_io/io_raster.py
melonora: done

src/spatialdata/_io/io_raster.py
melonora: not certain why it does not show in the file, but the comment was / is there all along:
check 48f0b81b071274da851a0678695d3dabc224c9a4

src/spatialdata/_io/format.py
melonora: I would keep it here, exploring the refactor you run in the situation that a lot of functions do not assume that `region_key` column is a categorical one. For example when just using `get_values`. This will immediately lead to errors. It also does not make sense for the function to require it to be like that.

src/spatialdata/_io/_utils.py
melonora: done

src/spatialdata/_io/io_zarr.py
```diff
@@ -71,112 +71,111 @@ def read_zarr(
...
+    selector = {"images", "labels", "points", "shapes", "tables"} if not selection else set(selection or [])
     logger.debug(f"Reading selection {selector}")
 
-    # read multiscale images
-    if "images" in selector and "images" in f:
-        with handle_read_errors(
-            on_bad_files,
-            location="images",
-            exc_types=(JSONDecodeError, MetadataError),
-        ):
```
LucaMarconato, line 92: can we restore this with the new error (OSError?)?

src/spatialdata/_io/io_zarr.py
```diff
@@ -71,112 +71,111 @@ def read_zarr(
...
             logger.debug(f"Found {count} elements in {group}")
 
-    # read multiscale labels
-    with ome_zarr_logger(logging.ERROR):
-        if "labels" in selector and "labels" in f:
-            with handle_read_errors(
-                on_bad_files,
-                location="labels",
-                exc_types=(JSONDecodeError, MetadataError),
-            ):
```
LucaMarconato: Same as for images: can we restore this with the new error type (OSError?)?

src/spatialdata/_io/format.py
melonora: ok still did the refactor

src/spatialdata/_io/io_zarr.py
```diff
@@ -5,21 +5,25 @@
...
-from spatialdata._io._utils import BadFileHandleMethod, handle_read_errors, ome_zarr_logger
+from spatialdata._io._utils import BadFileHandleMethod, _resolve_zarr_store, handle_read_errors, ome_zarr_logger
 from spatialdata._io.io_points import _read_points
 from spatialdata._io.io_raster import _read_multiscale
 from spatialdata._io.io_shapes import _read_shapes
 from spatialdata._io.io_table import _read_table
 from spatialdata._logging import logger
 
 
-def _open_zarr_store(store: str | Path | zarr.Group) -> tuple[zarr.Group, str]:
```
LucaMarconato, line 22: I removed this function! It was used only in parts of the code: `read_zarr()` and `_write_consolidated_metadata()`. The latter was calling `_open_zarr_store()` passing a `str | Path`. The former in principle could pass a `zarr.Group`, but this was never done in the whole codebase, nor tests. This means that the `res = x if y else z` line in `_open_zarr_store()` had a branch that was never used, and that I could easily drop the function.

src/spatialdata/_types.py
```diff
@@ -12,3 +15,5 @@
 
 Raster_T = DataArray | DataTree
 ColorLike = tuple[float, ...] | str
+
+StoreLike: TypeAlias = str | Path | UPath | zarr.storage.StoreLike | zarr.Group
```
LucaMarconato: This `StoreLike` is used only once in the codebase, and it could confuse the user by misinterpreting it as `zarr.storage.StoreLike`. I will therefore remove this type.

src/spatialdata/_core/spatialdata.py
LucaMarconato: The problem is that here we assume that the format is `Formatv05`, which may not be the case.

melonora: confirmed logic is not in _resolve_zarr_store.

src/spatialdata/_io/_utils.py
```diff
@@ -187,9 +145,13 @@ def overwrite_channel_names(group: zarr.Group, element: DataArray | DataTree) ->
         channel_names = element["scale0"]["image"].coords["c"].data.tolist()
 
     channel_metadata = [{"label": name} for name in channel_names]
-    omero_meta = group.attrs["ome"]["omero"]
+    omero_meta = group.attrs.get("omero", None) or group.attrs.get("ome", {}).get("omero")
```
LucaMarconato, line 154: This assumes that the either `omero` or `ome` are present, and if nothing is available will default to use `['ome']['omero']` (i.e. the NGFF 0.5 version).

Shouldn't we maybe use on of the two depending on a parameter specifying the format? And double check that we are consistent with what we find on disk? In this way if the disk has no metadata we are sure that we will write the correct version.

src/spatialdata/models/models.py
```diff
@@ -1157,8 +1180,9 @@ def parse(
             "instance_key": instance_key,
         }
         adata.uns[cls.ATTRS_KEY] = attr
+        convert_region_column_to_categorical(adata)
```
LucaMarconato, line 1183: I'd modify `convert_region_column_to_categorical()` to return `None`, since it modifies the `adata` argument in-place.

tests/io/test_multi_table.py
```diff
@@ -24,9 +24,14 @@ def test_set_get_tables_from_spatialdata(self, full_sdata: SpatialData, tmp_path
         adata2 = adata0.copy()
         del adata2.obs["region"]
         # fails because either none either all three 'region', 'region_key', 'instance_key' are required
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Region key `region` not in `adata.obs`."):
             full_sdata["not_added_table"] = adata2
 
+        adata3 = adata0.copy()
```
LucaMarconato, line 30: We should also have a test for null values in the instance_key column.

tests/models/test_models.py
```diff
@@ -520,15 +520,9 @@ def test_table_model_invalid_names(self, key: str, attr: str, parse: bool):
...
-        invalid_key = keys[1]
-        key_regex = re.escape(invalid_key)
         df = pd.DataFrame([[None] * len(keys)], columns=keys, index=["1"])
         adata = AnnData(np.array([[0]]), **{attr: df})
-        with pytest.raises(
-            ValueError,
-            match=f"Table contains invalid names(.|\n)*\n  {attr}/{invalid_key}: "
-            + f"Key `{key_regex}` is not unique, or another case-variant of it exists.",
-        ):
+        with pytest.raises(ValueError, match="Table contains invalid names.\nFor renaming, please"):
```
LucaMarconato: Incomplete sentence. I'd actually restore the old error.

LucaMarconato (commented):
New iteration of code review!

tests/io/test_readwrite.py
```diff
@@ -635,6 +598,7 @@ def test_incremental_io_attrs(points: SpatialData) -> None:
 cached_sdata_blobs = blobs()
 
 
+# TODO: make consolidated metadata open cleaner
```
LucaMarconato: can you please make this comment more explicit? I don't understand what you mean.

tests/io/test_readwrite.py
```diff
@@ -790,15 +754,48 @@ def test_reading_invalid_name(tmp_path: Path):
     valid_sdata.write(tmp_path / "data.zarr")
     # Circumvent validation at construction time and check validation happens again at writing time.
     (tmp_path / "data.zarr/points" / points_name).rename(tmp_path / "data.zarr/points" / "has whitespace")
-    (tmp_path / "data.zarr/shapes" / shapes_name).rename(tmp_path / "data.zarr/shapes" / "non-alnum_#$%&()*+,?@")
+    # This one is not allowed on windows
+    if os.name != "nt":
+        (tmp_path / "data.zarr/shapes" / shapes_name).rename(tmp_path / "data.zarr/shapes" / "non-alnum_#$%&()*+,?@")
```
LucaMarconato: rather than skipping the test on windows, I would identify the characters that make this fail, and remove them from the string in the test.

tests/io/test_readwrite.py
```diff
@@ -790,15 +754,48 @@ def test_reading_invalid_name(tmp_path: Path):
...
+    # We do this as the key of the element is otherwise not in the consolidated metadata, leading to an error.
+    valid_sdata.write_consolidated_metadata()
 
     with pytest.raises(ValidationError, match="Cannot construct SpatialData") as exc_info:
         read_zarr(tmp_path / "data.zarr")
 
     actual_message = str(exc_info.value)
     assert "points/has whitespace" in actual_message
-    assert "shapes/non-alnum_#$%&()*+,?@" in actual_message
+    if os.name != "nt":
```
LucaMarconato: Same here as above: rather than skipping the test on windows, I would identify the characters that make this fail, and remove them from the string in the test.

tests/io/test_readwrite.py
```diff
@@ -790,15 +754,48 @@ def test_reading_invalid_name(tmp_path: Path):
...
+        path = Path(tmpdir) / "data.zarr"
+        full_sdata.write(path, consolidate_metadata=False)
+
+        group = zarr.open_group(path, mode="r")
+        assert group.metadata.consolidated_metadata is None
+        second_read = SpatialData.read(path)
+        assert_spatial_data_objects_are_identical(full_sdata, second_read)
+
+
+def test_can_read_sdata_with_reconsolidation(full_sdata):
```
LucaMarconato: this test should also be done for SpatialDataContainerFormat01. I'd add a pytest.mark.parametrize to check for both. For Zarr v2 we need to check .zattrs, not zarr.json.

tests/conftest.py
```diff
@@ -84,6 +84,11 @@ def tables() -> list[AnnData]:
     return _tables
 
 
+@pytest.fixture  # (params=['images', 'labels', 'shapes', 'points', 'tables'])
```
LucaMarconato: what's this? 👀 It doesn't seem to be used in the codebase.

pyproject.toml
```diff
@@ -99,6 +100,12 @@ addopts = [
     "--import-mode=importlib",  # allow using test files with same name
     "-s"  # print output from tests
 ]
+# These are all markers coming from xarray, dask or anndata. Added here to silence warnings.
+markers = [
+    "slow: marks tests as slow (deselect with '-m \"not slow\"')",
+    "gpu: run test on GPU using CuPY.",
+    "skip_with_pyarrow_strings: skipwhen pyarrow string conversion is turned on",
+]
```
LucaMarconato: Add from @ilan-gold, but never used?

src/spatialdata/_io/io_zarr.py
melonora: postponing checking this in the remote PR after discussion. Added a TODO in the codebase

LucaMarconato: We synced offline. The warning does not disappear. Let's filter it, and address it later on.
Additional context: the NGFF mesh PR will also suffer from this problem because it would store binary file from the neuroglancer precomputed format inside Zarr, triggering this warning.

src/spatialdata/_core/spatialdata.py
melonora: is removed now.

src/spatialdata/_core/spatialdata.py
LucaMarconato: It overwrites the store, so it becomes empty.

src/spatialdata/_io/_utils.py
melonora: in case of images there is always "omero" channel metadata. Literally the only difference is whether this metadata is at the root level or within the 'ome' key. Given that this is overwritting channel names only in case of images we should be ok.

src/spatialdata/models/models.py
melonora: done

tests/io/test_readwrite.py
melonora: already did this todo, main thing was that we opened the zarr group multiple times which does not work cleanly when having consolidated_metadata. We open only once now. I removed the TODO

tests/io/test_readwrite.py
melonora: done

tests/io/test_readwrite.py
melonora: done

tests/conftest.py
melonora: ah yeah was to refactor one test, but didn't work.

pyproject.toml
melonora: They are never used, but it does spam warnings on import from these libraries. This silences those warnings

tests/conftest.py
melonora: removed

src/spatialdata/_io/io_zarr.py
melonora: so the top level no, the behaviour in case of consolidating is that it will actually use the consolidated metadata to do the lookup. You can remove the `.zattrs`, `.zgroup` or `zarr.json` at either the `images` or `labels` group and it won't have an affect. 
If you dont have consolidated metadata it will just result in the group not being read.

I added the OS error at the level of reading in the individual elements.

melonora:
> * [ ]  `test_overwrite_works_when_no_zarr_store()` fails for NGFF 0.4 because somehow something is written using NGFF 0.5 and a version mismatch error occurs.

You are trying to write an empty SpatialData object. Even though the object itself does not contain elements, still initially `parse_url` already initializes the store, so the first line in the following code snippet in write will already create a zarr object:
```
store = parse_url(file_path, mode="w", fmt=parsed["SpatialData"]).store
zarr_group = zarr.open_group(store=store, mode="w" if overwrite else "a")
self.write_attrs(zarr_group=zarr_group)
```

src/spatialdata/_core/spatialdata.py
LucaMarconato: I made this more robust.

tests/io/test_readwrite.py
LucaMarconato: Done.

tests/models/test_models.py
LucaMarconato: I restored some of the code.

tests/io/test_multi_table.py
LucaMarconato: Added.

src/spatialdata/_io/io_zarr.py
LucaMarconato: I believe this is now addressed.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Added a todo in the thread for this. Closing this discussion since it's redundant.

src/spatialdata/_core/spatialdata.py
LucaMarconato: We fixed this.

src/spatialdata/_io/_utils.py
ilan-gold: I meant here just turning `zarrs-python` on for the entire run of the test suite.

> We do not write to Zarr directly

But it looks like you're opening a store or no? This would be the incompaibility, if you were to try to use the store to do any io and `zarrs` is turned on

melonora: This does not open the store here, it just resolves it. It checks what filesystem to use and the path inside it and other requried options to configure it.

melonora:
I noticed that we were creating a new `multiscales` key outside `ome`. The last commit fixes that and also ensures that the `ome` metadata is properly overwritten. You can't just overwrite a metadata key value pair in `ome`. It will not write. You have to extract the `ome` metadata and then overwrite the whole metadata block if you want it to be persistent.

Furthermore, I have added the functionality of checking with which format the data is written so we can request the correct reader from `ome-zarr`.

melonora:
will reopen the PR as we can't change PR author.

PR #971, opened 2025-08-20
add remote support

melonora:


melonora:
@ap-- @berombau been a long time coming but almost ready to get all the work in regarding remote read and write support. Will add some tests for azure still as well. 

One blocker currently is that we require the newest version of ome-zarr which includes a fix that I implemented for writing images and labels on remote. The problem is that this version is only available with zarr>3. 
We are currently working on getting this zarr v3 support.

Issue #741, opened 2024-10-31
Benchmarks performance on HPC


LucaMarconato:
CC @melonora @berombau 

Working with Zarr files on a HPC may create problems due to Zarr v2. One of the challenges for the hackathon will be to setup a set of benchmarks, which will help evaluate the transition to Zarr v3.

berombau:
Linking to #768, which is more on documenting on how to setup on the HPC. This issue is more on doing benchmarks on HPC once setup and creating datasets for such a benchmark.

PR #985, opened 2025-09-17
support zarrv3

### LINKED ISSUE #979: deprecations and refactors in zarr v3 work.

### LINKED ISSUE #983: Better warnings when reading Zarr v2 data
Before making a release with the support for Zarr v3 we should improve the warnings messages when reading Zarr v2 data, and suggest to the users to rewrite the data to convert to the new format.

We could also explore a conversion of the metadata only, to avoid to rewrite the large images (this is possible, but is some extra work).

melonora:
closes #983 
closes #979 

related to #741 

This PR provides initial support for Zarr v3 in SpatialData. These changes include the following:

- To support the newer Zarr v3 versions, Python support for version 3.10 had to be dropped.
- the use of the `format` argument for writing operations will be deprecated in version 0.7.0 as format is a Python builtin,
  Please use `sdata_formats` instead, argument type stays the same.
- All functions and arguments that were marked for deprecation prior to the current SpatialData version now have been deprecated, vastly reducing the number of warnings for the user.
- Added option of reconsolidating metadata before reading `SpatialData`. This is added because if SpatialData is written with consolidated metadata and this metadata gets corrupted, the zarr store cannot be opened. Setting `reconsolidate_metadata` to `True` when reading, will rewrite the consolidated metadata store after which is then should be possible to open the zarr store again.
- If a `SpatialData` zarr store is read and detected to be stored in the old format, a warning is given to the user. Users can silence this with this PR only when they rewrite the `SpatialData` object to a new location. In a next PR, we will enable rewriting solely the metadata if possible (if the user did not store using sharding).

### DIFF
--- a/src/spatialdata/_io/io_zarr.py
+++ b/src/spatialdata/_io/io_zarr.py
@@ -7,6 +7,8 @@ from typing import Any, Literal, cast
 
 import zarr.storage
 from anndata import AnnData
+from dask.dataframe import DataFrame as DaskDataFrame
+from geopandas import GeoDataFrame
 from ome_zarr.format import Format
 from pyarrow import ArrowInvalid
 from zarr.errors import ArrayNotFoundError
@@ -22,7 +24,7 @@ from spatialdata._io.io_raster import _read_multiscale
 from spatialdata._io.io_shapes import _read_shapes
 from spatialdata._io.io_table import _read_table
 from spatialdata._logging import logger
-from spatialdata.models import SpatialElement
+from spatialdata._types import Raster_T
 
 
 def _read_zarr_group_spatialdata_element(
@@ -33,7 +35,7 @@ def _read_zarr_group_spatialdata_element(
     read_func: Callable[..., Any],
     group_name: Literal["images", "labels", "shapes", "points", "tables"],
     element_type: Literal["image", "labels", "shapes", "points", "tables"],
-    element_container: dict[str, SpatialElement | AnnData],
+    element_container: dict[str, Raster_T] | dict[str, DaskDataFrame] | dict[str, GeoDataFrame] | dict[str, AnnData],
     on_bad_files: Literal[BadFileHandleMethod.ERROR, BadFileHandleMethod.WARN],
 ) -> None:
     with handle_read_errors(
@@ -43,9 +45,6 @@ def _read_zarr_group_spatialdata_element(
     ):
         if group_name in selector and group_name in root_group:
             group = root_group[group_name]
-            # if isinstance(read_func, TablesReader):
-            #     read_func(root_store_path, group, element_container, on_bad_files=on_bad_files)
-            # else:
             count = 0
             for subgroup_name in group:
                 if Path(subgroup_name).name.startswith("."):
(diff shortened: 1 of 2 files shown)


Zethson:
I ran our sc-imaging and spatial guides with this PR and it all still seems to be green.

src/spatialdata/_io/io_table.py
```diff
@@ -85,7 +85,17 @@ def _read_table(
...
+
+class TablesReader:
+    def __call__(
+        self,
+        path: str,
+        group: zarr.Group,
+        container: dict[str, AnnData],
+        on_bad_files: Literal[BadFileHandleMethod.ERROR, BadFileHandleMethod.WARN],
+    ) -> None:
+        return _read_table(path, group, container, on_bad_files)
```
LucaMarconato: This function returns nothing, because `_read_table()` returns `None`.

I will fix this by extracting, from `_read_table()`, a function that actually reads and returns a single table, as we do for points and shapes.

src/spatialdata/_io/io_zarr.py
```diff
@@ -15,11 +16,85 @@
...
-from spatialdata._io.io_shapes import _read_shapes
-from spatialdata._io.io_table import _read_table
+from spatialdata._io.io_points import PointsReader
+from spatialdata._io.io_raster import MultiscaleReader
+from spatialdata._io.io_shapes import ShapesReader
+from spatialdata._io.io_table import TablesReader
 from spatialdata._logging import logger
+from spatialdata.models import SpatialElement
+
+ReadClasses = MultiscaleReader | PointsReader | ShapesReader | TablesReader
```
LucaMarconato: As you explained me, this change was made to avoid `mypy` compaints, but I still think that we should not add these wrapper classes since they could be just method, with a generic type annotation.

I will try to simplify this and see how it looks.

src/spatialdata/_io/io_zarr.py
```diff
@@ -15,11 +16,85 @@
...
+
+
+def _read_zarr_group_spatialdata_element(
```
LucaMarconato, line 30: Some of these arguments are redundant. I will simplify the function call.

src/spatialdata/_io/io_zarr.py
```diff
@@ -15,11 +16,85 @@
...
+    on_bad_files: Literal[BadFileHandleMethod.ERROR, BadFileHandleMethod.WARN],
+) -> None:
+    with handle_read_errors(
+        on_bad_files,
+        location=group_name,
+        exc_types=JSONDecodeError,
+    ):
+        if group_name in selector and group_name in root_group:
+            group = root_group[group_name]
+            if isinstance(read_func, TablesReader):
```
LucaMarconato: This if/else will be unnecessary after `_read_table()` is modified to read indeed a single table, and not all the tables.

src/spatialdata/_io/io_zarr.py
melonora: main point is that there is a slight difference in arguments for the functions that is causing the problem (different arguments for different functions), no the type hints itself.

src/spatialdata/_io/io_zarr.py
```diff
@@ -71,154 +128,56 @@ def read_zarr(
...
-                        KeyError,
-                        ArrayNotFoundError,
-                    ),
-                ):
-                    shapes[subgroup_name] = _read_shapes(f_elem_store)
-                    count += 1
-            logger.debug(f"Found {count} elements in {group}")
-    if "tables" in selector and "tables" in f:
-        with handle_read_errors(
+    group_readers: dict[
```
LucaMarconato, line 165: The new code is better than before because it reduces redundancy, but it's not very readable. I will simplify it.

src/spatialdata/_io/io_table.py
LucaMarconato: Addressed.

src/spatialdata/_io/io_zarr.py
LucaMarconato: New code pushed, not using classes.

src/spatialdata/_io/io_zarr.py
LucaMarconato: Addressed.

src/spatialdata/_io/io_zarr.py
LucaMarconato: I simplified it a little bit, but I kept the old structure. We can refine later on.

PR #1006, opened 2025-10-27
unpinning dask

melonora:
This PR unpins dask by adding a `.attrs` accessor for both dask Series and DataFrame. It pretty much functions as a drop-in replacement for the `.attrs` attribute in previous Dask versions. This PR is a prerequisite for unlocking zarr v3 sharding api.
This will also unlock spatialdata+rapids-singlecell in the same environment.

Note regarding Dask versions supported: 2025.2.0 onwards dropped all Dask legacy support. Furthermore, after extensive testing, I saw no other way than not to support version 2025.1.0 as we have a couple of operations that cause dependency issues in the Dask graphs. Using 2025.2.0 onwards, this will provide users with an error message that we can resolve, whereas previously, the computation would hang.

For now, some minor computations would be done in memory due to this mixed graph dependency problem. I can reconstruct the proper graph so computation can be completely lazy again, but this would work for a follow-up PR.
Lastly, for writing, `optimize_graph` has been set to `False`. Not doing this causes 'permission denied' errors on Windows, particularly when performing atomic writes, where partial files must be renamed once writing is completed. In general, there were some differences between the various OSs, so I enabled testing for the lowest version of Dask we support and the latest version. I also now include Windows in the CI.

The next follow-up work will be to open a PR to Dask that enables sharding support there. Once that is completed, we can create the sharding support in SpatialData.

### DIFF
--- a/src/spatialdata/_core/spatialdata.py
+++ b/src/spatialdata/_core/spatialdata.py
@@ -13,8 +13,7 @@ import pandas as pd
 import zarr
 from anndata import AnnData
 from dask.dataframe import DataFrame as DaskDataFrame
-from dask.dataframe import read_parquet
-from dask.delayed import Delayed
+from dask.dataframe import Scalar, read_parquet
 from geopandas import GeoDataFrame
 from shapely import MultiPolygon, Polygon
 from xarray import DataArray, DataTree
@@ -1985,9 +1984,7 @@ class SpatialData:
                     else:
                         shape_str = (
                             "("
-                            + ", ".join(
-                                [(str(dim) if not isinstance(dim, Delayed) else "<Delayed>") for dim in v.shape]
-                            )
+                            + ", ".join([(str(dim) if not isinstance(dim, Scalar) else "<Delayed>") for dim in v.shape])
                             + ")"
                         )
                     descr += f"{h(attr + 'level1.1')}{k!r}: {descr_class} with shape: {shape_str} {dim_string}"


src/spatialdata/transformations/operations.py
```diff
@@ -115,6 +115,8 @@ def get_transformation(
     from spatialdata.models._utils import DEFAULT_COORDINATE_SYSTEM
 
     transformations = _get_transformations(element)
+    # if not isinstance(transformations, dict):
+    #     print()
```
melonora:
```suggestion
```

src/spatialdata/transformations/_utils.py
```diff
@@ -55,7 +55,7 @@ def _set_transformations_to_element(element: Any, transformations: MappingToCoor
         attrs[TRANSFORM_KEY] = {}
     attrs[TRANSFORM_KEY] = transformations
     # this calls an eventual setter in the element class; modifying the attrs directly would not trigger the setter
-    element.attrs = attrs
+    # element.attrs = attrs
```
melonora:
```suggestion
```

tests/core/query/test_spatial_query.py
```diff
@@ -645,10 +645,13 @@ def _query(
         geopandas.testing.assert_geodataframe_equal(queried[name], queried_back[name])
 
 
-@pytest.mark.parametrize("with_polygon_query", [True, False])
+@pytest.mark.parametrize("with_polygon_query", [False])
```
melonora:
```suggestion
@pytest.mark.parametrize("with_polygon_query", [True, False])
```

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,165 @@
...
+def wrap_loc_with_attrs() -> None:
+    """Patch dd.DataFrame.loc to preserve _attrs."""
+    original_property = dd.DataFrame.loc  # this is a property
+
+    def loc_with_attrs(self: dd.DataFrame) -> Any:
+        df = self
+        loc = original_property.fget(df)
+
+        class LocWrapper:
+            def __init__(self, parent_loc: Any, parent_df: dd.DataFrame) -> None:
```
ilan-gold: `Any` for `parent_loc`?

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@register_series_accessor("attrs")
+class SeriesAttrsAccessor(_AttrsBase):
+    """Dict-like .attrs accessor for Dask Series."""
+
+    pass
+
+
+def wrap_with_attrs(method: Callable[..., Any]) -> Callable[..., Any]:
```
ilan-gold: I feel like these `Any` are the same no? Maybe use a generic? Same as in `def wrapper`?

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,165 @@
...
+        return result
+
+    return wrapper
+
+
+def wrap_loc_with_attrs() -> None:
+    """Patch dd.DataFrame.loc to preserve _attrs."""
+    original_property = dd.DataFrame.loc  # this is a property
+
+    def loc_with_attrs(self: dd.DataFrame) -> Any:
```
ilan-gold: More `Any`, This returns `LocWrapper`

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,165 @@
...
+        result = method(self, *args, **kwargs)
+        result.attrs = old_attrs
+        return result
+
+    return wrapper
+
+
+def wrap_loc_with_attrs() -> None:
+    """Patch dd.DataFrame.loc to preserve _attrs."""
+    original_property = dd.DataFrame.loc  # this is a property
```
ilan-gold: Why does this live outside `loc_with_attrs`?

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,165 @@
...
+            def __init__(self, parent_loc: Any, parent_df: dd.DataFrame) -> None:
+                self._parent_loc = parent_loc
+                self._parent_df = parent_df
+
+            def __getitem__(self, key: str) -> Any:
+                result = self._parent_loc[key]
+                if hasattr(self._parent_df, "_attrs"):
+                    result.attrs = self._parent_df._attrs.copy()
+                if hasattr(self._parent_df, "attrs"):
+                    result.attrs = self._parent_df.attrs.copy()
```
ilan-gold: What is going on here? `_attrs` and `attrs`? I thought `@register_dataframe_accessor("attrs")` would automatically handle `attrs`?

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,165 @@
...
+
+            def __repr__(self) -> str:
+                return repr(self._parent_loc)
+
+        return LocWrapper(loc, df)
+
+    dd.DataFrame.loc = property(loc_with_attrs)
+
+
+def wrap_iloc_with_attrs() -> None:
```
ilan-gold: This function looks 1-1 duplicated to `wrap_loc_with_attrs` except for `iloc` - please deduplicate

src/spatialdata/_io/accessor.py
melonora: This was because initially I was assigning to private attributes which is not available when having multiple partitions. However, can directly assign to `.attrs` here so I will change it.

src/spatialdata/_io/accessor.py
melonora: deduplicated

src/spatialdata/_core/operations/rasterize.py
```diff
@@ -653,20 +653,28 @@ def rasterize_shapes_points(
...
 
     label_index_to_category = None
     if VALUES_COLUMN in data and data[VALUES_COLUMN].dtype == "category":
         if isinstance(data, DaskDataFrame):
-            data[VALUES_COLUMN] = data[VALUES_COLUMN].cat.as_known()
+            # We have to do this because as_known() does not preserve the order anymore in latest dask versions
+            # TODO discuss whether we can always expect the index from before to be monotonically increasing, because
+            # then we don't have to check order.
+            if index:
+                data[VALUES_COLUMN] = data[VALUES_COLUMN].cat.set_categories(data.index, ordered=True)
```
LucaMarconato, line 675: Do you think that we could report this to `dask`? Maybe it is an unintended change. Or was it more that the order was never guaranteed in the first place?

src/spatialdata/_core/operations/transform.py
```diff
@@ -432,15 +433,20 @@ def _(
...
     # transformed.attrs = {TRANSFORM_KEY: {DEFAULT_COORDINATE_SYSTEM: Identity()}}
     assert isinstance(transformed, DaskDataFrame)
     for ax in axes:
         indices = xtransformed["dim"] == ax
         new_ax = xtransformed[:, indices]
-        transformed[ax] = new_ax.data.flatten()
+        # TODO: discuss with dask team
+        # This is not nice, but otherwise there is a problem with the joint graph of new_ax and transformed, causing
+        # a getattr missing dependency of dependent from_dask_array.
+        new_col = pd.Series(new_ax.data.flatten().compute(), index=transformed.index)
```
LucaMarconato, line 445: I agree that it's a good idea to discuss with Dask team about this.

src/spatialdata/_io/_utils.py
```diff
@@ -337,15 +337,25 @@ def _search_for_backing_files_recursively(subgraph: Any, files: list[str]) -> No
...
+                    else:
+                        for task in v.args[0].value:
+                            if isinstance(task.args[0], dict) and "piece" in task.args[0]:
+                                parquet_file, check0, check1 = task.args[0]["piece"]
+                                if not parquet_file.endswith(".parquet") or check0 is not None or check1 is not None:
+                                    raise ValueError(
+                                        f"Unable to parse the parquet file from the dask subgraph {subgraph}. Please "
+                                        f"report this bug."
+                                    )
+                                files.append(os.path.realpath(parquet_file))
```
LucaMarconato, line 368: Please add some comments to explain the rationale between the new code.

src/spatialdata/_io/_utils.py
```diff
@@ -337,15 +337,25 @@ def _search_for_backing_files_recursively(subgraph: Any, files: list[str]) -> No
                             )
                         parquet_file = t[0]
                         files.append(str(UPath(parquet_file).resolve()))
-                    elif isinstance(v, tuple) and len(v) > 1 and isinstance(v[1], dict) and "piece" in v[1]:
+                    elif "piece" in v.args[0]:
                         # https://github.com/dask/dask/blob/ff2488aec44d641696e0b7aa41ed9e995c710705/dask/dataframe/io/parquet/core.py#L870
-                        parquet_file, check0, check1 = v[1]["piece"]
+                        parquet_file, check0, check1 = v.args[0]["piece"]
                         if not parquet_file.endswith(".parquet") or check0 is not None or check1 is not None:
```
LucaMarconato, line 348: Two comments on `_search_for_backing_files_recursively()`. It is important to test with multiple Dask versions (because over teh versions the structure of the graph changed a bit). I would thefore set a min and max Dask version in the tests.

src/spatialdata/_io/_utils.py
```diff
@@ -337,15 +337,25 @@ def _search_for_backing_files_recursively(subgraph: Any, files: list[str]) -> No
...
                             raise ValueError(
                                 f"Unable to parse the parquet file from the dask subgraph {subgraph}. Please "
                                 f"report this bug."
                             )
```
LucaMarconato, line 352: Second comment on `_search_for_backing_files_recursively()`: it would be amazing if we could drop this function entirely by relying on an API from Dask. Let's chat with the Dask developers to see if they have something to suggest.

src/spatialdata/models/models.py
```diff
@@ -808,9 +808,7 @@ def _(
                 sort=sort,
                 **kwargs,
             )
-            # we cannot compute the divisions whne the index is not monotonically increasing and npartitions > 1
-            if not table.known_divisions and (sort or table.npartitions == 1):
-                table.divisions = table.compute_current_divisions()
+            # TODO: dask does not allow for setting divisions directly anymore. We have to decide on forcing the user.
             if feature_key is not None:
```
LucaMarconato, line 812: Ok we need to improve a bit the code around this.

src/spatialdata/transformations/_utils.py
```diff
@@ -55,7 +55,6 @@ def _set_transformations_to_element(element: Any, transformations: MappingToCoor
         attrs[TRANSFORM_KEY] = {}
     attrs[TRANSFORM_KEY] = transformations
     # this calls an eventual setter in the element class; modifying the attrs directly would not trigger the setter
-    element.attrs = attrs
```
LucaMarconato, line 58: please remove also the comment above.

tests/core/query/test_spatial_query.py
```diff
@@ -648,7 +648,10 @@ def _query(
 @pytest.mark.parametrize("with_polygon_query", [True, False])
 def test_query_points_multiple_partitions(points, with_polygon_query: bool):
     p0 = points["points_0"]
-    p1 = PointsModel.parse(dd.from_pandas(p0.compute(), npartitions=10))
+    attrs = p0.attrs.copy()
+    ddf = dd.from_pandas(p0.compute(), npartitions=10)
+    ddf.attrs = attrs
+    p1 = PointsModel.parse(ddf)
```
LucaMarconato, line 654: Not ideal, but ok.

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,131 @@
+from collections.abc import Callable, Iterator, MutableMapping
+from typing import Any, Literal
+
+import dask.dataframe as dd
+from dask.dataframe.extensions import register_dataframe_accessor, register_series_accessor
+
+
+class _AttrsBase(MutableMapping[str, str | dict[str, str]]):
```
LucaMarconato: We want more flexibility for the values, I'd suggest to do something similar to what `pandas` does: https://github.com/pandas-dev/pandas/blob/46ed6b1c22d78619bb50bc3dee39b9bf718a21d9/pandas/core/generic.py#L321

I propose `MutableMapping[str, Any]`.

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def __init__(self, dask_obj: dd.DataFrame | dd.Series):
+        self._obj = dask_obj
+        if not hasattr(dask_obj, "_attrs"):
+            dask_obj._attrs = {}
+
+    def __getitem__(self, key: str) -> Any:
+        return self._obj._attrs[key]
+
+    def __setitem__(self, key: str, value: str | dict[str, str]) -> None:
```
LucaMarconato: Same here as above.

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Copies _attrs from self before calling method, then assigns to result.
+    Safe for lazy operations like set_index, assign, map_partitions.
+    """
+
+    def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
+        old_accessor = getattr(self, "attrs", {})
+        if hasattr(old_accessor, "_obj") and hasattr(old_accessor._obj, "_attrs"):
+            old_attrs = old_accessor._obj._attrs.copy()
+        elif isinstance(old_accessor, dict):
```
LucaMarconato: please add comments on when these cases are triggered

src/spatialdata/_io/accessor.py
LucaMarconato: I agree (and yes same with the `Any` returned by `wrapper`).

src/spatialdata/__init__.py
```diff
@@ -1,20 +1,12 @@
...
-        "dask.dataframe to disable dask-expr. The support is being worked on, for more information please see"
-        "https://github.com/scverse/spatialdata/pull/570"
-    )
 from importlib.metadata import version
 
+from spatialdata._io.accessor import (
+    DfAttrsAccessor,
+    SeriesAttrsAccessor,
+    wrap_indexer_with_attrs,
+    wrap_with_attrs,
```
LucaMarconato: - `wrap_with_attrs` is not used here, only in `spatialdata._io.accessor`
- `wrap_indexer_with_attrs` is used only here, not in `spatialdata._io.accessor`.

1. I'd move the following
```python
wrap_indexer_with_attrs("loc")
wrap_indexer_with_attrs("iloc")
```
in `spatialdata._io.accessor`.
2. We could even remove the import of `wrap_indexer_with_attrs` and `wrap_with_attrs`, unless we want the users to be able to use them (for instance for methods that are not wrapped already). In such case we should document this somewhere.

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,131 @@
...
+    "compute",
+    "drop",
+    "__getitem__",
+    "copy",
+    "cat",
+    "map_partitions",
+]
+
+for method_name in methods_to_wrap:
+    if hasattr(dd.DataFrame, method_name):
```
LucaMarconato: can we remove this if?

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,131 @@
...
+        original_method = getattr(dd.DataFrame, method_name)
+        setattr(dd.DataFrame, method_name, wrap_with_attrs(original_method))
```
LucaMarconato: One consistency comment: for methods, the `getattr` -> `setattr` logic is applied outside the `wrap_with_attrs` function, while for indexers, the `getattr` -> `setattr` logic is done automatically inside `wrap_indexer_with_attrs`. I would do the following changes:
- (minor): rename `wrap_with_attrs` -> `write_method_with_attrs`
- move the `getattr` -> `setattr` logic inside `write_method_with_attrs`.

LucaMarconato: Basically both functions would return `None`, instead of having the first returning a function`, and the second returning `None`.

LucaMarconato:
I found a bug. In the code below the `elif` was looking strange to me:

```python
def wrap_with_attrs(method: Callable[..., Any]) -> Callable[..., Any]:
    """Wrap a Dask DataFrame method to preserve _attrs.

    Copies _attrs from self before calling method, then assigns to result.
    Safe for lazy operations like set_index, assign, map_partitions.
    """

    def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
        old_accessor = getattr(self, "attrs", {})
        if hasattr(old_accessor, "_obj") and hasattr(old_accessor._obj, "_attrs"):
            old_attrs = old_accessor._obj._attrs.copy()
        elif isinstance(old_accessor, dict):
            old_attrs = old_accessor.copy()
        else:
            old_attrs = {}
        result = method(self, *args, **kwargs)
        result.attrs = old_attrs
        return result

    return wrapper
```

Why would `old_accessor` be a`dict`? The reason is that `result.attrs = old_attrs` calls `__setitem__` of `_AttrsBase` only when the accessor is already present, but `result` doesn't have the accessor available.

Practically, consider the code
```python
_ = (
        df.set_index("a")
        .drop("c", axis=1)
    )
```

When `set_index()` is called we have:
1. `old_attrs = old_accessor._obj._attrs.copy()` -> `old_accessor` is a `dict`.
2. `result` becomes a `Dask` object without the accessor
3. `result.attrs = old_attrs` -> `result.attrs` is now a `dict`

Now when `drop()` is called:
1. `isinstance(old_accessor, dict)` is now true.

I'll write a fix for this.

src/spatialdata/_core/operations/rasterize.py
melonora: I will discuss it during their community meeting. I would have to dive a bit deeper into the exact cause, but they themselves don't seem to define `set_categories` so to me it seems like it comes from pandas dataframe but then the pandas dataframe only works per partition, but I am not certain about that. I did not want to spend too much time on it for now though as they can point me in the right direction much quicker.

src/spatialdata/_io/accessor.py
```diff
@@ -0,0 +1,136 @@
...
+    setattr(dd.DataFrame, indexer_name, property(indexer_with_attrs))
+
+
+methods_to_wrap = [
+    "set_index",
+    "compute",
+    "drop",
+    "__getitem__",
+    "copy",
+    "cat",
```
LucaMarconato: cat is never used to wrap an accessor because hasattr(dd.DataFrame, 'cat') is False.

src/spatialdata/_io/accessor.py
LucaMarconato: Not relevant anymore because now the code is shorter and we do need that signature.

src/spatialdata/_core/operations/transform.py
LucaMarconato: This now works after using the loc accessors, correct?

tests/core/query/test_spatial_query.py
LucaMarconato: This is now addressed.

src/spatialdata/_io/accessor.py
LucaMarconato: This is now addressed.

src/spatialdata/_io/accessor.py
LucaMarconato: This is now addressed.

src/spatialdata/_io/accessor.py
LucaMarconato: I simplified the logic and this can be closed.

src/spatialdata/__init__.py
LucaMarconato: I simplified the logic and this can be closed.

src/spatialdata/_io/accessor.py
melonora: ah yeah, was first from a different implementation which allowed users to add methods themselves in case we forgot the support of any.

LucaMarconato: Yes, we can remove it because `cat` was never wrapped. The questions remain if we should wrap `cat`, I'll open below. This discussion can now be closed.

src/spatialdata/_io/accessor.py
LucaMarconato: I simplified the logic and this can be closed.

tests/core/query/test_spatial_query.py
melonora: not ideal, but I am not certain whether we would want to wrap `from_pandas`. It is not a dataframe method.

src/spatialdata/_io/accessor.py
LucaMarconato: We can proceed as follows:
1. we do not wrap `cat` (everything seems to work)
2. if we see we need to wrap `cat`, let's modify `wrap_method_with_attrs` to access methods from `dd.Series` (now it uses `getattrs` and `setattrs` with `dd.DataFrame`).

src/spatialdata/transformations/_utils.py
melonora: removed.

src/spatialdata/_io/accessor.py
melonora: because otherwise you get a recursion error. You just replace the loc attribute so you get a recursion. The way I implemented it now it stores the original property before patching it.

src/spatialdata/_core/operations/transform.py
melonora: after implementation of the loc/iloc wrapper this issue does not exist anymore so I reverted the change.

src/spatialdata/_io/accessor.py
LucaMarconato: Actually 1 is the way to go: `pandas.Series` do not have `.attrs` in `.cat`.

LucaMarconato: This conversation can be closed. Also, I changed the code, so now also some methods from `dd.Series` are wrapped.

src/spatialdata/_io/_utils.py
melonora: as far as I am aware there is no api for this. Can discuss adding one, but I checked the whole api.

tests/core/query/test_spatial_query.py
LucaMarconato: I now use `.update()`, so I changed the code above.

src/spatialdata/_io/_utils.py
melonora: done and removed old code.

src/spatialdata/_io/_utils.py
melonora: github workflow adjusted.

src/spatialdata/_core/operations/rasterize.py
```diff
@@ -585,6 +585,7 @@ def rasterize_images_labels(
         raise ValueError(f"Unsupported schema {schema}")
 
     # resample the image
+    # TODO results here are vastly different between dask 2025.1.0 and 2025.2.0, from the latter on tests fail.
```
melonora:
```suggestion
```

melonora:
@LucaMarconato think everything is in order here now. Do you see remaining blockers?

melonora:
Action point for new PR is to ensure that the indices for dask dataframes are monotonically increasing across multiple partitions. Right now, it may not be the case, and as a result partitions may overlap, which creates problems when evaluating the Dask graph. However, this will be a separate PR.

Issue #1009, opened 2025-10-30
Categories missing with highly partitioned dask dataframes in PointsModel


jonas2612:
1. Reproduce using the [`blobs` dataset](https://spatialdata.scverse.org/en/stable/api/datasets.html#spatialdata.datasets.blobs)

    ```python
    import numpy as np
    import pandas as pd
    import dask.dataframe as dd
    from spatialdata.datasets import blobs
    
    s = blobs()
    tbl = next(iter(s.tables.values()))
    df = tbl.obs.copy()
    n_cats = 15
    cats = pd.Index([f"G{i:04d}" for i in range(n_cats)], dtype="string")
    
    rng = np.random.default_rng(0)
    n = len(df)
    k_front = min(10_000, n)
    front = rng.choice(cats[:20], size=k_front)
    back  = rng.choice(cats, size=n - k_front)
    df["gene"] = pd.Index(np.concatenate([front, back]), dtype="string")
    
    ddf_many = dd.from_pandas(df, npartitions=217)
    
    c1 = ddf_many["gene"].astype(str).astype("category").head(1).cat.categories
    print("many partitions, categories seen via head(1):", len(c1))  # typically ~20
    
    ddf_as_known = ddf_many["gene"].astype("category").cat.as_known()
    print("with .cat.as_known(), categories:", len(ddf_as_known._meta.cat.categories))
    ```

**Describe the bug**
The setting of the categories in L888 of scr/spatialdata/models/models.py is not taking into account all categories. If the categories are set per partition, not all categories will be properly registered, leading to an inadvertent filtering of the points dataframe and feature loss.

**To Reproduce**
[...]

PR #1010, opened 2025-10-30
fix categories import

jonas2612:
Possible solution to Issue #1009

### DIFF
--- a/src/spatialdata/models/models.py
+++ b/src/spatialdata/models/models.py
@@ -891,7 +891,6 @@ class PointsModel:
             # It also just changes the state of the series, so it is not a big deal.
             if isinstance(data[c].dtype, CategoricalDtype) and not data[c].cat.known:
                 try:
-                    # data[c] = data[c].cat.set_categories(data[c].head(1).cat.categories)
                     data[c] = data[c].cat.set_categories(data[c].compute().cat.categories)
                 except ValueError:
                     logger.info(f"Column `{c}` contains unknown categories. Consider casting it.")


melonora:
Hi @jonas2612, Thanks for the PR. However, before I look further into this, could you please check whether the problem persists when you use the PR that unpins dask, #1006? I had a different issue with partitioning, but it could be connected so want to see if it would be fixed.  

Also, in general, `cat.as_known` in the newer dask versions does not necessarily preserve the order anymore, so it might be better to wait with this PR a tiny bit as we expect to have the unpinning dask PR reviewed soon. It would then be better to branch of from that point given what I just mentioned.

jonas2612:
Hi @melonora. Yes, of course. I'll check it today and get back to you

jonas2612:
That indeed looks very similar to the bug I get.
Still, I checked it with your PR too, but the error still exists:
image
Here, I expect either 500 or 550 genes, but only receive 17.

I understand the difficulty, although do not understand the background well, why the ordering of the categories is important. But if it's better for you, I can branch off and redo the change at a later time point after the PR is reviewed

melonora:
@jonas2612, the reason why it is important is because `as_known` is basically reassigning a column. If the order is changed, a particular category belonging to a certain index can have changed.

LucaMarconato:
Thanks @jonas2612 for the PR, I will merge it now. It introduces a performance regression but we can address it later.

I did some quick tests and it seems that 
```python
data[c] = data[c].cat.as_known()
```

takes the same time as 
```python
data[c] = data[c].cat.set_categories(data[c].compute().cat.categories)
```

Since there may be some concern around the order of categories with `as_known()` (@melonora do you have updates from the Dask team?), let's use `compute()` for now. I will open an issue to remember to explore faster alternatives.

LucaMarconato:
I added a test, which fails before the new change in the pair. As I explain in the test, `compute()` does not preserve the order of categories (same as `as_known()`). We may want to just accept it as the new behavior, or follow up with the dask team.

LucaMarconato:
Thanks for the fix! Actually the performance was impacted quite a lot but I took the opportunity (in a follow up) to add a clear warning message to the user that there should be no needed to actually compute the categories, and inviting to convert that column to categorical before calling the parser: https://github.com/scverse/spatialdata/pull/1061/changes.

Your code is then used as a fallback.

PR #1027, opened 2025-12-05
Lazy load torch

quentinblampey:
If `torch` is installed, importing spatialdata will also import `torch` because of `ImageTilesDataset`.

This PR aims to speed up the import by making the import of torch lazy.

It's a little bit tricky to keep all the dosctrings + auto-completion. For now, it seems the auto-completion is working fine for the instances, but not the `__init__` docstrings.

If you think this PR seems interesting, I can work on fixing the docstrings, else we can just close it


Zethson:
Looks fine but I'll leave this PR up to the maintainers now.

Issue #1030, opened 2025-12-11
The `chunks` parameter is ignored for DataArray models inheriting from `RasterSchema`


MSHelm:
Hey everyone,

I wanted to read in an externally generated segmentation and parse it as a `Label2DModel`. In the harpy package they require identical chunks between the label and the values layer, therefore I wanted to pass in the same chunks into the label model. But it turns out that this parameter gets ignored entirely for a DataArray. I think the culprit is that it never gets passed to `from_array`:

https://github.com/scverse/spatialdata/blob/53b9438a328c5fc2a451d2c8afab439b945ba2b8/src/spatialdata/models/models.py#L186-L189

Repro:
```python
import spatialdata as sd
import numpy as np

arr = np.random.rand(2048, 2048)
model = sd.models.Labels2DModel.parse(
    arr, 
    dims=("y", "x"), 
    transformations={"global": sd.transformations.Identity()},
    chunks={"x": 1024, "y": 1024 } # I also tried passing (1024, 1024) with same effect.
)

print(model.chunks)
# ((2048,), (2048,))
```

Maybe related to https://github.com/scverse/spatialdata/issues/81

PR #1031, opened 2025-12-12
use chunks parameter for dataarray models

MSHelm:
I tried my hand to fix https://github.com/scverse/spatialdata/issues/1030

This should handle all instances where we want to have a `DataArray` after parsing a `RasterSchema` model. For `DataTrees` the chunking is handled in `to_multiscale` 

As I am new to your codebase I hope I adhered to all your guidelines. If not I am happy about any comments what I could improve.

### DIFF
--- a/src/spatialdata/models/models.py
+++ b/src/spatialdata/models/models.py
@@ -185,7 +185,7 @@ class RasterSchema(DataArraySchema):
         # if there are no dims in the data, use the model's dims or provided dims
         elif isinstance(data, np.ndarray | DaskArray):
             if not isinstance(data, DaskArray):  # numpy -> dask
-                data = from_array(data.data)
+                data = from_array(data)
             if dims is None:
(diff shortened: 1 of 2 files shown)


src/spatialdata/models/models.py
```diff
@@ -185,7 +185,7 @@ def parse(
         # if there are no dims in the data, use the model's dims or provided dims
         elif isinstance(data, np.ndarray | DaskArray):
             if not isinstance(data, DaskArray):  # numpy -> dask
-                data = from_array(data)
+                data = from_array(data.data)
```
LucaMarconato: why this change?

LucaMarconato: I'll revert and merge but happy to patch later in case it's needed.

MSHelm: I think I just did that change, because on line 173 it also uses `from_array(data.data)`

PR #1068, opened 2026-02-10
New lazy to_multiscale() for labels

### LINKED ISSUE #1069: `to_multiscale()` is not lazy for labels
I noticed that in `multiscale-spatial-image==2.0.3` the `to_multiscale()` implementation that is used for labels has a `.compute()` that leads to unnecessary computations and high-memory usage: https://github.com/spatial-image/multiscale-spatial-image/blob/ecb6aa410d1e0f3e82daa61ad2e252e921acbe49/multiscale_spatial_image/to_multiscale/_dask_image.py#L196.

Removing the `.compute()` seems to be enough to fix the issue, but this is not a straightforward option because `multiscale-spatial-image` moved the downscaling backend to `ngff-zarr`, adding it as a dependency, and `ngff-zarr` adds some extra dependencies that would make `spatialdata` heavier. At some point we may add `ngff-zarr` as a dependency anyway (because RFC-5 is being implemented both in `ngff-zarr` and `ome-zarr-models-py`, and we will pick one), but we need to think more about this. An initial discussion about dependencies can be found here https://github.com/fideus-labs/ngff-zarr/issues/340. 

Furthermore`multiscale-spatial-image==2.1.0` adds some upper bounds that would have to be relaxed, and does not support the latest version of `ngff-zarr` (on the other hand `2.0.3` doesn't have upper bounds so we can for the moment stay with it). As a last point, `ngff-zarr` seems to have a problem leading to very large dask graphs (affecting performance) for large data. This is being worked on, but it makes sense to wait. https://github.com/fideus-labs/ngff-zarr/issues/48

LucaMarconato:
Closes #1069

This PR (still **work-in-progress**) aims at implementing a function to generate a pyramidal image given a dask array, by relying on `ome-zarr` internal APIs and by extending functionalities currently not available via `ome_zarr.scale.Scaler`. Such a function is also provided by the `multiscale-spatial-image` package, but as explained in the linked issue we hit a performance roadblock. 

`ome-zarr-py` does not support downscaling of the `z` dimension, nor downscaling via scale factors specified per dimension (both cases are supported by `multiscale-spatial-image`); these cases will be covered by this PR. The implementation could also live in `ome-zarr-py` (CC @will-moore @joshmoore @jo-mueller), I'd be happy if someone later can pick up the code so I can remove it from `spatialdata` and use it from upstream.

Side technical note:
- the previous implementation in `spatialdata` uses `xarray_coarsen` (from `multiscale-spatial-image`) for images, (which is fully lazy), and `itk_nearest` (from `multiscale-spatial-image`) for labels (which is not lazy and therefore problematic). Since downscaling for images doesn't give problems, we could use the old implementation, but I'm reimplementing it with `ome-zarr-py` APIs anyway. So far it seems that both implementations lead to identical results when the scale factors are all 2.

### DIFF
--- a/src/spatialdata/_io/io_raster.py
+++ b/src/spatialdata/_io/io_raster.py
@@ -13,7 +13,7 @@ from ome_zarr.writer import write_image as write_image_ngff
 from ome_zarr.writer import write_labels as write_labels_ngff
 from ome_zarr.writer import write_multiscale as write_multiscale_ngff
 from ome_zarr.writer import write_multiscale_labels as write_multiscale_labels_ngff
-from xarray import DataArray, Dataset, DataTree
+from xarray import DataArray, DataTree
 
 from spatialdata._io._utils import (
     _get_transformations_from_ngff_dict,
@@ -27,6 +27,7 @@ from spatialdata._io.format import (
 from spatialdata._utils import get_pyramid_levels
 from spatialdata.models._utils import get_channel_names
 from spatialdata.models.models import ATTRS_KEY
+from spatialdata.models.pyramids_utils import dask_arrays_to_datatree
 from spatialdata.transformations._utils import (
     _get_transformations,
     _get_transformations_xarray,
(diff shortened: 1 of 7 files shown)


LucaMarconato:
The PR is ready. I'll merge and test it with downstream libraries, as it will provide a significant performance boost that I want to make available soon. Additionally, I've created an issue (https://github.com/scverse/spatialdata/issues/1070) to align the code with the z-support from ome-zarr-py once the corresponding PR is merged (https://github.com/ome/ome-zarr-py/pull/516).

PR #1071, opened 2026-02-12
test: make pre-release tests work

### LINKED ISSUE #1039: Better pre-release CI workflow
- [ ] I'll modify the `test.yaml` workflow at a later point in a follow-up PR.

_Originally posted by @LucaMarconato in https://github.com/scverse/spatialdata/issues/1034#issuecomment-3707256822_
            
Move the pre-release job outside PR tests, and instead turn it into a periodic action run on `main`. Otherwise a PR may be blocked due to unrelated code changes (e.g. a pre-release in numpy breaking numba).

flying-sheep:
Fixes #1039

I changed the test setup to use `uv add`, which actually makes sense and results in a single solve per env, as I’ve been preaching since the dawn of time.

Don’t use `[uv] pip install` in CI people. Ever. Why does nobody trust me on that?

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,22 +1,22 @@
 # https://docs.readthedocs.io/en/stable/config-file/v2.html
 version: 2
 build:
-  os: ubuntu-24.04
-  tools:
-    python: '3.13'
-  jobs:
-    post_checkout:
-      # unshallow so version can be derived from tag
-      - git fetch --unshallow || true
-    create_environment:
-      - asdf plugin add uv
-      - asdf install uv latest
-      - asdf global uv latest
-    build:
-      html:
-        - uv sync --group=docs --group=torch
-        - make --directory=docs build
-        - mv docs/_build $READTHEDOCS_OUTPUT
+    os: ubuntu-24.04
+    tools:
+        python: "3.13"
+    jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -32,29 +32,27 @@ jobs:
...
             - name: Install dependencies
               run: |
                   if [[ "${PRERELEASE}" == "allow" ]]; then
-                    uv sync --extra test
-                    : # uv sync --extra test --prerelease ${PRERELEASE}
-                    uv pip install git+https://github.com/scverse/anndata.git
-                    uv pip install --prerelease allow pandas
-                  else
-                    uv sync --extra test
+                    sed -i '' 's/requires-python.*//' pyproject.toml # otherwise uv complains that anndata requires python>=3.12 and we only do >=3.11 😱
```
flying-sheep, line 44: Changing the whole test setup to hatch would remove the need for this hack btw.

Not saying you should do it if the current setup works for you, only that having something manage multiple test conditions for you (locally the same way as in CI) is something I enjoy in developing my projects (even though learning hatch’s config semantics takes time)

melonora: Could also use pixi no? E.g. would solve the same thing (bit biased here as I use pixi more haha))

Issue #456, opened 2024-02-13
Reduce import time


aeisenbarth:
The time to import `spatialdata` is significant, and can impact some usage scenarios.

(This issue is only for tracking. I don't consider it priority at the current stage of the project. Once the API and code structure stabilize and become cleaner, it might be worth to introduce some techniques to tackle this.)

## Example

I am working on a project that takes about 4-5 seconds to import. Partially this is due to other libraries as well, but the biggest chunk is SpatialData.

```bash
$ python -m timeit -n 1 -r 1 -c "import spatialdata"
1 loop, best of 1: 2.28 sec per loop
```

For comparison, numpy takes 81ms.

## Use cases

It may not have a big impact for datascience in a Jupyter notebook where you import once and work all day with the same session.

- **CLI tools:** `my-tool --help` should respond immediately. This can be mitigated by removing my CLI tool's `__main__` top-level imports and use local imports only in those command functions that actually use the imported modules. This way, only when the user issues a command that does processing with SpatialData, it is imported, whereas other commands respond immediately.

- **Napari plugins:** With NPE2, plugins are imported when invoking them in the GUI (for example clicking the menu). When a user invokes a menu item that opens a plugin's widget which uses SpatialData, the user interface remains unresponsive for multiple seconds. For a moment, users think nothing happened.

## Workarounds

As a user of SpatialData, we can:

- Use `if TYPE_CHECKING` if we just import datastructures for type hinting.
- Import submodules instead of top-level imports, e.g. `from spatialdata.transformations import Affine` instead of `import spatialdata`. The major cause is that many libraries want to expose much of their API at a higher-level module (`spatialdata.__init__`), which is also recommended since it allows to make deeper module paths private and change them when needed. I believe that when doing a submodule import, Python does not skip the top-level module, so this may not work.
- Use local imports in functions. I would avoid this where ever possible, because it makes code less organized and readable, and is a source of bugs (some refactoring tools don't catch all local imports and you would only notice at run time, if the bugged function is executed at all).

We can not do anything when:

- When we implement a class that inherits from a spatialdata class, we must import the real parent class.
- When we implement a Pydantic model with a field of type SpatialData (validation cannot work with just a forward reference)

LucaMarconato:
[...]

PR #1075, opened 2026-02-18
Faster `spatialdata` import time

LucaMarconato:
Gives a shot at #456. I'll close it, but in the future we could explore further optimizations (see the `profimp` report below). 

@ivirshup @flying-sheep I'll tag you as reviewers since you did similar work in https://github.com/scverse/scanpy/issues/756.

This PR:
- makes the import of `spatialdata` and submodules lazy
    - also add tests to ensure that the lazy-import machinery remains in sync
- reduces the burden of importing some heavier packages used sporadically, by making those imports local
- improves the `asv` benchmark
- addes `perfimp` to the dev requirements

Results:
```
| Change   | Before [3d9dee3e] <main>   | After [ed9f72f4] <faster-import>   |   Ratio | Benchmark (Parameter)                            |
|----------|----------------------------|------------------------------------|---------|--------------------------------------------------|
| -        | 1.04±0.02s                 | 828±5ms                            |    0.79 | benchmark_imports.timeraw_import_read_zarr       |
| -        | 1.06±0.1s                  | 780±7ms                            |    0.74 | benchmark_imports.timeraw_import_SpatialData     |
| -        | 1.09±0.05s                 | 768±7ms                            |    0.7  | benchmark_imports.timeraw_import_models_elements |
| -        | 1.05±0.01s                 | 583±5ms                            |    0.56 | benchmark_imports.timeraw_import_transformations |
| -        | 1.05±0.05s                 | 20.3±0.1ms                         |    0.02 | benchmark_imports.timeraw_import_spatialdata     |
```

# Output from `profimp`
Command `profimp "from spatialdata import read_zarr" --html > profimp.html && open profimp.html`
image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -183,6 +183,9 @@ select = [
 ]
 unfixable = ["B", "C4", "UP", "BLE", "T20", "RET"]
 
+[tool.ruff.lint.isort]
+required-imports = ["from __future__ import annotations"]
+
 [tool.ruff.lint.pydocstyle]
 convention = "numpy"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,7 @@ docs = [
 benchmark = [
     "asv",
     "memray",
+    "profimp",
 ]
 torch = [
     "torch"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,6 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     "bump2version",
-    "profimp",
 ]
 test = [
     "pytest",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     "bump2version",
-    "sentry-prevent-cli",
+    "profimp",
 ]
 test = [
     "pytest",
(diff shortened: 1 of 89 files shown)


pyproject.toml
```diff
@@ -183,6 +183,9 @@ select = [
 ]
 unfixable = ["B", "C4", "UP", "BLE", "T20", "RET"]
 
+[tool.ruff.lint.isort]
+required-imports = ["from __future__ import annotations"]
```
flying-sheep, line 189: I added this. It makes it so the `UP` lints will always flag when you’re just using an import in an annotation, and therefore will automatically skip such imports at runtime by sticking them into `if TYPE_CHECKING:` blocks.

flying-sheep (approved):
Looks good!

In case you want to do everything we do in scanpy, we also do this to enforce that a list of slow-to-import packages aren’t accidentally imported:

1. define a pytest plugin that isn’t part of the scanpy module (`testing.scanpy._pytest` or a `conftest.py` in the repository root both work, the former has to be activated using `tool.pytest.addopts = ["-p", "testing.scanpy._pytest"]`)
2. import `scanpy` in there and record `sys.modules` before and after
3. check the diff against the list of to-be-lazy-loaded modules.

unfortunately that approach means we can’t use our own warnings in the `tool.pytest.filterwarnings` setting, but I can’t think of another way to do it except for importing it in a subprocess.

PR #999, opened 2025-10-22
Local cloud

sophiamaedler:
I was trying to initialize a spatialdata object directly from S3 as done in the tests here: 

```python
from upath import UPath
import spatialdata as sd
test = UPath( "s3://spatialdata/spatialdata-sandbox/merfish.zarr", endpoint_url="https://s3.embl.de", anon=True )
sd.read_zarr(test)
```

Was failing with:
```python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 4
      2 import spatialdata as sd
      3 test = UPath( "s3://spatialdata/spatialdata-sandbox/merfish.zarr", endpoint_url="https://s3.embl.de/", anon=True )
----> 4 sd.read_zarr(test)

File ~/src/spatialdata/_io/io_zarr.py:282, in read_zarr(store, selection, on_bad_files)
    272     attrs = None
    274 sdata = SpatialData(
    275     images=images,
    276     labels=labels,
   (...)    
    281 )
--> 282 sdata.path = _create_upath(_store)
    283 return sdata

File ~/src/spatialdata/_core/spatialdata.py:590, in SpatialData.path(self, value)
    588     self._path = value
    589 else:
--> 590     raise TypeError("Path must be `None`, a `str` or a `Path` object.")
    592 if not self.is_self_contained():
    593     logger.info(
    594         "The SpatialData object is not self-contained "
    595         "(i.e. it contains some elements that are Dask-backed "
    596         "from locations outside {self.path})."
    597     )

TypeError: Path must be `None`, a `str` or a `Path` object.
```

The implemented changes fix the issues and result in the sdata object being successfully read from S3.

The code now returns:
```python
SpatialData object, with associated Zarr store: s3://spatialdata/spatialdata-sandbox/merfish.zarr
├── Images
│     └── 'rasterized': DataArray[cyx] (1, 522, 575)
├── Points
│     └── 'single_molecule': DataFrame with shape: (<Delayed>, 3) (2D points)
├── Shapes
│     ├── 'anatomical': GeoDataFrame shape: (6, 1) (2D shapes)
│     └── 'cells': GeoDataFrame shape: (2389, 2) (2D shapes)
└── Tables
      └── 'table': AnnData (2389, 268)
with coordinate systems:
    ▸ 'global', with elements:
        rasterized (Images), single_molecule (Points), anatomical (Shapes), cells (Shapes)
with the following Dask-backed elements not being self-contained:
[...]

PR #1087, opened 2026-03-02
 Cloud storage support (Azure, S3, GCS)

### LINKED ISSUE #441: Support for private remote object storage
Right now, there is support for local .zarr stores and remote stores publically accessible via HTTP or S3.
Private remote stores are more difficult, as they need certain options or credentials that are not representable by simply a string or Path. One option is to use a zarr.storage.FSStore, which can have storage_options or any fsspec.spec.AbstractFileSystem.

Two pull requests enable this:
- Support init of ome_zarr_py.io.ZarrLocation with zarr.storage.FSStore ([#349](https://github.com/ome/ome-zarr-py/pull/349))
- Support remote private storage by consistent use of substore

SamirMoustafa:
# Cloud storage support (Azure, S3, GCS)

## Summary

Add read/write support for SpatialData on remote object storage via `UPath`, fixing the issue reported in [#999](https://github.com/scverse/spatialdata/pull/999) where `sd.read_zarr(UPath("s3://..."))` failed because `SpatialData.path` did not accept `UPath`. [PR #971](https://github.com/scverse/spatialdata/pull/971) ("add remote support") pursued the same goal but remains a draft, blocked on zarr v3/ome-zarr and async fsspec after dask unpinning. This PR delivers working remote support by fixing the path setter, wrapping fsspec in an async filesystem where required for current zarr, and testing Azure, S3, and GCS via Docker emulators. It also addresses [#441](https://github.com/scverse/spatialdata/issues/441) (private remote object storage): credentials go via `UPath` kwargs or via a pre-opened `zarr.Group` (e.g. `read_zarr(zarr.open("s3://...", storage_options={...}))`). Fixes #441.

## Supported features

- **Path handling:** `SpatialData.path` accepts `None`, `str`, `Path`, or `UPath` (enables remote-backed objects).
- **Read:** `SpatialData.read(upath)` and `read_zarr(upath)` for Azure Blob (`az://`), S3 (`s3://`), and GCS (`gs://`) using [universal-pathlib](https://github.com/fsspec/universal_pathlib) (`UPath`). For private stores, `read_zarr(zarr_group)` is also supported when the store is opened with `zarr.open(..., storage_options=...)`.
- **Write:** `sdata.write(upath)` and element-level writes to the same backends; parquet (points/shapes) and zarr (raster/tables) written via fsspec with async filesystem support where required.
- **Consolidated metadata:** Read/write of consolidated metadata on remote stores (e.g. `zmetadata`) supported.

## Testing

Remote storage is tested with **Docker-based emulators** (Azurite for Azure, moto for S3, fake-gcs-server for GCS). In CI we build `tests/io/remote_storage/Dockerfile.emulators`, start the emulators on Ubuntu, then run the full test suite including `tests/io/remote_storage/`. **These remote-storage tests run only on Ubuntu** (Linux), because they depend on Docker; on Windows and macOS we skip `tests/io/remote_storage/` and run the rest of the suite. To run the remote tests locally you need Docker and can start the emulators with the same image and ports (5000, 10000, 4443) as in the workflow.

## Example (three providers)

```python
from upath import UPath

# Azure Blob Storage
az_path = UPath("az://my-container/data.zarr", connection_string="<your-connection-string>")
sdata = SpatialData.read(az_path)

# Amazon S3 (e.g. public bucket or custom endpoint)
s3_path = UPath(
    "s3://bucket/data.zarr",
    endpoint_url="https://s3.embl.de",  # omit for default AWS
    anon=True,
)
sdata = SpatialData.read(s3_path)

# Google Cloud Storage
gs_path = UPath("gs://my-bucket/data.zarr", token="anon", project="my-project")
sdata = SpatialData.read(gs_path)

# Write works the same way (any provider)
# sdata.write(az_path)
```

Credentials and options are passed through `UPath` (e.g. `connection_string`, `endpoint_url`, `anon`, `token`, `project`) as supported by the underlying fsspec backend.

---

# Release notes

- Add cloud storage support: read and write SpatialData from/to Azure Blob, S3, and GCS using `UPath`. `SpatialData.path` now accepts `UPath` in addition to `str` and `Path`. Fixes initialization from remote stores (e.g. S3) as in [#999](https://github.com/scverse/spatialdata/pull/999). Fixes [#441](https://github.com/scverse/spatialdata/issues/441) (private remote object storage).


selmanozleyen:
- Maybe add the docker tests as a different jobs.
- Add pytest marks for tests that expect docker setup

tests/conftest.py
```diff
@@ -89,12 +93,18 @@ def tables() -> list[AnnData]:
...
+    # Use two regions so the table categorical has two categories; otherwise anndata does not
+    # write the obs/region/codes/c/0 chunk (only codes/zarr.json), causing 404 on remote read.
     return SpatialData(
         images=_get_images(),
         labels=_get_labels(),
         shapes=_get_shapes(),
         points=_get_points(),
-        tables=_get_tables(region="labels2d", region_key="region", instance_key="instance_id"),
+        tables=_get_tables(
+            region=["labels2d", "poly"],
```
selmanozleyen: We shouldn't change the full_sdata for only the version we want to test. See test_relational_query.py file for example. This doesn't need to change in this PR but it changed to accommadate these changes here

tests/core/operations/test_spatialdata_operations.py
```diff
@@ -559,14 +559,15 @@ def test_init_from_elements(full_sdata: SpatialData) -> None:
 
 
 def test_subset(full_sdata: SpatialData) -> None:
-    element_names = ["image2d", "points_0", "circles", "poly"]
+    # Exclude labels and poly so the default table (annotating labels2d and poly) is not included
+    element_names = ["image2d", "points_0", "circles"]
```
selmanozleyen: Same point as test_relational_query.py, this file shouldn't be touched for this PR.

src/spatialdata/_io/io_zarr.py
```diff
@@ -231,12 +233,26 @@ def read_zarr(
...
+    elif isinstance(store, str):
+        sdata.path = UPath(store) if "://" in store else Path(store)
+    elif isinstance(store, Path):
+        sdata.path = store
+    elif isinstance(store, zarr.Group):
+        if isinstance(resolved_store, LocalStore):
+            sdata.path = Path(resolved_store.root)
+        elif isinstance(resolved_store, FsspecStore):
+            sdata.path = UPath(str(_FsspecStoreRoot(resolved_store)))
+        else:
```
selmanozleyen, line 247: Reading from a remote zarr.Group still loses the original storage options when sdata.path is reconstructed. In the zarr.Group branch you rebuild the path from str(_FsspecStoreRoot(...)), which throws away the original fs config. That means a later write() / write_element() on the returned SpatialData can no longer round-trip to Azure/GCS unless the credentials happen to come from global env.

selmanozleyen:
Here I also added suggestions for https://github.com/SamirMoustafa/spatialdata/pull/1 so that we can explicitly use a zarr store. Get's rid of the `_FsspecStoreRoot` tricks

src/spatialdata/_io/io_shapes.py
```diff
@@ -150,6 +164,60 @@ def _write_shapes_v01(shapes: GeoDataFrame, group: zarr.Group, element_format: F
...
+        import fsspec
+
+        fs_dict = json.loads(fs.to_json())
+        fs_dict["asynchronous"] = False
+        sync_fs = fsspec.AbstractFileSystem.from_json(json.dumps(fs_dict))
+        sync_fs.put_file(tmp_path, path._path)
+    else:
+        fs.put(tmp_path, str(path))
+
+
```
selmanozleyen, line 220: We shouldn't need to have these parquet sidecars for authentication of each cloud system. If we can have a full zarr-native spatialdata there is no need to re-extract Azure/S3/GCS credentials for parquet APIs and
less distinction between "Zarr backing" and "non-Zarr sidecar next to the Zarr group".

PR #1089, opened 2026-03-09
Deconcatenate SpatialData to a list of SpatialData objects

gboscagli:
Following issue scverse/spatialdata-io#318, I created a `deconcatenate()` function, alongside a comprehensive test.

This function takes as input a `SpatialData` object with >1 regions and returns one (or more) `SpatialData` objects populated only with data from a specific region and transformed to a given coordinate system. Briefly, this function is a wrapper of `match_sdata_to_table()`.

We can also think of an optional parameter controlling the conversion to `AnnData`. However, `to_legacy_anndata()` lives inside spatialdata-io, so @LucaMarconato and other users suggested to add the centroids directly in the `AnnData` table of `SpatialData`.

Thoughts?


LucaMarconato:
Hey, thanks for starting the PR. I moved the work in #1124 so I have push rights. Let's follow up there.

PR #1092, opened 2026-03-12
ome zarr chunks

### LINKED ISSUE #1090: `SpatialData.write()` fails for `spatialdata.datasets.blobs()` with `ome-zarr==0.14.0` (`TypeError: Expected an iterable of integers`)
## Summary

After rebuilding a fresh environment `SpatialData.write()` started failing.

The same code works in an older environment with the same `spatialdata`, `dask`, and `zarr` versions, but with `ome-zarr==0.13.0` instead of `0.14.0`.

This looks like a regression at the `spatialdata` / `ome-zarr` boundary when `chunks` are passed from xarray/dask to the writer.

## Minimal reproducer

```python
import os
import tempfile

sdata = blobs()
out = os.path.join(tempfile.gettempdir(), "sd_blobs_write_issue.zarr")
sdata.write(out)
```

## Observed behavior

The call to `sdata.write(out)` raises:

```python
TypeError: Expected an iterable of integers. Got ((3,), (512,), (512,)) instead.
```

## Full traceback

```python
  File "<stdin>", line 9, in <module>
  File ".../site-packages/spatialdata/_utils.py", line 263, in wrapper
    return f(*args, **kwargs)
  File ".../site-packages/spatialdata/_core/spatialdata.py", line 1176, in write
    self._write_element(
  File ".../site-packages/spatialdata/_core/spatialdata.py", line 1229, in _write_element
    write_image(
  File ".../site-packages/spatialdata/_io/io_raster.py", line 371, in write_image
    _write_raster(
  File ".../site-packages/spatialdata/_io/io_raster.py", line 197, in _write_raster
    _write_raster_dataarray(
  File ".../site-packages/spatialdata/_io/io_raster.py", line 275, in _write_raster_dataarray
    write_single_scale_ngff(
  File ".../site-packages/ome_zarr/writer.py", line 651, in write_image
    dask_delayed_jobs = _write_pyramid_to_zarr(
  File ".../site-packages/ome_zarr/writer.py", line 754, in _write_pyramid_to_zarr
    da.to_zarr(
  File ".../site-packages/dask/array/core.py", line 4101, in to_zarr
    z = root.create_array(name=array_name, **zarr_array_kwargs)
  File ".../site-packages/zarr/core/group.py", line 2715, in create_array
    self._sync(
  File ".../site-packages/zarr/core/array.py", line 777, in _create_metadata_v3
    chunk_grid_parsed = RegularChunkGrid(chunk_shape=chunk_shape)
  File ".../site-packages/zarr/core/chunk_grids.py", line 180, in __init__
    chunk_shape_parsed = parse_shapelike(chunk_shape)
  File ".../site-packages/zarr/core/common.py", line 200, in parse_shapelike
    raise TypeError(msg)
TypeError: Expected an iterable of integers. Got ((3,), (512,), (512,)) instead.
```

## Version comparison

Failing environment:

```text
Python 3.12.12
spatialdata 0.7.2
ome-zarr 0.14.0
zarr 3.1.5
dask 2026.1.1
```

Working environment:

```text
Python 3.12.12
spatialdata 0.7.2
ome-zarr 0.13.0
zarr 3.1.5
dask 2026.1.1
```

So the only relevant version difference I found between the two environments is `ome-zarr`.

## Suspected cause

It looks like `spatialdata` passes xarray/dask chunk metadata in the form returned by `.chunks`, i.e. a tuple of tuples:

```python
((3,), (512,), (512,))
```

In `spatialdata/_io/io_raster.py`, `raster_data.chunks` is forwarded into `storage_options["chunks"]`:

```python
chunks = raster_data.chunks
...
storage_options = {"chunks": chunks}
```

Then in `ome_zarr/writer.py`, that value is forwarded into `zarr_array_kwargs["chunks"]`:

```python
chunks_opt = options.pop("chunks", None)
...
zarr_array_kwargs["chunks"] = chunks_opt
```

But `zarr` expects a flat chunk shape like:

```python
(3, 512, 512)
```

not the dask chunk grid:

```python
((3,), (512,), (512,))
```

## Expected behavior

`blobs().write(path)` should succeed, or the chunk metadata should be normalized before being passed to Zarr.

## Workaround

Pinning `ome-zarr==0.13.0` avoids the problem in my setup.

### LINKED ISSUE #1091: Single-scale image writes are serialized as multiscale pyramids with ome-zarr==0.14.0
Writing a single-scale xarray.DataArray image with SpatialData.write() produces a multiscale OME-Zarr pyramid on disk when using ome-zarr 0.14.0, even though the input is not multiscale.

### Expected behavior
A single-scale image should remain single-scale on disk unless pyramid generation was explicitly requested.

Roundtrip expectation:

input: xarray.DataArray
output after read_zarr(): xarray.DataArray
Actual behavior
The image is written as a pyramid (s0, s1, s2, ...), and read_zarr() returns a multiscale object on roundtrip.

Minimal reproducible example
```
import tempfile
from pathlib import Path
from numpy.random import default_rng

import dask.array as da

from spatialdata import SpatialData, read_zarr
from spatialdata.models import Image2DModel

RNG = default_rng(0)

data = da.from_array(
    RNG.random((3, 800, 1000)),
    chunks=((3,), (300, 200, 300), (512, 488)),
)

image = Image2DModel.parse(data, dims=("c", "y", "x"))
print(type(image))  # xarray.DataArray

sdata = SpatialData(images={"image": image})

tmpdir = Path(tempfile.mkdtemp())
path = tmpdir / "data.zarr"

sdata.write(path)
sdata2 = read_zarr(path)

print(type(sdata2["image"]))
```
On-disk result
The written image group contains multiple scales:

data.zarr/images/image/
  s0/
  s1/
  s2/
  s3/
  s4/

### Suspected cause
spatialdata writes single-scale images through ome_zarr.writer.write_image().

In current ome_zarr, write_image() defaults to building a pyramid via:

scale_factors=(2, 4, 8, 16)
So calling it without overriding scale_factors writes a multiscale pyramid even for a single-scale input.

### LINKED ISSUE #1093: `PointsModel.parse()` fails when `feature_key` and `instance_key` are omitted
`PointsModel.parse` fails when called without `feature_key` and `instance_key`. Both arguments are marked as optional in the current docstring/docs, but in practice at least one must be provided.

## Reproduce
This code reproduces the issue when parsing transcripts from a Xenium dataset:

```python
import dask.dataframe as dd
from spatialdata.models import PointsModel
from spatialdata.transformations import Identity

ddf = dd.read_parquet("path_to_xenium_output/transcripts.parquet")

points_parse_kwargs = {
    'coordinates': {'x': 'x_location', 'y': 'y_location', 'z': 'z_location'},
    'transformations': {'global': Identity()}
}

points = PointsModel.parse(ddf, **points_parse_kwargs)
```

Fails with:

```
AttributeError: 'DataFrame' object has no attribute 'attrs'
```

## Cause

The `_add_metadata_and_validate` method only creates `data.attrs` if `feature_key` or `instance_key` is [provided](https://github.com/scverse/spatialdata/blob/main/src/spatialdata/models/models.py#L886):

```python
if feature_key is not None or instance_key is not None:
    data.attrs[ATTRS_KEY] = {}
```

Later `_parse_transformations` assumes `attrs` [exists](https://github.com/scverse/spatialdata/blob/main/src/spatialdata/models/models.py#L60):

```python
@get_transformations.register(DaskDataFrame)
def _(e: DaskDataFrame) -> ...:
    return _get_transformations_from_dict_cont(e.attrs)
```

If neither key is passed, `_parse_transformations` fails.

ArneDefauw:
Fix for https://github.com/scverse/spatialdata/issues/1090. 

Note that unit tests still fail for ome-zarr ==0.14.0, due to https://github.com/scverse/spatialdata/issues/1091

### DIFF
--- a/src/spatialdata/_io/io_raster.py
+++ b/src/spatialdata/_io/io_raster.py
@@ -90,34 +90,27 @@ def _normalize_explicit_chunks(chunks: object) -> tuple[int, ...] | int:
     return normalized
 
 
-def _prepare_single_scale_storage_options(
+def _prepare_storage_options(
     storage_options: JSONDict | list[JSONDict] | None,
-) -> JSONDict | list[JSONDict] | None:
+    data: list[da.Array],
+) -> list[JSONDict]:
     if storage_options is None:
-        return None
+        return [{"chunks": _normalize_explicit_chunks(arr.chunks)} for arr in data]
     if isinstance(storage_options, dict):
+        if "chunks" not in storage_options:
+            return [{**storage_options, "chunks": _normalize_explicit_chunks(arr.chunks)} for arr in data]
         prepared = dict(storage_options)
-        if "chunks" in prepared:
-            prepared["chunks"] = _normalize_explicit_chunks(prepared["chunks"])
-        return prepared
-    return [dict(options) for options in storage_options]
+        prepared["chunks"] = _normalize_explicit_chunks(prepared["chunks"])
+        return prepared  # type: ignore[return-value]
 
-
-def _prepare_multiscale_storage_options(
-    storage_options: JSONDict | list[JSONDict] | None,
-) -> JSONDict | list[JSONDict] | None:
-    if storage_options is None:
-        return None
-    if isinstance(storage_options, dict):
-        prepared = dict(storage_options)
-        if "chunks" in prepared:
-            prepared["chunks"] = _normalize_explicit_chunks(prepared["chunks"])
-        return prepared
-
-    prepared_options = [dict(options) for options in storage_options]
-    for options in prepared_options:
-        if "chunks" in options:
-            options["chunks"] = _normalize_explicit_chunks(options["chunks"])
+    prepared_options = []
+    for i, options in enumerate(storage_options):
+        opts = dict(options)
+        if "chunks" not in opts:
+            opts["chunks"] = _normalize_explicit_chunks(data[i].chunks)
+        else:
+            opts["chunks"] = _normalize_explicit_chunks(opts["chunks"])
+        prepared_options.append(opts)
     return prepared_options
 
 
@@ -335,7 +328,7 @@ def _write_raster_dataarray(
         raise ValueError(f"{element_name} does not have any transformations and can therefore not be written.")
     input_axes: tuple[str, ...] = tuple(raster_data.dims)
     parsed_axes = _get_valid_axes(axes=list(input_axes), fmt=raster_format)
-    storage_options = _prepare_single_scale_storage_options(storage_options)
+    storage_options = _prepare_storage_options(storage_options, [data])
     # Explicitly disable pyramid generation for single-scale rasters. Recent ome-zarr versions default
     # write_image()/write_labels() to scale_factors=(2, 4, 8, 16), which would otherwise write s0, s1, ...
     # even when the input is a plain DataArray.
@@ -405,7 +398,7 @@ def _write_raster_datatree(
         raise ValueError(f"{element_name} does not have any transformations and can therefore not be written.")
 
     parsed_axes = _get_valid_axes(axes=list(input_axes), fmt=raster_format)
-    storage_options = _prepare_multiscale_storage_options(storage_options)
+    storage_options = _prepare_storage_options(storage_options, data)
     ome_zarr_format = get_ome_zarr_format(raster_format)
     dask_delayed = write_multi_scale_ngff(
         pyramid=data,
--- a/src/spatialdata/_io/io_raster.py
+++ b/src/spatialdata/_io/io_raster.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, Literal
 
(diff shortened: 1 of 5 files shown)


src/spatialdata/_io/io_raster.py
```diff
@@ -38,6 +39,88 @@
...
+    storage_options: JSONDict | list[JSONDict] | None,
+) -> JSONDict | list[JSONDict] | None:
+    if storage_options is None:
+        return None
+    if isinstance(storage_options, dict):
+        prepared = dict(storage_options)
+        if "chunks" in prepared:
+            prepared["chunks"] = _normalize_explicit_chunks(prepared["chunks"])
+        return prepared
+    return [dict(options) for options in storage_options]
```
LucaMarconato: This function behaves like `_prepare_multiscale_storage_options()`, without normalizing the list of storage options case. Can we remove it and just use `_prepare_multiscale_storage_options()` (after renaming it to `_prepare_storage_options()`?

src/spatialdata/_io/io_raster.py
```diff
@@ -38,6 +39,88 @@
 )
 
 
+def _is_flat_int_sequence(value: object) -> TypeGuard[Sequence[int]]:
+    if isinstance(value, str | bytes):
+        return False
```
LucaMarconato, line 45: Why do we need this? Please either remove or document.

src/spatialdata/_io/io_raster.py
```diff
@@ -38,6 +39,88 @@
...
+    if not isinstance(value, Sequence):
+        return False
+    return all(isinstance(v, int) for v in value)
+
+
+def _is_dask_chunk_grid(value: object) -> TypeGuard[Sequence[Sequence[int]]]:
+    if isinstance(value, str | bytes):
+        return False
```
LucaMarconato, line 53: Same for this check.

src/spatialdata/_io/io_raster.py
```diff
@@ -38,6 +39,88 @@
...
+def _is_regular_dask_chunk_grid(chunk_grid: Sequence[Sequence[int]]) -> bool:
+    # Match Dask's private _check_regular_chunks() logic without depending on its internal API.
+    for axis_chunks in chunk_grid:
+        if len(axis_chunks) <= 1:
+            continue
+        if len(set(axis_chunks[:-1])) > 1:
+            return False
+        if axis_chunks[-1] > axis_chunks[0]:
+            return False
+    return True
```
LucaMarconato, line 116: I'd add a docstring with examples (or examples in-line with the code) to show what fails and what not.

I would add the following:
triggers the continue in the first if:
- [(4,)]
- [()]

triggers the first return False
- [(4, 4, 3, 4)]

triggers the second return False
- [(4, 4, 4, 5)]

exits with the last return True
- [(4, 4, 4, 4)], succeeds, all chunks equal
- [(4, 4, 4, 1)], succeeds, final chunk is < of the initial one

LucaMarconato: Also: I would add all the examples above in a test, for the function `_is_regular_dask_chunk_grid()`.

tests/io/test_readwrite.py
```diff
@@ -623,6 +626,49 @@ def test_bug_rechunking_after_queried_raster():
         queried.write(f)
 
 
+def test_write_irregular_dask_chunks_without_explicit_storage_options(tmp_path: Path) -> None:
+    data = da.from_array(RNG.random((3, 800, 1000)), chunks=((3,), (300, 200, 300), (512, 488)))
+    image = Image2DModel.parse(data, dims=("c", "y", "x"))
+    sdata = SpatialData(images={"image": image})
+
+    sdata.write(tmp_path / "data.zarr")
```
LucaMarconato: I would find it more natural if this test was failing. Now that `chunks = raster_data.chunks` has been removed it, writing doesn't fail, but it ignores the chunks in the data. I think a natural behavior is that if storage option specifies `chunks`, these are used, otherwise the ones from the data (and if the `chunks from the data are irregular and no storage options are specified, an error would be raised).

LucaMarconato (changes requested):
Thanks for the PR! I would make some changes as described.

LucaMarconato:
- [the commit `930922c`](https://github.com/scverse/spatialdata/pull/1092/commits/930922c2ad96c3cf2d399b154b5d871c7248e4af) closes https://github.com/scverse/spatialdata/issues/1093, I included it here since I was encountering that when running some debug scripts.

tests/io/test_readwrite.py
LucaMarconato: I now implemented this in https://github.com/scverse/spatialdata/pull/1092/changes/20696b558589b42daf6fe227ebe2250ad77aab0d changing the test so that it expects to fail.

src/spatialdata/_io/io_raster.py
LucaMarconato: I unified the two functions in https://github.com/scverse/spatialdata/pull/1092/changes/20696b558589b42daf6fe227ebe2250ad77aab0d. Happy to hear what you think (in case we need two, we can revert, but I think we can proceed with one function).

src/spatialdata/_io/io_raster.py
LucaMarconato: Removed here: https://github.com/scverse/spatialdata/pull/1092/changes/2450bd437914004cc940279482f2dc60293e0575

src/spatialdata/_io/io_raster.py
LucaMarconato: Removed here: https://github.com/scverse/spatialdata/pull/1092/changes/2450bd437914004cc940279482f2dc60293e0575

src/spatialdata/_io/io_raster.py
LucaMarconato: Added docstring and tests here: https://github.com/scverse/spatialdata/pull/1092/changes/2450bd437914004cc940279482f2dc60293e0575

LucaMarconato:
I implemented the changes mentioned in the code review. Please let me know if you agree with the changes. If yes I'll merge and work on a release.

LucaMarconato (approved):

PR #1097, opened 2026-03-23
Add/pixi tasks and Claude Skills for code profiling

LucaMarconato:
Adding dev tooling via pixi tasks.


LucaMarconato:
Adding also Claude skills configuration. I don't expect skills to work out-of-the-box and in my experience it's easy to have them go out of sync with the rest of the codebase. For instance in a previous versions of the code `speedscope` was installed separately, while now it is being run with `npx`, while this change is not reflected in the `py-spy` skill.

Rather than overoptimizing, let's start from here and whoever is running the skills is encouraged to commit improvements. Worst case the skill will not be used, but interesting commands can be found by the agent (or human) within its recipe.

PR #1100, opened 2026-03-25
Fix reset_index crash when obs index name matches existing column

timtreis:
## Summary
- `_inner_join_spatialelement_table` and `_left_join_spatialelement_table` call `table.obs.reset_index()` which raises `ValueError: cannot insert <name>, already exists` when the obs index name matches an existing column (e.g. `EntityID` in Merfish data)
- Adds a small helper `_reset_index_preserving_existing_columns` that drops the index when its name already exists as a column, avoiding the collision

### DIFF
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -34,6 +34,17 @@ from spatialdata.models import (
 )
 
 
+def _reset_index_preserving_existing_columns(df: pd.DataFrame) -> pd.DataFrame:
+    """Reset the index of a DataFrame, dropping the index if its name already exists as a column.
+
+    Avoids ``ValueError: cannot insert <name>, already exists`` when the index
+    name collides with an existing column (e.g. "EntityID" in Merfish data).
+    """
+    if df.index.name is not None and df.index.name in df.columns:
+        return df.reset_index(drop=True)
+    return df.reset_index()
+
+
 def get_element_annotators(sdata: SpatialData, element_name: str) -> set[str]:
     """
     Retrieve names of tables that annotate a SpatialElement in a SpatialData object.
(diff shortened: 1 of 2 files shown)


src/spatialdata/_core/query/relational_query.py
```diff
@@ -34,6 +34,17 @@
...
 
 
+def _reset_index_preserving_existing_columns(df: pd.DataFrame) -> pd.DataFrame:
+    """Reset the index of a DataFrame, dropping the index if its name already exists as a column.
+
+    Avoids ``ValueError: cannot insert <name>, already exists`` when the index
+    name collides with an existing column (e.g. "EntityID" in Merfish data).
+    """
+    if df.index.name is not None and df.index.name in df.columns:
+        return df.reset_index(drop=True)
```
LucaMarconato, line 44: So here we are dropping the index instead of a column. This could be a problem if the column is different from the index. 

What's the origin behind this problem? Is it that for the merfish data we should have had `sdata['table'].index.name = None`? That would be a better fix, while still raising an exception here.

tests/core/query/test_relational_query.py
```diff
@@ -1262,3 +1264,28 @@ def test_filter_by_table_query_complex_combination(complex_sdata):
         assert ("circles", idx) in table_instance_ids
     for idx in result["poly"].index:
         assert ("poly", idx) in table_instance_ids
+
+
+@pytest.mark.parametrize("how", ["inner", "left"])
+def test_join_spatialelement_table_obs_index_name_collision(how):
+    """join_spatialelement_table must not crash when obs index name matches an existing column.
```
LucaMarconato, line 1271: I think it should crash in general to avoid a silent branching in the behavior. The ok case would be if both the index and the column have the same values.

LucaMarconato:
Since the code path triggering this has been addressed in `spatialdata-plot`, and I think that raising an exception is good here, I will close this PR. 

If someone wants to iterate on this, feel free to reopen (please in doing so address the review comments I added).

PR #1104, opened 2026-04-01
Speedup for bounding_box_query

MeyerBender:
Speedup for `bounding_box_query()`, especially in the case of many points in the object. Main improvements:
- removed unnecessary `compute()` call
- rewrote transformation logic: in the case of identity or scaling transformations, we can avoid the two-step procedure of doing a coarse query in intrinsic space and then a fine-grained on in the global coordinate space

With this, we achieved speedups between 30% and 70% (depending on the data). In the future, one could consider using a spatial index for the point to speed things up further.

@LucaMarconato it would be great if you could add a test for the case we discussed at the hackathon, just to ensure our rewrite of the transformation logic didn't break anything.

CC @dschaub95


src/spatialdata/_core/query/spatial_query.py
```diff
@@ -386,6 +391,7 @@ def _bounding_box_mask_points(
     axes: tuple[str, ...],
     min_coordinate: list[Number] | ArrayLike,
     max_coordinate: list[Number] | ArrayLike,
+    points_df: pd.DataFrame | None = None,
```
LucaMarconato: `_bounding_box_mask_points()` is called only in two locations, and in both `points_df` is available. I will therefore remove the `points` argument, and only keep `points_df`.

src/spatialdata/_core/query/spatial_query.py
```diff
@@ -404,31 +410,35 @@ def _bounding_box_mask_points(
...
     {max_coordinate_docs}
+    points_df
+        A pre-computed pandas dataframe. Useful if the points_df has already been materialized, otherwise the
+        methods simply calls .compute() on the dask data frame
 
     Returns
     -------
     The masks for the points inside the bounding boxes.
     """
     element_axes = get_axes_names(points)
```
LucaMarconato: This function uses single dispatch and expects a Dask `DataFrame`. I will extend this to support also `pd.DataFrame`. In the future we will allow points to be `pd.DataFrame` also inside the `SpatialData` object.

LucaMarconato:
Thanks! This is the test case I wanted to add (it's a case in which the bounding box of the backward-transformed bounding box contains extra points with respect to the backward-transformed queried points).
image

Here the queried point is C, and the bounding box of the backward-transformed bounding box would contain A, B, C.

The test passes!

PR #1106, opened 2026-04-14
Support sharding through config and raster_write_kwargs

melonora:
This PR adds the following:
- passing kwargs for zarr.create_array directly as `raster_write_kwargs` for io functions like `.write` and `.write_element`. This also adds the ability to write sharded arrays. Support for anndata sharding is to be added in a follow up PR.
- proper docstrings for the new `raster_write_kwargs` argument.
- Extension of the current config to include `raster_chunks` and `raster_shards`. The config can now be stored in a default location or a custom location. Additionally, environment variables can be set to temporarily override the values.  
- Adding zarrs as a dependency and enabled the codec by default to allow for faster io when writing shards. This is a discussion point of whether we should do this or provide more of an opt-in for advanced users.
 
Additional changes
-  Minimal supported version of dask is 2026.3.0. The reason here is that only this provides the api in such a way that you don't risk zarr format 2 being written in a zarr v3 group and vice versa + it includes the setting that prevents collaps of partitions of dask dataframes after reading from parquet.

@LucaMarconato


melonora:
Failing atm due to ome-zarr not yet being released. You can test locally with ome-zarr-py from main.

Also, need to add support for zarrs to improve speed of shard io

Mr-Milk:
Should we also allow the control of sharding for anndata?

melonora:
Yes, but not as part of this PR. I will adjust the config though to accommodate.

LucaMarconato:
> Adding zarrs as a dependency and enabled the codec by default to allow for faster io when writing shards. This is a discussion point of whether we should do this or provide more of an opt-in for advanced users.

I'll loop in @ilan-gold and @flying-sheep. What's the plan for `anndata` on this?

LucaMarconato:
The dask min version is high: we could avoid the new lower bound and warn the user that sharding is supported only from 2026.3.0.

Alternatively, since also 2025.12.0 is already quite high, introduce the new lower bound.

src/spatialdata/_core/_utils.py
```diff
@@ -164,3 +165,36 @@ def get_unique_name(name: str, attr: str, is_dataframe_column: bool = False) ->
         setattr(sanitized, attr, new_dict)
 
     return None if inplace else sanitized
+
+
+def create_raster_element_kwargs(
+    raster_write_kwargs: dict[str, dict[str, Any] | list[dict[str, Any]]] | list[dict[str, Any]],
+    element_name: str,
+    element_names: set[str],
+) -> dict[str, Any] | list[dict[str, Any]]:
```
LucaMarconato: This function would strongly benefit from a docstring. In particular explaining:
- what are the input types
- what the output types
- who is calling this function
- who is consuming the output of this function

In particular, if the input/output types are implicitly defined in a library, e.g. by Dask, I suggest to write that this is the format defined by X library.

src/spatialdata/_core/_utils.py
```diff
@@ -164,3 +165,36 @@ def get_unique_name(name: str, attr: str, is_dataframe_column: bool = False) ->
...
+    element_raster_write_kwargs = None
+    if isinstance(raster_write_kwargs, dict) and (kwargs := raster_write_kwargs.get(element_name)):
+        element_raster_write_kwargs = kwargs
+
+    if not element_raster_write_kwargs:
```
LucaMarconato: Why not just `else`? In general I would try to simplify a bit the branching of this code.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1173,6 +1194,13 @@ def write(
         store.close()
 
         for element_type, element_name, element in self.gen_elements():
+            element_raster_write_kwargs = None
```
LucaMarconato: Instead of calling this in `write()` and in `write_element()`, can we move this code to `_write_element()` so it appears only once?

src/spatialdata/_core/_utils.py
```diff
@@ -164,3 +165,36 @@ def get_unique_name(name: str, attr: str, is_dataframe_column: bool = False) ->
...
+        if isinstance(raster_write_kwargs, dict):
+            for name in element_names:
+                raster_write_kwargs.pop(name, None)
```
LucaMarconato: This seems to be mutating the user's input. We should make a copy instead.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1291,12 +1326,32 @@ def write_element(
         shapes_geometry_encoding
             Whether to use the WKB or geoarrow encoding for GeoParquet. See :meth:`geopandas.GeoDataFrame.to_parquet`
             for details. If None, uses the value from :attr:`spatialdata.settings.shapes_geometry_encoding`.
+        raster_write_kwargs
```
LucaMarconato, line 1352: Duplicated docstring. I would rather use the `docstring_parameter` decorator as shown here https://github.com/scverse/spatialdata/blob/d8bf265410d7c6148752ec266e8115656e92b801/src/spatialdata/_core/query/spatial_query.py#L46

src/spatialdata/_core/spatialdata.py
```diff
@@ -1155,7 +1156,27 @@ def write(
         shapes_geometry_encoding
             Whether to use the WKB or geoarrow encoding for GeoParquet. See :meth:`geopandas.GeoDataFrame.to_parquet`
             for details. If None, uses the value from :attr:`spatialdata.settings.shapes_geometry_encoding`.
-        """
+        raster_write_kwargs
```
LucaMarconato, line 1189: please replace with docstring parameter since this docstring is repeated later https://github.com/scverse/spatialdata/blob/d8bf265410d7c6148752ec266e8115656e92b801/src/spatialdata/_core/query/spatial_query.py#L46

src/spatialdata/_core/spatialdata.py
```diff
@@ -1155,7 +1156,27 @@ def write(
...
+            Storage options for raster elements.These options are passed to the zarr storage backend for writing and
```
LucaMarconato: Typo: please add space in ".These"

src/spatialdata/_core/_utils.py
LucaMarconato: I see that the input type is defined in the docstring of `write()`.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1155,7 +1156,27 @@ def write(
...
+            2. Dictionary per raster element
+                A dictionary where:
+                - Keys = names of raster elements
+                - Values = storage options for each element
+                    - For single-scale data: a dictionary
+                    - For multiscale data: a list of dictionaries (one per scale)
+            3. List of dictionaries (multiscale only)
+                A list where each dictionary defines the storage options for one scale of a multiscale raster element.
+
+            Important Notes
```
LucaMarconato: Here I would strongly recommend adding an example for users so that they have a recommendation on what to write, at least for Zarr v3. The tests already contain this information.

melonora:
> The dask min version is high: we could avoid the new lower bound and warn the user that sharding is supported only from 2026.3.0.
> 
> Alternatively, since also 2025.12.0 is already quite high, introduce the new lower bound.

Sadly, we can't because of the problem of the partition collapse when reading from parquet. The dask versions that used the kwargs as keyword argument allowed for writing zarr v3 in a zarr v2 root group without throwing an error.

src/spatialdata/_core/_utils.py
```diff
@@ -164,3 +165,36 @@ def get_unique_name(name: str, attr: str, is_dataframe_column: bool = False) ->
         setattr(sanitized, attr, new_dict)
 
     return None if inplace else sanitized
+
+
+def create_raster_element_kwargs(
+    raster_write_kwargs: dict[str, dict[str, Any] | list[dict[str, Any]]] | list[dict[str, Any]],
```
LucaMarconato: In `io_raster.py` we use `JSONDict`. I'd use that also here, which makes the type more readable.

src/spatialdata/_io/io_raster.py
```diff
@@ -283,12 +283,27 @@ def _write_raster(
     raster_format
         The format used to write the raster data.
     storage_options
-        Additional options for writing the raster data, like chunks and compression.
+        Storage options for raster elements.These options are passed to the zarr storage backend for writing and
```
LucaMarconato: Similar to the docstring in write(). I wonder if we can simplify this docstring by importing some subdosctrings from the other one (like the Important Notes part). But maybe not worth overengineering.

src/spatialdata/_core/_utils.py
melonora: yeah I did it there because this is private API and write is public API, but can add it to private API as well if prefered.

src/spatialdata/_io/io_raster.py
LucaMarconato: Typo: needs a space in `".These"`.

src/spatialdata/_core/_utils.py
melonora: Please reopen if you would prefer that.

src/spatialdata/_io/io_raster.py
```diff
@@ -283,12 +283,27 @@ def _write_raster(
     raster_format
         The format used to write the raster data.
     storage_options
```
LucaMarconato, line 286: It's worth mentioning that these storage options have been validated via `create_raster_element_kwargs()`.

src/spatialdata/_core/spatialdata.py
melonora: Do you mean as part of the docstring or actual doc? I would provide a follow up PR with specific docs.

src/spatialdata/_io/io_raster.py
```diff
@@ -305,6 +320,24 @@ def _write_raster(
         for c in channels:
             metadata["metadata"]["omero"]["channels"].append({"label": c})  # type: ignore[union-attr, index, call-overload]
 
+    if isinstance(storage_options, dict):
+        storage_options = {
+            **{k.split("_")[1]: v for k, v in asdict(settings).items() if k in ("raster_chunks", "raster_shards")},
```
LucaMarconato: I'd bundle this into an helper function since it is called 3 times.

src/spatialdata/_io/io_raster.py
```diff
@@ -305,6 +320,24 @@ def _write_raster(
...
+            {
+                **{k.split("_")[1]: v for k, v in asdict(settings).items() if k in ("raster_chunks", "raster_shards")},
+                **x,
+            }
+            for x in storage_options
+        ]
+    elif not storage_options:
+        storage_options = {
+            k.split("_")[1]: v for k, v in asdict(settings).items() if k in ("raster_chunks", "raster_shards")
+        }
```
LucaMarconato: I'd add a comment explaning why we are renaming the arguments (who expects the renamed values? ome-zarr-py, dask or zarr? Why don't we pass `chunks` and `shards` directly?)

src/spatialdata/__init__.py
```diff
@@ -131,6 +131,10 @@
     "settings",
 ]
 
+import zarr
+
+zarr.config.set({"codec_pipeline.path": "zarrs.ZarrsCodecPipeline"})
```
LucaMarconato: Summoning @ilan-gold and @flying-sheep to comment on this please.

ilan-gold:
> I'll loop in @ilan-gold and @flying-sheep. What's the plan for anndata on this?

I haven't made it a default dependency for `anndata` but it will be enabled by default if discovered.

src/spatialdata/__init__.py
ilan-gold: In general, I don't use global settings like this.  I would only use it in a targeted way where you know it is what you want.

pyproject.toml
```diff
@@ -26,8 +26,8 @@ dependencies = [
     "annsel>=0.1.2",
     "click",
     "dask-image",
-    "dask>=2025.12.0,<2026.1.2",
-    "distributed<2026.1.2",
+    "dask>=2026.3.0",
+    "distributed>=2026.3.0",
```
LucaMarconato: @melonora can we drop this dep altogether?

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     "xarray>=2024.10.0",
     "xarray-spatial>=0.3.5",
     "zarr>=3.0.0",
+    "zarrs",
```
LucaMarconato: Let's follow up on this.

src/spatialdata/config.py
```diff
@@ -1,15 +1,27 @@
...
+from platformdirs import user_config_dir
+
+
+def _config_path() -> Path:
+    """Return the platform-appropriate path to the user config file."""
+    return Path(user_config_dir(appname="spatialdata")) / "settings.json"
+
 
 @dataclass
 class Settings:
```
LucaMarconato, line 8: General comment. Sounds like this file is a duplicate of the settings mechanism from `scverse-misc`: https://github.com/scverse/scverse-misc/blob/main/src/scverse_misc/_settings.py.

I think `scverse-misc` doesn't support reading settings from `.json` (they support `.env` files), but since it's based on Pydantic, it would be a quick extension. 

I'll go trough the PR and review. In a follow-up PR we should probably align to use `scverse-misc`.

@ilia-kats @flying-sheep, devs behind `scverse-misc`, may have extra comments.

src/spatialdata/config.py
```diff
@@ -1,15 +1,27 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+import json
+import os
+from dataclasses import asdict, dataclass
+from pathlib import Path
 from typing import Literal
 
+from platformdirs import user_config_dir
```
LucaMarconato: `platformdirs` is a transitive dependency; it should be added to `pyproject.toml`.

src/spatialdata/config.py
```diff
@@ -18,13 +30,161 @@ class Settings:
         Chunk sizes bigger than this value (bytes) can trigger a compression error.
         See https://github.com/scverse/spatialdata/issues/812#issuecomment-2559380276
         If detected during parsing/validation, a warning is raised.
+    raster_chunks
+        The chunksize to use for chunking an array. Length of the tuple must match
+        the number of dimensions.
+    raster_shards
+        The default shard size (zarr v3) to use when storing arrays. Length of the tuple
+        must match the number of dimensions.
```
LucaMarconato: Feels a bit weird to have these 2 variables in a class that is for global settings, since, if I am getting this right, we for cyx data we would need `len(raster_chunks) == 3`, while for `yx` data we would need `len(raster_chunks) == 2`.

src/spatialdata/config.py
```diff
@@ -18,13 +30,161 @@ class Settings:
...
     """
 
+    custom_config_path: Path | None = None
```
LucaMarconato: I don't think the settings should contain this path. Feels out of place.

src/spatialdata/config.py
```diff
@@ -18,13 +30,161 @@ class Settings:
...
     shapes_geometry_encoding: Literal["WKB", "geoarrow"] = "WKB"
```
LucaMarconato: It feels weird because I could have a situation like the following:

`pathA/settings.json` containing `custom_config_path == pathB/settings.json`
`pathB/settings.json` containing `custom_config_path == pathA/settings.json`

src/spatialdata/config.py
```diff
@@ -18,13 +30,161 @@ class Settings:
...
+            global_path = _config_path()
+            global_path.parent.mkdir(parents=True, exist_ok=True)
+            try:
+                with global_path.open(encoding="utf-8") as f:
+                    global_data = json.load(f)
+            except (json.JSONDecodeError, OSError):
+                global_data = {}
+            global_data["custom_config_path"] = str(target)
+            with global_path.open("w", encoding="utf-8") as f:
+                json.dump(global_data, f, indent=2)
```
LucaMarconato: This handles the circular path scenario that I described here: https://github.com/scverse/spatialdata/pull/1106#discussion_r3227091703

src/spatialdata/config.py
```diff
@@ -18,13 +30,161 @@ class Settings:
...
+    def apply_env(self) -> None:
+        """Apply environment variable overrides on top of the current state.
+
+        Env vars take precedence over both the config file and any
+        in-session assignments. Useful in CI pipelines or HPC clusters
+        where you cannot edit the config file.
+
+        Supported variables
+        -------------------
+        SPATIALDATA_CUSTOM_CONFIG_PATH -> custom_config_path
```
LucaMarconato: inconsistent arrow font

src/spatialdata/config.py
```diff
@@ -18,13 +30,161 @@ class Settings:
...
+        SPATIALDATA_SHAPES_GEOMETRY_ENCODING   → shapes_geometry_encoding
+        SPATIALDATA_LARGE_CHUNK_THRESHOLD_BYTES → large_chunk_threshold_bytes
+        SPATIALDATA_RASTER_CHUNKS                     → chunks
+        SPATIALDATA_RASTER_SHARDS                     → shards (integer or "none")
```
LucaMarconato: replace "chunks" with "raster_chunks" and "shards" with "raster_shards"?

src/spatialdata/config.py
```diff
@@ -1,15 +1,27 @@
...
+from dataclasses import asdict, dataclass
+from pathlib import Path
 from typing import Literal
 
+from platformdirs import user_config_dir
+
+
+def _config_path() -> Path:
+    """Return the platform-appropriate path to the user config file."""
+    return Path(user_config_dir(appname="spatialdata")) / "settings.json"
```
LucaMarconato: why not moving this inside `config_path()`?

tests/utils/test_config.py
```diff
@@ -0,0 +1,61 @@
...
+        s.large_chunk_threshold_bytes = 1_000_000_000
+        s.raster_chunks = (512, 512)
+        s.raster_shards = (1024, 1024)
+        s.save()
+        s = settings_cls().load()
+        assert s.shapes_geometry_encoding == "geoarrow"
+        assert s.large_chunk_threshold_bytes == 1_000_000_000
+        assert s.raster_chunks == [512, 512]
+        assert s.raster_shards == [1024, 1024]
+        assert s.custom_config_path is None
```
LucaMarconato: Fine. But since we called `save()`, shouldn't `custom_config_path` be equal to the path we are reading from? 

This is a consequence of https://github.com/melonora/spatialdata/blob/f44221ab23d6b136084157203e1d0b466979318c/src/spatialdata/config.py#L65

`path` is `None`, so it gets assigned the value from `_config_path()`, without saving it in the in-memory value of `custom_config_path`.

tests/utils/test_config.py
```diff
@@ -0,0 +1,61 @@
...
+        s = settings_cls().load()
+        assert s.shapes_geometry_encoding == "geoarrow"
+        assert s.large_chunk_threshold_bytes == 1_111_111_111
+        assert s.raster_chunks == (40, 40, 40)
+        assert s.raster_shards is None
+        assert s.custom_config_path == str(target_path)
+
+        s.reset()
+        s.save()
+        assert s.custom_config_path is None  # This returns False
```
LucaMarconato: what's the comment about?

src/spatialdata/_core/_utils.py
LucaMarconato: I would put some context at least. Right now if a contributor sees that function, it can be quite confusing.

LucaMarconato:
> I haven't made it a default dependency for anndata but it will be enabled by default if discovered.

- [x] @melonora I would do the same that @ilan-gold is doing in `anndata`, and therefore drop the dependency (we can add this as an optional dependency).

src/spatialdata/__init__.py
LucaMarconato: I agree, @melonora I'd suggest removing this. We should still make it discoverable to the users somehow. We could write a small tutorial on how to enable this.

tests/io/test_readwrite.py
```diff
@@ -748,6 +770,181 @@ def test_single_scale_image_roundtrip_stays_dataarray(tmp_path: Path) -> None:
...
+        path,
+        raster_write_kwargs={name: {"chunks": write_chunks, "shards": write_shards}},
+    )
+
+    arr = zarr.open_group(path / zarr_subpath / name, mode="r")["s0"]
+    assert arr.chunks == write_chunks
+    assert arr.shards == write_shards
+
+    other_arr = zarr.open_group(path / zarr_subpath / other_name, mode="r")["s0"]
+    assert other_arr.chunks == base_chunks
```
LucaMarconato, line 976: we could add an explicit check that shards are `None` (?) here.

tests/io/test_readwrite.py
```diff
@@ -748,6 +770,181 @@ def test_single_scale_image_roundtrip_stays_dataarray(tmp_path: Path) -> None:
...
+
+
+def test_write_raster_elements_sharding_chunking(tmp_path: Path) -> None:
```
LucaMarconato, line 980: What are we testing that it is not tested before? That `write_element()` supports sharding? In that case I'd rename the test.

LucaMarconato (changes requested):

src/spatialdata/__init__.py
ilan-gold: You can also use it by default on certain methods.  I'm not arguing against its use in a library, just in this way. See: https://github.com/scverse/anndata/blob/main/src/anndata/_io/zarr.py#L30-L46 which is only used for the top-level `AnnData.write_zarr` and `anndata.read_zarr` because those are "bulk operations" where I don't expect users to be very opinionated.

LucaMarconato:
Looks good in general. There are a few decisions to be made, but once done most of the requested changes can be one-shot by an agent.

pyproject.toml
LucaMarconato: Shouldn't we bump to `ome_zarr>=0.16.0`

src/spatialdata/_io/io_raster.py
melonora: I would not overengineer for now, also dependent on the function the options can be slightly different.

src/spatialdata/config.py
melonora: I will add a comment. This is because I also got asked whether we could add these settings for tables.

src/spatialdata/_io/io_raster.py
melonora: because of future PR for table chunks / shards

src/spatialdata/__init__.py
melonora: ok can remove, but then maybe should we move this context manager to scverse-misc?

flying-sheep: that’s a real trivial one, just copy it!

I’d add a comment explaining why the warning filter is added (@ilan-gold can you link the issue pls?), otherwise the warning filter is just cargo-culted and not kept for a reason.

src/spatialdata/config.py
melonora: would this be something for this PR or follow up PR before release? I believe before settings were not stored at all in any case and I think in general we need a PR to move towards using scverse-misc, also with adding the context manager there.

src/spatialdata/config.py
melonora: want to split in image_shards/chunks and label or just remove all together? I would maybe vote for the former.

src/spatialdata/config.py
melonora: depends on whether certain settings are project specific I believe. This allows for that. Otherwise global settings on hpc could be shared by more than 1 user even though there are project specific needs.

src/spatialdata/_io/io_raster.py
melonora: I would not bundle this. Looking in hindsight the refactor is obvious and does not warrant a helper function.

melonora: will add the refactor with next commit

src/spatialdata/config.py
melonora: I believe this to be handled per the code you linked with next comment.

tests/io/test_readwrite.py
melonora: yeah did that independently because of issues downstream with non matching shards / chunks. These errors are specific and not connected to our implementation per se so I kept it separate. For me it indicates that by `_sharding_chunking` which is seen not to be tested in the other functions.

src/spatialdata/__init__.py
ilan-gold: The warning is there in case zarrs doesn't support the store type you passed in to `read_zarr`, could use a comment sure

src/spatialdata/_core/spatialdata.py
LucaMarconato: I'd put a quick example of how it's called right it in the docstring.

src/spatialdata/config.py
LucaMarconato: there is still the problem of 2D vs 3D. I'd remove altogether. We could use instead an helper function that computes reasonable defaults based on the input shape.

src/spatialdata/config.py
LucaMarconato: Since `config.py` works, we can do that on a follow up release to avoid delaying this PR even more. On the other hand, if there is time/energy/will, I'd go directly with `scverse-misc` so that we don't merge code that will be removed soon after. Up to you!

LucaMarconato (commented):
Changes look good so far.

tests/io/test_readwrite.py
melonora: don't think it is necessary to add here. The other writing tests basically don't specify it. This is particularly for when it is specified. If it is somehow still specified you would get permission errors straight away as the chunks and shards would not match.

src/spatialdata/_core/_utils.py
LucaMarconato: Yes please, a short docstring explaining why we need this function/the logic behind this would be important.

src/spatialdata/_core/_utils.py
LucaMarconato: This comment remains open, can you please address @melonora?

src/spatialdata/_core/spatialdata.py
LucaMarconato: This comment is still open, the following code block:
```python
         if element_type in ("images", "labels") and raster_write_kwargs:
                element_names = set(self.images.keys()).union(self.labels.keys())
                element_raster_write_kwargs = create_raster_element_kwargs(
                    raster_write_kwargs, element_name, element_names
                )
```
could be moved to `_write_element()` instead of being duplicate in `write()` and `write_element()`.

src/spatialdata/_core/_utils.py
LucaMarconato: This is still unresolved.

src/spatialdata/_core/spatialdata.py
LucaMarconato: Still open

src/spatialdata/_core/spatialdata.py
LucaMarconato: Checked, still open

src/spatialdata/_core/spatialdata.py
LucaMarconato: This is still unresolved. From the docstring alone I find it tricky to understand what to call. I suggest to add a link to the tests working with sharding -> no code duplication, no need to wait for an additional notebook.

src/spatialdata/_io/io_raster.py
LucaMarconato: This docstring is long and mostly overlapping with the versions found in write() and write_element(). It's really not ideal to have the versions dangling around.

src/spatialdata/_io/io_raster.py
LucaMarconato: Please can you add this as a comment for the maintainers? It's not obvious.

pyproject.toml
LucaMarconato: It's not optional, closing this conversation.

src/spatialdata/config.py
LucaMarconato: I see settings got reverted to what they were before the sharding PR. So I'll resolve this conversation.

LucaMarconato: We can move to `scverse-misc` settings in a future PR.

tests/io/test_readwrite.py
LucaMarconato: Not necessary, but I think it's still interesting to check that it is either `None` or the same value as for `chunks`, and that it doesn't change over releases. Could you please add it?

src/spatialdata/_utils.py
```diff
@@ -354,3 +356,19 @@ def _check_match_length_channels_c_dim(
...
+
+
+# TODO: get this in scverse-misc and import from there
+@contextmanager
+def zarrs_context() -> Generator[None, None, None]:
+    with (
+        zarr.config.set({"codec_pipeline.path": "zarrs.ZarrsCodecPipeline"}) if find_spec("zarrs") else nullcontext(),
+        warnings.catch_warnings() if find_spec("zarrs") else nullcontext(),
+    ):
+        if find_spec("zarrs"):
```
LucaMarconato, line 369: Please add this comment from @ilan-gold 

> The warning is there in case zarrs doesn't support the store type you passed in to read_zarr, could use a comment sure

LucaMarconato (changes requested):
@melonora thanks for the changes. I re-reviewed all the conversations (open and closed) and checked the new code changes. The most crucial things (zarrs, settings, dependencies) are resolved. There are still conversation open (I closed a few and reopened one), please can you address the remaining comments?

After this I will do a fast review; we can merge soon.

src/spatialdata/_core/spatialdata.py
```diff
@@ -1161,12 +1163,32 @@ def write(
...
+            - The available key–value pairs in these dictionaries depend on the Zarr format used for writing.
```
Tomaz-Vieira: I find it rather hard to know what is or isn't allowed here, as the argument type is very loose and correct parameters depend on what zarr will take. So this is hard for users in interactive mode to get right, and also hard for dependands in non-interactive mode to determine if they have well-formed storage options.

If we don't want (or need) to expose all zarr options to the user right now, we could for now just expose like

```python
    def write_element(
        ...
        raster_write_args: Mapping[str, RasterWriteArgs] | Sequence[RasterWriteArgs] | RasterWriteArgs
        ...
    ):
        ...

    class RasterWriteArgs:
        shards: ShardsLike | None = None

    # or maybe this, if we want to allow raw dicts

    class RasterWriteARgs(TypedDict):
        shards: ShardsLike | None = None

```

Personally, I also don't like being able to pass in dict or list or dict of dicts, and would rather just have a single way to pass in those parameters.

tests/io/test_readwrite.py
melonora: done, it should be None if it is not specified.

src/spatialdata/_core/spatialdata.py
melonora: There are some arguments that even vary based on whether you are writing zarr_format 2 versus zarr_format 3 which might even still change. This means we would have to keep on checking. We can, but that would be a decision to make.

src/spatialdata/_utils.py
melonora: added

src/spatialdata/_core/_utils.py
melonora: added, ultimately just to normalize to what down the line zarr.create_array would expect for one specific element.

src/spatialdata/_core/spatialdata.py
Tomaz-Vieira: That is indeed super annoying, but I don't think we should be surfacing the complexities (and instabilities!) of our dependencies onto users of `spatialdata`. If it's annoying for us to deal with it, it'll be even worse for random users to figure it out

src/spatialdata/_core/_utils.py
melonora: refactored

src/spatialdata/_core/_utils.py
melonora: did shallow copy as only top level keys are popped.

src/spatialdata/_io/io_raster.py
melonora: not relevant anymore

src/spatialdata/_core/_utils.py
```diff
@@ -164,3 +166,41 @@ def get_unique_name(name: str, attr: str, is_dataframe_column: bool = False) ->
         setattr(sanitized, attr, new_dict)
 
     return None if inplace else sanitized
+
+
+def create_raster_element_kwargs(
```
Tomaz-Vieira, line 171: Just to register my core argument from our discussion yesterday, I believe this function should rather be a constructor (or factory function) of a class with a name like `RasterElementWriteOptions`, rather than being a free-standing function that returns a loose `dict[str, Any]`.

If one sees a function that expects a `RasterElementWriteOptions` as an argument, then it is very natural to go find that class and a way to construct it, and everything that is or isn't accepted inside that class is immediately clear from the constructor signature and/or class field declarations. If we use these loose dictionaries, then the expectations for such dicts will only exist in comments or internal knowledge. It is also a lot harder to spot API breakages.

melonora:
@LucaMarconato @Tomaz-Vieira @ajkswamy documenting the decisions based on our discussion.

- @LucaMarconato we want to start with deprecating writing using zarr_format 2 as this would make it easier with regards to what kind of arguments we have to deal with. Otherwise we get different arguments for both formats that we have to take into account. This is in line with the plan of ome-zarr-py to start deprecating to old formats.
- write_kwargs for users still writing format 2 will throw an error

for writing zarr_format 3
- We will have a class constructor that would allow for the most basic arguments to be typed.
- We allow for raster_write_kwargs (curated to exclude for example name, dtype and storage_options) that allows the users to also specify other arguments in zarr.create_array, but this won't be typed for now.

@Tomaz-Vieira @ajkswamy am I missing anything?

LucaMarconato:
> After this I will do a fast review; we can merge soon.

@melonora ready for re-review? I will be off for 1 week, @Tomaz-Vieira if you already re-reviewed and give the green light we can merge.

melonora:
No I need to update this

PR #1107, opened 2026-04-17
feat: generalize Zarr I/O in `spatialdata` to local and add remote (fsspec / UPath) backends

selmanozleyen:
(disclaimer prettified with AI)

This PR generalizes Zarr I/O in `spatialdata` so local and remote
(fsspec / UPath) backends are treated uniformly, and tightens the write
path to be safe against accidental clobbering of remote data.

## Main changes

- **`ZarrStore` abstraction** (`src/spatialdata/_store.py`): a dataclass
  wrapping a `UPath` with its embedded fsspec filesystem, plus
  `make_zarr_store` / `make_zarr_store_from_group` constructors and
  `open_read_store` / `open_write_store` context managers. All Zarr opens
  go through these.
- **Read-only plumbing**: `_resolve_zarr_store(..., read_only=...)`
  forwards the flag to every `LocalStore` / `FsspecStore` construction
  site, so readers cannot accidentally write.
- **Arrow-filesystem I/O** for points / shapes Parquet (`io_points.py`,
  `io_shapes.py`): `read_parquet` / `to_parquet` now take
  `filesystem=arrow_fs` derived from the store's fsspec fs. Includes a
  fix for a `to_parquet` categorical-schema regression introduced by the
  handoff (demote `known=[]` categoricals back to `unknown` in
  `_read_points` so `write_points`'s `as_known()` can unify categories
  across partitions).
- **Overwrite semantics for remote targets**: top-level `SpatialData.write`
  to a *remote* `UPath` (i.e. not `PosixUPath` / `WindowsUPath`) requires
  `overwrite=True`, because we cannot reliably probe existence on
  arbitrary remotes. Local paths and internal per-element writes are
  unaffected.
- **zarr v3 cleanup**: removed dead `ConsolidatedMetadataStore` branches
  that were unreachable on the `zarr>=3.0.0` pin.
- **Dask task-graph adapter**: `_extract_parquet_paths_from_task`
  restores `get_dask_backing_files` for the post-PR-#1006 dask expression
  API (`FragmentWrapper.fragment.path`) while still tolerating the legacy
  `{"piece": (file, None, None)}` shape; the matching key filter is
  broadened to `"parquet" in name.lower()` to catch fused-expression keys
  such as `readparquetpyarrowfs-fused-values-*`.
- **Legacy branches are tagged, not hidden**: every `getattr` branch that
  exists only to tolerate older dask/zarr shapes carries a `TODO(...)`
  marker with an explicit retirement condition
  (`TODO(legacy-dask)`, `TODO(dask-task-api)`, `TODO(zarr-v3-store-path)`,
  `TODO(async-pyarrow-fs)`).
- **Abstraction stress tests** (`tests/io/test_store_abstractions.py`):
  memory-fsspec round-trip of every element type, HTTP-like no-listing
  filesystem, tamper-evident read, and write-side consolidated-metadata
  coverage. Plus unit tests in `tests/io/test_store.py` for
  `_resolve_zarr_store` and `ZarrStore` construction.

## Relation to prior work

This branch was forked off
[`SamirMoustafa:cloud-storage-support`](https://github.com/SamirMoustafa/spatialdata/tree/cloud-storage-support),
which introduced the initial plumbing for remote storage and a dedicated
CI job for remote-storage tests. Building on that, this PR:

- **Kept**: the motivation of supporting UPath-addressable remote backends
  end-to-end.
- **Refactored**: the ad-hoc `path` / `storage_options` threading into a
  single `ZarrStore` value object, so every reader and writer takes one
  argument instead of reconstructing filesystems downstream.
- **Replaced**: the provider-specific (S3 / Azure) test scaffolding with
  provider-agnostic `fsspec.filesystem("memory")` stress tests. No moto /
  Azurite needed; the abstractions are exercised against a synthetic
  remote that mimics the awkward parts of cloud backends (no directory
  listing, read-only enforcement, byte-level tamper detection).
- **Added**: the read-only store mode, the remote-overwrite guard, the
  categorical fix, and the dask task-graph adapter.

The abstraction-stress tests here cover the same correctness invariants
[`#971`](https://github.com/scverse/spatialdata/pull/971) targeted on the
remote side, without requiring cloud credentials or emulators. Real-network
integration can be a follow-up behind a `--run-remote` marker.

## Known gap left deliberately open

Not hidden behind `@pytest.mark.xfail` -- I would rather have reviewer
input on where to wire the fix than merge it silenced.

**Consolidated metadata is written but not consumed on read.**
`tests/io/test_store_abstractions.py::TestConsolidatedMetadataOnRead::test_read_zarr_opens_via_consolidated_metadata`
fails with `saw 120 GETs < 10`: reading a single-image `SpatialData` from a
memory UPath currently issues ~120 small `cat_file` calls because
`read_zarr` opens the root with `zarr.open_group(store, mode="r")` and
does not pass `use_consolidated=True`. This is a **latency** regression
over remote stores only; correctness is unaffected. Suggestions on
whether the fix should live in `read_zarr` itself, in `open_read_store`,
or be opt-in via a parameter are very welcome.

> The previously-listed "Dask `Task`-iteration" gap (introduced by
> [`53b9438a`](https://github.com/scverse/spatialdata/commit/53b9438a328c5fc2a451d2c8afab439b945ba2b8),
> PR [#1006](https://github.com/scverse/spatialdata/pull/1006)) is now
> **fixed** in this PR via the task-graph adapter above.

## Test plan

- [ ] `pytest tests/io/test_store.py tests/io/test_store_abstractions.py`
  (expected: **18 passed, 1 failed** -- the consolidated-metadata
  invariant noted above).
- [ ] `pytest tests/io/test_readwrite.py` vs main (expected: matches
  baseline, no new regressions; the Task-iteration failures present on
  main are resolved here).
- [ ] Manual round-trip with a `memory://` UPath using `SpatialData.write`
  / `SpatialData.read`.
- [ ] `grep -n "TODO(" src/spatialdata/_store.py src/spatialdata/_io/_utils.py`
  lists four markers, each with a documented retirement condition.

cc: @melonora @SamirMoustafa


src/spatialdata/_store.py
```diff
@@ -0,0 +1,161 @@
...
+
+import pyarrow.fs as pafs
+import zarr
+from upath import UPath
+from zarr.storage import FsspecStore, LocalStore
+
+PathLike: TypeAlias = Path | UPath
+
+
+def normalize_path(path: str | PathLike, storage_options: dict[str, Any] | None = None) -> PathLike:
```
SamirMoustafa, line 14: `normalize_path(..., storage_options=...)` correctly passes options into `UPath(...)` for URLs. Keep error messages consistent across the codebase (some places mention `None`, some don’t)

src/spatialdata/_io/_utils.py
```diff
@@ -39,6 +39,12 @@
 from spatialdata.transformations.transformations import BaseTransformation, _get_current_output_axes
 
 
+def join_fsspec_store_path(store_path: str, relative_path: str) -> str:
```
SamirMoustafa, line 42: `join_fsspec_store_path()` is a simple helper. Maybe add a few small tests (empty path, leading `/`, nested path) to avoid future breakage

src/spatialdata/_core/spatialdata.py
```diff
@@ -1037,18 +1035,56 @@ def _symmetric_difference_with_zarr_store(self) -> tuple[list[str], list[str]]:
...
-        file_path: str | Path,
+        file_path: str | Path | UPath,
         overwrite: bool = False,
         saving_an_element: bool = False,
     ) -> None:
-        from spatialdata._io._utils import _backed_elements_contained_in_path, _is_subfolder, _resolve_zarr_store
+        """
+        Guard against unsafe writes for **local** paths (zarr check, Dask backing, subfolders).
+
+        For :class:`upath.UPath`, ``overwrite=False`` is rejected: we cannot reliably check
```
SamirMoustafa, line 1045: Remote write guard: for remote `UPath`, `overwrite=False` is rejected because “exists” checks are not reliable. Consider a a clearer error type than `NotImplementedError` (I think it's a safety rule, not a missing feature)

src/spatialdata/_io/io_points.py
```diff
@@ -13,34 +11,43 @@
...
+    version = _parse_version(group, expect_attrs_key=True)
     assert version is not None
     points_format = PointsFormats[version]
 
-    store_root = f.store_path.store.root
-    path = store_root / f.path / "points.parquet"
-    # cache on remote file needed for parquet reader to work
-    # TODO: allow reading in the metadata without caching all the data
-    points = read_parquet("simplecache::" + str(path) if str(path).startswith("http") else path)
+    fs, parquet_path = arrow_fs_and_path(group, "points.parquet")
```
SamirMoustafa: `read_parquet(..., filesystem=fs)` can create broken empty categoricals you fix that right after reading!

SamirMoustafa (commented):
1. `tests/core/operations/test_transform.py`: CI fails on `test_transform_points_with_multiple_partitions` (`cannot reshape ...`)
2. `tests/core/test_centroids.py`: After we fixed the transform failure, centroids tests may fail because the centroid table columns came out as `y, x` but the test expects `x, y`

These files haven’t changed in the PR, but the failure blocks the merge.

src/spatialdata/_core/spatialdata.py
selmanozleyen: But with icechunk this feature might be implemented. Thats why I raise a not implemented error

SamirMoustafa (commented):
Thanks again for the PR and for helping get it done.

While testing, I used a small script [example_remote_storage.py](https://github.com/user-attachments/files/28360634/example_remote_storage.py) to write and read SpatialData on remote storage. `memory://` worked with the PR as-is. For Azure (`abfs://`), I had to add a few workarounds for points/shapes parquet (basically writing to a temp file and uploading to blob storage), because adlfs didn’t work well with the normal fsspec path.

I’m not asking to include that script in the PR, as the tests here are the right check. But I’m sharing it as it might helps later when someone wants a proper, generic fix instead of the patches I used

selmanozleyen:
Do you need anything from my side now?

LucaMarconato:
Thanks for the review and updates! We'll coordinate the review with @melonora. It may take a while, the next two weeks are filled with non-code duties.

SamirMoustafa:
> Do you need anything from my side now?

From my side, there are no further requests. Once @LucaMarconato and @melonora are okay with it, I expect it can be merged soon.
Thanks again @selmanozleyen, for the support and for working through this with me

pennycuda:
Hello! I have another open PR at #1141 , but my colleague Mark pointed out that this PR existed and was close to being merged. My original PR involved changes to allow zipped zarr stores as inputs for SpatialData and ome-zarr-py. This is due to HuBMAP Consortium's desire to minimize the amount of files we store while also maintaining our requirement that our output files be easily readable by end users. 

When I made my PR to ome-zarr-py https://github.com/ome/ome-zarr-py/pull/587 , one of their developers @will-moore mentioned that they were overhauling the IO usage and close to merging a new reading and writing API with this PR: https://github.com/ome/ome-zarr-py/pull/515 This PR that Will pointed me to makes the use of `parse_url()` and `ZarrLocation()` unnecessary and the new reading methods are much nicer to work with.

So, I cloned both this SpatialData PR branch and that ome-zarr-py PR branch and added the new ome-zarr-py functionality as well as the ability to accept zipped zarr stores as inputs to SpatialData with my branch here https://github.com/hubmapconsortium/spatialdata/tree/pennycuda/zip-store-rework

I understand that this is a kind of messy situation where I want to merge into two unmerged PRs that were closer to being merged than my original PR, so please let me know how to proceed! 

Thank you! 

Penny Cuda, she/her HuBMAP Consortium

selmanozleyen:
Hi! it's totally fine! In fact it's nice to see efforts not being duplicated here!

pennycuda:
@selmanozleyen Great! Should I make a PR to you then with my branch that includes my additions for ZipStore support and the updated but not yet merged ome-zarr-py usage?

selmanozleyen:
Depends on how urgent it is. I think to wait for this to be merged would be easier for the reviewers to have them separate atm. And you can rebase after this is merged

mruffalo:
Thanks @selmanozleyen -- there are a lot of stakeholders at HuBMAP that are eager to allow reading from a zipped Zarr store, since it's much more practical to distribute single zip files than whole directory trees when sharing processing results. Do you have a sense of the timeline to merge this PR and then move on to the zip-specific functionality? Can we help with any development or testing resources?

Much appreciated!

selmanozleyen:
@LucaMarconato and @melonora can give a more concrete answer

pennycuda:
Hi @selmanozleyen, I was pointed to this zarr PR https://github.com/zarr-developers/zarr-python/pull/3612 that "allows users to move specific zarr groups to any destination store. Also, avoiding the whole decoding / encoding step (copying raw bytes)." I think it could be of use to you in your changes involving Groups!

src/spatialdata/_core/spatialdata.py
ianhi: 👋 hi @selmanozleyen icechunk developer with a bio background and calling in  @li-em who even has a omics background (im a microscopist). We're happy to chat about icechunk in spatialdata space whenever you want!

selmanozleyen: Hi, thanks. But it's really not up to me. It's up to the maintainers really so we move on with this PR or not. If we are fine with not supporting shapes (because we currently represent shapes in zarr), then this is going to be smooth

selmanozleyen: Long term, we have on disk format to represent shapes as well in zarr

melonora: @ianhi Hey there! I send you a dm on the scverse zulip

SamirMoustafa:
@LucaMarconato @melonora Is this PR currently blocked by the recent icechunk/shapes discussion, or is there anything else needed to merge?

PR #1113, opened 2026-05-05
Reduce warnings

LucaMarconato:
Reduce warnings, in particular:
- Use zip(strict=True) when converting tuple chunks to dict in RasterSchema
- Copy AnnData views before modifying .uns in TableModel.parse
- Use explicit string obs index in AnnData construction to avoid implicit index warnings
- Add .copy() after compute()/slice to avoid pandas SettingWithCopyWarning (CoW)
- Remove unused docs/_templates autosummary files
- Update tutorials submodule
- Fix pandas CoW issues


LucaMarconato:
Recap:
- Reduced the number of warnings in general
- Approximately halved the numbers of warnings emitted (and not caught) during tests. ~100 left
- Greatly reduced the number of logs (INFO, WARNING), during tests.

PR #1114, opened 2026-05-05
Add Python 3.14 support

LucaMarconato:
Add tests for Python 3.14 to ensure it is supported.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,12 +19,15 @@ jobs:
             fail-fast: false
             matrix:
                 include:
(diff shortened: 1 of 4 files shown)


Zethson:
Since you're already at it, might as well drop 3.11?
We agreed to follow https://scientific-python.org/specs/spec-0000/
Might make your life easier.

LucaMarconato:
I would prefer to drop 3.11 only when it breaks; any time now, since with SPEC-0 we (try to) guarantee only 3.12+. @melonora WDYT?

LucaMarconato:
I'm merge so that the CI and integration testing are back to green. We can iterate for SPEC-0.

PR #1116, opened 2026-05-05
Reduce time cost of tests

LucaMarconato:
Reducing the time required to run tests. This complements https://github.com/scverse/spatialdata/pull/1113, which reduces the number of emitted warnings (@melonora observed that this was impacting the runtime).


LucaMarconato:
This PR also adjusts the tests in the CI (modified in https://github.com/scverse/spatialdata/pull/1114) to keep only lower-bound and upper-bound Python versions, which is the philosophy used in [integration-testing](https://github.com/scverse/integration-testing/actions/runs/25354708073/job/74341456248). 

Actually since we test 3.11 and 3.14, and integration testing tests 3.12 every night, we cover also the 3.12 case, while keeping the per-commit CI leaner.

LucaMarconato:
So far the tests should be running ~12% faster than in `main`. We could merge already, or optimize further.

LucaMarconato:
The last fail is unrelated to the code change and likely due to the parallel test execution. I'll investigate.

LucaMarconato:
Tests are significantly faster now! 

This below is from `main`
image

This is with this PR:
image

LucaMarconato:
Tests now locally run under 1 minute! 
image

PR #1124, opened 2026-05-11
Split `SpatialData` objects via `deconcatenate()`

LucaMarconato:
Follow up on the work from @gboscagli started here #1089.


LucaMarconato:
@gboscagli I started from your code and reworked a bit. I changed a bit the arguments and tried to infer more parameters by default. Could you try it out/comment please? 😊

gboscagli:
@LucaMarconato I tried the new function and it works nicely! I like the new implementation and the idea of refactoring `deconcatenate()` inside `concatenate()`.

Long term, we could think of deconcatenating images also, as we discussed in scverse/spatialdata-io#318.

PR #1149, opened 2026-06-19
Add downloadable cells dataset via scverse-misc

timtreis:
## Summary

Exposes `spatialdata.datasets.cells()` alongside `blobs`/`raccoon`. It downloads the `cells` example dataset and loads it as a `SpatialData` object.

Dataset fetching is delegated to `scverse-misc[datasets]` (>=0.1.0, the first released version with the datasets feature). The registry shipped in `src/spatialdata/datasets.yaml` is parsed and fetched (downloaded, hash-verified, cached, loaded) at call time.

The actual download is exercised by `test_cells_download`, marked `slow`; the network-free `test_cells_registry` validates the shipped registry on every run.

### DIFF
--- a/src/spatialdata/datasets.py
+++ b/src/spatialdata/datasets.py
@@ -107,7 +107,8 @@ def cells(path: str | None = None) -> SpatialData:
     registry = importlib.resources.files("spatialdata").joinpath("datasets.yaml")
     with importlib.resources.as_file(registry) as registry_path:
         base_url, datasets = parse_registry(registry_path)
-    return fetch(datasets["cells"], cache_dir, base_url=base_url)
+    sdata: SpatialData = fetch(datasets["cells"], cache_dir, base_url=base_url)
+    return sdata
 
 
 class RaccoonDataset:
--- a/src/spatialdata/datasets.py
+++ b/src/spatialdata/datasets.py
@@ -31,7 +31,7 @@ from spatialdata.models import (
 )
 from spatialdata.transformations import Identity
 
-__all__ = ["blobs", "raccoon"]
+__all__ = ["blobs", "cells", "raccoon"]
 
 
 def blobs(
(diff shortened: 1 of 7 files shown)


LucaMarconato:
Didn't dive deep but this looks great. One important thing is missing, for each dataset we must list the license, and if the license requires attribution, we need to list that. Could you please add this?

timtreis:
@LucaMarconato added the license info 👌

tests/datasets/test_datasets.py
```diff
@@ -26,3 +31,26 @@ def test_datasets() -> None:
...
+    file = entry.file(name="cells.zip")
+    assert file.sha256 == "dc9613cb9e16fd2cd8d83f3a9586eeda4af5ba8ba366f1066efb51305820c5fb"
+    assert file.resolve_url(base_url) == "https://exampledata.scverse.org/spatialdata/cells.zip"
+
+
+@pytest.mark.slow
+def test_cells_download(tmp_path) -> None:
+    # Downloads ~3 MB from the scverse example data bucket; opt out with `-m "not slow"`.
+    sdata = cells(path=str(tmp_path))
+    assert isinstance(sdata, SpatialData)
```
ajkswamy, line 59: In case we want to write tests based on this dataset, it would be good if this test would verify some particular of the dataset similar to `test_datasets` above, like number of each type of elements expected, their shapes/sizes, etc.

src/spatialdata/datasets.py
```diff
@@ -79,6 +79,38 @@ def raccoon() -> SpatialData:
     return RaccoonDataset().raccoon()
 
 
+def cells(path: str | None = None) -> SpatialData:
+    """
+    Cells dataset.
```
ajkswamy, line 106: It would be good to include a brief description of what this dataset contains, to help developers adopt it.

src/spatialdata/datasets.py
```diff
@@ -79,6 +79,38 @@ def raccoon() -> SpatialData:
...
+    -------
+    SpatialData object with the cells dataset.
+    """
+    import importlib.resources
+    from pathlib import Path
+
+    import pooch
+    from scverse_misc.datasets import fetch, parse_registry
+
+    cache_dir = Path(path) if path is not None else Path(pooch.os_cache("spatialdata"))
```
ajkswamy: please also cover the "else" case here in tests.

src/spatialdata/datasets.py
```diff
@@ -79,6 +79,38 @@ def raccoon() -> SpatialData:
...
+    registry = importlib.resources.files("spatialdata").joinpath("datasets.yaml")
```
ajkswamy: I see the three lines here are also used in the unit test for this function. Would it make sense to make this into a separate function. Maybe it will be reused if we add more datasets that use this pattern from `spatialdata-misc`

ajkswamy (approved):
LGTM, here are some suggestions for improvements:

1. Please add a link to the associated PR in `spatialdata-notebooks`. Please also add a couple of lines describing the motivation for this PR along which what value it adds. I see the assocated PR in `spatialdata-notebooks` already has a couple of good lines in this direction.

tests/datasets/test_datasets.py
```diff
@@ -26,3 +32,41 @@ def test_datasets() -> None:
...
+
+def test_cache_dir() -> None:
+    # Network-free: both branches of the cache-directory resolution.
+    assert _cache_dir("/tmp/example") == Path("/tmp/example")
+    assert _cache_dir(None) == Path(pooch.os_cache("spatialdata"))
+
+
+@pytest.mark.slow
+def test_cells_download(tmp_path) -> None:
+    # Downloads ~3 MB from the scverse example data bucket; opt out with `-m "not slow"`.
```
LucaMarconato: Rather than opt-out, I'll try turning this into opt-in and skip the test by default.

LucaMarconato: Done, made opt-in and opting in in the GitHub CI. 

It's a small dataset here and not a big deal, but later when we test cloud support it's good that we differentiate between tests require a network and tests that do not.

LucaMarconato:
Thanks both @timtreis and @ajkswamy. The new dataset downloader works great! 

Made only some minor changes, auto-merging now.

PR #1150, opened 2026-06-19
feat(centroids): add return_area + SpatialData persist to get_centroids

timtreis:
## Motivation

`get_centroids` already computes per-label pixel counts (`np.bincount`) when finding label centroids, then discards them, we could store as `area`. This PR surfaces that area into an element's annotating table, squidpy-style (`obsm["spatial"]` / `obs["area"]`), so downstream tools (e.g. squidpy) and repeated renders can reuse them instead of recomputing.

stacked on https://github.com/scverse/spatialdata/pull/1151

### DIFF
--- a/src/spatialdata/_core/centroids.py
+++ b/src/spatialdata/_core/centroids.py
@@ -14,6 +14,7 @@ from xarray import DataArray, DataTree
 
 from spatialdata._core.operations.transform import transform
 from spatialdata._core.spatialdata import SpatialData
+from spatialdata._utils import _affine_matrix_multiplication
 from spatialdata.models import get_axes_names, get_table_keys
 from spatialdata.models._utils import SpatialElement
 from spatialdata.models.models import Labels2DModel, Labels3DModel, PointsModel, ShapesModel, get_model
@@ -23,6 +24,9 @@ from spatialdata.transformations.transformations import BaseTransformation, Iden
 BoundingBoxDescription = dict[str, tuple[float, float]]
 
 PersistAs = Literal["Points", "adata"]
+# squidpy-style storage keys for persist_as="adata".
+_SPATIAL_KEY = "spatial"
+_AREA_KEY = "area"
 
 
 def _validate_coordinate_system(e: SpatialElement, coordinate_system: str) -> None:
@@ -64,8 +68,7 @@ def _transform_centroid_coords(
     if isinstance(t, Identity):
         return xy
     matrix = t.to_affine_matrix(input_axes=tuple(axes), output_axes=tuple(axes))
-    n = len(axes)
-    return xy @ matrix[:n, :n].T + matrix[:n, n]
+    return _affine_matrix_multiplication(matrix, xy)
 
 
 @singledispatch
@@ -194,9 +197,7 @@ def _points_from_centroids(
     df: pd.DataFrame, area: np.ndarray | None, e: SpatialElement, coordinate_system: str
 ) -> DaskDataFrame:
     """Build a Points element from intrinsic centroids, transformed into ``coordinate_system``."""
-    out = df.copy()
-    if area is not None:
-        out["area"] = np.asarray(area, dtype=float)
+    out = df.assign(area=np.asarray(area, dtype=float)) if area is not None else df
     t = get_transformation(e, coordinate_system)
     assert isinstance(t, BaseTransformation)
     points = PointsModel.parse(out, transformations={coordinate_system: t})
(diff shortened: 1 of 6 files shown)


tests/core/test_centroids.py
```diff
@@ -172,9 +181,138 @@ def test_get_centroids_labels(
...
+
+
+def test_get_centroids_element_persist_adata_raises(labels):
+    # an element on its own has no annotating table; persist_as='adata' needs the SpatialData.
+    with pytest.raises(ValueError, match="persist_as='adata'"):
+        get_centroids(labels["labels2d"], persist_as="adata")
+
+
+def test_get_centroids_sdata_persist_into_table(full_sdata):
+    # `table` annotates `labels2d` (instance_id 0..99 == label values); background (0) -> NaN.
```
LucaMarconato: the comment is confusing, I think here with 0..99 we mean the iloc (the value of instance_id at index 0 is not 0). Better to change the comment.

tests/core/test_centroids.py
```diff
@@ -172,9 +181,138 @@ def test_get_centroids_labels(
...
+    ids, counts = np.unique(np.asarray(full_sdata["labels2d"].data), return_counts=True)
+    count_of = dict(zip(ids, counts, strict=True))
+    area = table.obs["area"].to_numpy()
+    for row, label_id in enumerate(inst):
+        if label_id != 0:
+            assert area[row] == count_of[label_id]
+
+
+def test_get_centroids_sdata_persist_fastpath_matches_transform(full_sdata):
+    # the in-memory affine fast path (adata) must equal the dask transform() path (element Points).
```
LucaMarconato: I see what it means, but still an obscure comment. Also `transform()` is not a `dask` method.

src/spatialdata/_core/centroids.py
```diff
@@ -1,24 +1,33 @@
...
 from spatialdata.transformations.operations import get_transformation
-from spatialdata.transformations.transformations import BaseTransformation
+from spatialdata.transformations.transformations import BaseTransformation, Identity
 
 BoundingBoxDescription = dict[str, tuple[float, float]]
 
+PersistAs = Literal["Points", "adata"]
+# squidpy-style storage keys for persist_as="adata".
+_SPATIAL_KEY = "spatial"
+_AREA_KEY = "area"
```
LucaMarconato: This should go in `models.py` inside `TableModel` since it's basically an addition to the file format.

src/spatialdata/_core/centroids.py
```diff
@@ -29,37 +38,92 @@ def _validate_coordinate_system(e: SpatialElement, coordinate_system: str) -> No
...
+    if coordinate_system is None and persist_as != "adata":
+        raise ValueError("`coordinate_system=None` (intrinsic coordinates) is only supported with persist_as='adata'.")
+
+
+def _transform_centroid_coords(
+    xy: np.ndarray, axes: list[str], e: SpatialElement, coordinate_system: str | None
+) -> np.ndarray:
+    """Apply the element's affine to centroid coords in-memory; ``None``/``Identity`` pass through.
+
+    ``axes`` is the column order of ``xy`` (e.g. ``["x", "y"]``).
```
LucaMarconato: this already works for any axes name and order, but the function makes it looks like it's only working for x and y. Better renaming.

src/spatialdata/_core/centroids.py
```diff
@@ -29,37 +38,92 @@ def _validate_coordinate_system(e: SpatialElement, coordinate_system: str) -> No
...
+    """
+    if coordinate_system is None:
+        return xy
+    t = get_transformation(e, coordinate_system)
+    assert isinstance(t, BaseTransformation)
+    if isinstance(t, Identity):
+        return xy
+    matrix = t.to_affine_matrix(input_axes=tuple(axes), output_axes=tuple(axes))
+    return _affine_matrix_multiplication(matrix, xy)
```
LucaMarconato: Probably the matrix multiplication is slower than just a transposition, but if this is a bottleneck we'll find out later during profiling. Let's keep as is.

src/spatialdata/_core/centroids.py
```diff
@@ -29,37 +38,92 @@ def _validate_coordinate_system(e: SpatialElement, coordinate_system: str) -> No
...
-    Get the centroids of the geometries contained in a SpatialElement, as a new Points element.
+    Get the centroids of the geometries contained in a SpatialElement.
 
     Parameters
     ----------
     e
-        The SpatialElement. Only points, shapes (circles, polygons and multipolygons) and labels are supported.
+        The SpatialElement (points, shapes — circles, polygons and multipolygons — or labels), or a
+        :class:`~spatialdata.SpatialData` object. When a ``SpatialData`` is passed, the second
+        positional argument is the name of the element to measure (see the ``SpatialData`` overload).
```
LucaMarconato, line 107: Not a big fan of this (positional argument depending on the single dispatch argument), but it works.

src/spatialdata/_core/centroids.py
```diff
@@ -29,37 +38,92 @@ def _validate_coordinate_system(e: SpatialElement, coordinate_system: str) -> No
...
-        The coordinate system in which the centroids are computed.
+        The coordinate system in which the centroids are computed. ``None`` returns the intrinsic
+        coordinates without applying any transformation (only supported with ``persist_as="adata"``).
     return_background
-        If True, the centroid of the background label (0) is included in the output.
+        If True, the centroid of the background label (0) is included in the output (labels only).
+    return_area
+        If True, also return the per-instance area: the pixel/voxel count for labels and the geometric
+        area for shapes (``pi * r**2`` for circles). Not supported for points (raises). With
+        ``persist_as="Points"`` the area is added as a feature column of the returned Points element.
```
LucaMarconato, line 116: There is a design problem: the area is always returned untransformed (=in the intrinsic coordinate system). I think we should be consistent: if we transform the centroids (and we do this) we should transform the area. Otherwise we should not transform the centroids (or we should not return the area unless we have an identity/intrinsic). 

I like to return the area, so I would consider always requiring `coordinate_system=None` (or mapped to a system where there is an `identity`) when we also compute the area.

src/spatialdata/_core/centroids.py
```diff
@@ -77,66 +141,210 @@ def _get_centroids_for_labels(xdata: xr.DataArray) -> pd.DataFrame:
...
+        area = df.pop("area").to_numpy() if return_area else None
+        return df, area, raster
+    if model is ShapesModel:
+        xy_df, area = _get_centroids_for_shapes(element, return_area)
+        return xy_df, area, element
+    if model is PointsModel:
+        if return_area:
+            raise ValueError("`return_area` is not supported for points elements (points have no area).")
+        axes = get_axes_names(element)
+        assert axes in [("x", "y"), ("x", "y", "z")]
```
LucaMarconato: we should turn this assert into a if raises

tests/core/test_centroids.py
```diff
@@ -172,9 +181,138 @@ def test_get_centroids_labels(
...
+def test_get_centroids_sdata_persist_instance_key_mismatch_raises(full_sdata):
+    # instance ids stored with a dtype that doesn't match the element's integer labels must fail
+    # loudly instead of silently writing NaN coordinates into obsm["spatial"].
+    table = full_sdata["table"]
+    table.obs["instance_id"] = table.obs["instance_id"].astype(str)
+    with pytest.raises(ValueError, match="No instance id annotating"):
+        get_centroids(full_sdata, "labels2d", persist_as="adata")
+
+
+def test_get_centroids_sdata_persist_refuses_dim_mismatch(full_sdata):
```
LucaMarconato, line 308: This edge case only occurs when a table is annotating multiple elements, and those elements have different axes (i.e. xyz vs xy). I would make this clear. ~~Note that the table could still be annotating some elements having   xy axes, and some elements having yx axes. I will make this comment also elsewhere.~~ (edit: there is no xy <> yx ambiguity, the ordering is always `x, y[, z]`).

src/spatialdata/_core/centroids.py
```diff
@@ -77,66 +141,210 @@ def _get_centroids_for_labels(xdata: xr.DataArray) -> pd.DataFrame:
...
+        )
+    if len(annotators) > 1:
+        raise ValueError(
+            f"Element {element_name!r} is annotated by multiple tables ({', '.join(annotators)}); "
+            f"pass `table_name=` to choose one."
+        )
+    return annotators[0]
+
+
+def _write_centroids_into_table(
```
LucaMarconato, line 275: This function seems to do the job, but I think there is a big design problem. Nothing stops the user in having a table annotating multiple elmements and where:
- ~~is annotating some elements with xy axes and some elements with yx axes -> the information of axes is lost once the centroids are in obsm~~
- the get_centroids function gets called on different coordinate systems for different elements -> the information of which coordinate systems the centroids belong to is lost once centroids are in obsm.

I have some ideas but not a clear solution in mind. This requires some work into definiting how the Spatial AnnData specification works.

src/spatialdata/_core/centroids.py
```diff
@@ -77,66 +141,210 @@ def _get_centroids_for_labels(xdata: xr.DataArray) -> pd.DataFrame:
...
+            return_background=return_background,
+            return_area=return_area,
+        )
+
+    # persist_as == "adata": resolve the annotating table and write the centroids into it.
+    if coordinate_system is not None:
+        _validate_coordinate_system(element, coordinate_system)
+    table_name = _resolve_annotating_table(e, element_name, table_name)
+    df, area, raster = _intrinsic_centroid_frame(element, return_background, return_area)
+    coord_cols = sorted(df.columns)  # canonical x, y[, z] (squidpy obsm["spatial"] order)
```
LucaMarconato, line 403: Important point to put in the docstring (and later in the specs).

LucaMarconato: (that the order is always x, y[, z])

src/spatialdata/_core/centroids.py
LucaMarconato: Actually the first is not a problem since the returned centroids are always `x, y[, z]` (due to the assertion above)
```
    axes = get_axes_names(element)
    assert axes in [("x", "y"), ("x", "y", "z")]
```

src/spatialdata/_core/spatialdata.py
```diff
@@ -543,6 +543,36 @@ def aggregate(
...
         )
 
+    def get_centroids(
+        self,
+        element_name: str,
+        coordinate_system: str | None = "global",
+        return_background: bool = False,
+        return_area: bool = False,
+        persist_as: Literal["Points", "adata"] = "Points",
+        table_name: str | None = None,
```
LucaMarconato, line 553: if we pass a table name for a table that does not exist, we could createa a table annotating the element. WDYT?

tests/core/test_centroids.py
LucaMarconato: I'd rewrite please.

src/spatialdata/_core/centroids.py
LucaMarconato: Let's keep as is.

tests/core/test_centroids.py
LucaMarconato: Action here: add a comment in the test explaining when this occurs.

src/spatialdata/_core/centroids.py
LucaMarconato: This requires to make a decision regarding the design, I don't have a solution in mind.

src/spatialdata/_core/centroids.py
LucaMarconato: The second point requires a design decision.

LucaMarconato:
Thanks @timtreis, the code looks good! Most of my comments are minor and they can be easily addressed (probably one-shot with an agent).

But there are 2 discussion items pointing to a design problem that we need to resolve. Happy to hear your thoughts on them!

src/spatialdata/_core/spatialdata.py
```diff
@@ -543,6 +543,39 @@ def aggregate(
             **kwargs,
         )
 
+    def get_centroids(
+        self,
+        element_name: str,
+        coordinate_system: str | None = "global",
+        return_background: bool = False,
+        return_area: bool = False,
```
selmanozleyen, line 551: hi, after talking with @francescadr, she said she'd ideally want to know where the area comes from i.e, if it's nucleus or cell boundaries. So can we perhaps make the area_key exposed and defaulted as `{input}_area` so that we know which boundaries it's computed from in the pipeline?

FrancescaDr: Yes I think this would be good to know! Also, at least in [Xenium](https://www.10xgenomics.com/support/software/xenium-onboard-analysis/1.9/analysis/xoa-output-understanding-outputs), the standard is to get the cell_area and nucleus_area separate. And since the cell_area is usually dependent on the segmentation and nucleus area not it would be good to differentiate.

LucaMarconato: Thanks for the feedback, that's useful to drive design decision from reported use cases! I will start a conversation about this in a issue and tag the people from Bioconductor since this would introduces a changes to the SpatialData format (so better to do it explicitly and not as a byproduct of a PR).

PR #1151, opened 2026-06-20
Migrate supported Python to 3.12, 3.13, 3.14

timtreis:
## Summary

Drops Python 3.11 and adds 3.14 as a supported version. anndata>=0.12 already requires `>=3.12`, so 3.11 was effectively broken.

## Notes
- Left as-is (already inside the new range): readthedocs pins 3.13, release workflow pins 3.12, Dockerfile uses system `python3`.
- ruff `pyupgrade` now targets py312; a `pre-commit run --all-files` pass may surface new auto-fixes.
- Compat to `ome_zarr>=0.18` needed two fixes:
  - **omero channel metadata:** 0.18's writers no longer emit the `omero` block we passed, so channel names were lost on write/read. spatialdata now writes it itself.
  - **integer labels:** 0.18 requires integer `label-value`s. `Labels{2,3}DModel.parse` now rejects float labels (meaningless for masks). **Breaking:** cast float labels to int, e.g. `.astype(np.uint16)`.

### DIFF
--- a/src/spatialdata/models/models.py
+++ b/src/spatialdata/models/models.py
@@ -399,6 +399,12 @@ class RasterSchema:
                 cls._check_chunk_size_not_too_large(data[d][name])
 
 
+def _validate_labels_dtype(data: DataArray | DataTree) -> None:
+    dtype = data.dtype if isinstance(data, DataArray) else data["scale0"]["image"].dtype
+    if not (np.issubdtype(dtype, np.integer) or np.issubdtype(dtype, np.bool_)):
+        raise ValueError(f"Labels must have an integer dtype, found {dtype}. Cast the data, e.g. `.astype(np.uint16)`.")
+
+
 class Labels2DModel(RasterSchema):
     dims = (Y, X)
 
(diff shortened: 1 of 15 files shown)


Zethson:
Let's update https://github.com/conda-forge/spatialdata-feedstock/blob/main/recipe/meta.yaml#L3 then as well and fix the broken recipe (doesn't support Python 3.13+) @LucaMarconato

tests/core/operations/test_rasterize_bins.py
```diff
@@ -276,8 +276,8 @@ def _get_sdata(n: int):
             value_key="instance_id",
         )
 
-    # if bins is a DataArray, it should hold integers
-    image = Labels2DModel.parse(RNG.normal(size=(3, 3)), dims=("y", "x"))
+    # if bins is a DataArray, it should hold integers (cast after parsing, which itself rejects floats)
+    image = Labels2DModel.parse(RNG.integers(0, 10, size=(3, 3)), dims=("y", "x")).astype(float)
```
LucaMarconato: this tests that `rasterize_bins()` is validating the labels element; validation should be performed by the model classes, not inside the rasterize APIs. I'll remove.

LucaMarconato: (before we needed this, but now that labels are integers by construction, we don't need this anymore)

src/spatialdata/models/models.py
```diff
@@ -399,6 +399,12 @@ def _check_chunk_size_not_too_large(cls, data: DataArray | DataTree) -> None:
                 cls._check_chunk_size_not_too_large(data[d][name])
 
 
+def _validate_labels_dtype(data: DataArray | DataTree) -> None:
```
LucaMarconato: This function is good, but it's in the wrong place. I moved it inside the `RasterSchema` class and I now call it when validating the labels types (not when parsing them).

Issue #1153, opened 2026-06-22
Add SPEC-0 badge


LucaMarconato:
It would be beneficial to implement a CI job that verifies our compliance with SPEC-0 and adds a badge to indicate the current status: https://github.com/scientific-python/specs/issues/240.

This job should run daily or weekly, rather than only on merge, as prolonged periods without code changes could result in us falling outside the SPEC-0 timeline. Consequently, the badge should turn red.

CC @ajkswamy

Zethson:
Actually, the scverse-integration test repository adheres to spec-0. So if you really want to, you can use that badge.
No need to implement a new workflow.

PR #1164, opened 2026-07-14
Transformation manager foundation

ajkswamy:
Part of tbd

## Increment 1: TransformationManager Class Foundation

Implement `TransformationManager` with coordinate system, element, and transformation management.
Validate against duplicate names and ensure elements belong to a CS before registration (i.e. create if not specified).
Add `transformation_manager: TransformationManager` as an attributed to `SpatialData` (not yet used).
This delivers the core data model and API for centralized transformation storage,
enabling inspection of transformation graphs without changing any existing behavior.

#### TransformationManager Internal Model

```python
import networkx as nx


class TransformationManager:
    def __init__(self):
        self.graph: nx.MultiDiGraph
        # has coordinate systems as `NgffCoordinateSystem` objects as nodes;
        # transforms as instances of BaseTransform are passed as attributes while adding edges
        self.element_to_cs_mapping: dict[str, NgffCoordinateSystem] = {}
        # maps `element_name` to the coordinate system to which the element belongs.
```

#### Notes:

- attributes of `TransformationManager` should be private (propery pattern)
- `element_name` is unique across all elements of a spatialdata object

##### Coordinate System Management

- [x] `add_coordinate_system(cs: NgffCoordinateSystem) -> None`
    - Register a new coordinate system.
    - Make sure to throw error if coordinate system already exists
- [x]  `remove_coordinate_system(cs: NgffCoordinateSystem) -> None`:
    - Remove an existing coordinate system.
    - Throw error if coordinate system has not yet been added
- [x] `list_coordinate_systems() -> list[NgffCoordinateSystem]`:
    - List all registered coordinate systems.

##### Element Management

- [x] `add_element(element_name: str) -> None`:
    - Register an element to an existing coordinate system
    - throw warning and do nothing if the element has already been registered
- [x] `unset_element(element_name: str) -> None`:
    - Unregister an element from the coordinate system to which it belongs
    - throw warning and do nothing if element has not been registered to any coordinate system
- [x] `get_element_coordinate_system(element_name: str) -> NgffCoordinateSystem`:
    - Get the name of the coordinate system to which an element belongs.
    - Throw error if `element_name` doesn't belong to any existing coordinate system

##### Transformation Management

- [x] `add_transformation(input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem, transformation: BaseTransformation) -> None`:

- Add a transformation between coordinate systems.
- throw error if either coordinate systems have not been added

- [x] `get_existing_transformation(input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem) -> BaseTransformation | None`:

- Retrieve a transformation between coordinate systems.
- throw error if either coordinate systems have not been added or if there is no direct transformation defined between
  them

- [x] `remove_transformation(input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem) -> None`:
    - Remove a transformation between coordinate systems.
    - throw error if either coordinate systems have not been added

### DIFF
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -57,8 +57,6 @@ _LAZY_IMPORTS: dict[str, str] = {
     # _core.query.spatial_query
     "bounding_box_query": "spatialdata._core.query.spatial_query",
     "polygon_query": "spatialdata._core.query.spatial_query",
-    # _core.transformation_manager
-    "TransformationManager": "spatialdata._core.transformation_manager",
     # _core.spatialdata
     "SpatialData": "spatialdata._core.spatialdata",
     # _io._utils
@@ -118,8 +116,6 @@ __all__ = [
     "bounding_box_query",
     "polygon_query",
     # _core.transformation_manager
-    "TransformationManager",
-    # _core.spatialdata
     "SpatialData",
     # _io._utils
     "get_dask_backing_files",
@@ -211,9 +207,6 @@ if TYPE_CHECKING:
 
     # _core.spatialdata
     from spatialdata._core.spatialdata import SpatialData
-
-    # _core.transformation_manager
-    from spatialdata._core.transformation_manager import TransformationManager
 
     # _io._utils
     from spatialdata._io._utils import get_dask_backing_files
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -57,9 +57,8 @@ _LAZY_IMPORTS: dict[str, str] = {
     # _core.query.spatial_query
     "bounding_box_query": "spatialdata._core.query.spatial_query",
     "polygon_query": "spatialdata._core.query.spatial_query",
-    # _core.transformation_manager
-    # "TransformationManager": "spatialdata._core.transformation_manager",
-    # _io._utils
+    # _core.spatialdata
+    "SpatialData": "spatialdata._core.spatialdata",
     "get_dask_backing_files": "spatialdata._io._utils",
     # _io.format
     "SpatialDataFormatType": "spatialdata._io.format",
@@ -210,7 +209,8 @@ if TYPE_CHECKING:
     from spatialdata._core.spatialdata import SpatialData
 
     # _core.transformation_manager
-    # from spatialdata._core.transformation_manager import TransformationManager
+    from spatialdata._core.transformation_manager import TransformationManager
+
     # _io._utils
     from spatialdata._io._utils import get_dask_backing_files
 
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -57,6 +57,8 @@ _LAZY_IMPORTS: dict[str, str] = {
     # _core.query.spatial_query
     "bounding_box_query": "spatialdata._core.query.spatial_query",
     "polygon_query": "spatialdata._core.query.spatial_query",
+    # _core.transformation_manager
+    "TransformationManager": "spatialdata._core.transformation_manager",
     # _core.spatialdata
     "SpatialData": "spatialdata._core.spatialdata",
     # _io._utils
@@ -115,6 +117,8 @@ __all__ = [
     # _core.query.spatial_query
     "bounding_box_query",
     "polygon_query",
+    # _core.transformation_manager
+    "TransformationManager",
     # _core.spatialdata
     "SpatialData",
     # _io._utils
@@ -207,6 +211,9 @@ if TYPE_CHECKING:
 
     # _core.spatialdata
     from spatialdata._core.spatialdata import SpatialData
+
+    # _core.transformation_manager
+    from spatialdata._core.transformation_manager import TransformationManager
 
     # _io._utils
     from spatialdata._io._utils import get_dask_backing_files
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -58,7 +58,7 @@ _LAZY_IMPORTS: dict[str, str] = {
     "bounding_box_query": "spatialdata._core.query.spatial_query",
     "polygon_query": "spatialdata._core.query.spatial_query",
     # _core.transformation_manager
-    "TransformationManager": "spatialdata._core.transformation_manager",
+    # "TransformationManager": "spatialdata._core.transformation_manager",
     # _io._utils
     "get_dask_backing_files": "spatialdata._io._utils",
     # _io.format
@@ -116,7 +116,6 @@ __all__ = [
     "bounding_box_query",
     "polygon_query",
     # _core.transformation_manager
-    "TransformationManager",
     # _core.spatialdata
     "SpatialData",
     # _io._utils
@@ -211,8 +210,7 @@ if TYPE_CHECKING:
     from spatialdata._core.spatialdata import SpatialData
 
     # _core.transformation_manager
-    from spatialdata._core.transformation_manager import TransformationManager
-
+    # from spatialdata._core.transformation_manager import TransformationManager
     # _io._utils
     from spatialdata._io._utils import get_dask_backing_files
 
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -57,8 +57,8 @@ _LAZY_IMPORTS: dict[str, str] = {
     # _core.query.spatial_query
     "bounding_box_query": "spatialdata._core.query.spatial_query",
     "polygon_query": "spatialdata._core.query.spatial_query",
-    # _core.spatialdata
-    "SpatialData": "spatialdata._core.spatialdata",
+    # _core.transformation_manager
+    "TransformationManager": "spatialdata._core.transformation_manager",
     # _io._utils
     "get_dask_backing_files": "spatialdata._io._utils",
     # _io.format
@@ -116,6 +116,8 @@ __all__ = [
     "bounding_box_query",
     "polygon_query",
     # _core.transformation_manager
+    "TransformationManager",
+    # _core.spatialdata
     "SpatialData",
     # _io._utils
     "get_dask_backing_files",
@@ -207,6 +209,9 @@ if TYPE_CHECKING:
 
     # _core.spatialdata
     from spatialdata._core.spatialdata import SpatialData
+
+    # _core.transformation_manager
+    from spatialdata._core.transformation_manager import TransformationManager
 
     # _io._utils
     from spatialdata._io._utils import get_dask_backing_files
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -114,7 +114,6 @@ __all__ = [
     # _core.query.spatial_query
     "bounding_box_query",
     "polygon_query",
-    # _core.transformation_manager
     # _core.spatialdata
     "SpatialData",
     # _io._utils
--- a/src/spatialdata/__init__.py
+++ b/src/spatialdata/__init__.py
@@ -207,9 +207,6 @@ if TYPE_CHECKING:
     # _core.spatialdata
     from spatialdata._core.spatialdata import SpatialData
 
-    # _core.transformation_manager
-    from spatialdata._core.transformation_manager import TransformationManager
-
     # _io._utils
     from spatialdata._io._utils import get_dask_backing_files
 
--- a/src/spatialdata/_core/transformation_manager/__init__.py
+++ b/src/spatialdata/_core/transformation_manager/__init__.py
@@ -290,7 +290,8 @@ class TransformationManager:
             If the element does not exist.
         """
         self.check_if_element_exists_else_raise_error(element_name)
-        return self.element_to_cs_mapping[element_name]
+        with suppress_direct_internal_attribute_access_warning():
+            return self.element_to_cs_mapping[element_name]
 
     def unset_element(self, element_name: str) -> None:
         """
@@ -525,7 +526,7 @@ class TransformationManager:
             return (
                 f"TransformationManager("
                 f"  coordinate_systems={list(self.graph.nodes())}, "
-                f"  coordinate_transforms={list(self.graph.edges())}, "
+                f"  coordinate_transforms={[x[TRANSFORM_KEY] for *_, x in self.graph.edges(data=True)]}, "
                 f"  elements={list(self.element_to_cs_mapping.keys())}"
                 f")"
             )
--- a/src/spatialdata/_core/transformation_manager/__init__.py
+++ b/src/spatialdata/_core/transformation_manager/__init__.py
@@ -131,7 +131,7 @@ class TransformationManager:
 
         self._graph.add_edge(input_cs, output_cs, **{TRANSFORM_KEY: transformation})
 
-    def get_transformation(
+    def get_existing_transformation(
         self, input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem
     ) -> BaseTransformation | None:
         """
@@ -210,7 +210,7 @@ class TransformationManager:
         if element_cs is None:
             raise KeyError(f"Element '{element_name}' does not belong to any coordinate system.")
 
-        return self.get_transformation(element_cs, target_cs)
+        return self.get_existing_transformation(element_cs, target_cs)
 
     def build_nx_graph(self) -> nx.MultiDiGraph:
         """
@@ -289,6 +289,86 @@ class TransformationManager:
                 sequence.append(edge_data[0][TRANSFORM_KEY])
             all_sequences.append(sequence)
         return all_sequences
+
+    def _get_transformations_associated_with_cs(
+        self, cs: NgffCoordinateSystem
+    ) -> list[tuple[NgffCoordinateSystem, NgffCoordinateSystem]]:
+        """
+        Get all transformations associated with a coordinate system.
+
+        Parameters
+        ----------
+        cs
+            The coordinate system to check.
+
+        Returns
+        -------
+        List of tuples representing transformations (input_cs, output_cs).
+        """
+        self.check_if_coordinate_system_exists(cs)
+
+        transformations = []
+        # Check outgoing edges (cs -> other)
+        for successor in self._graph.successors(cs):
+            transformations.append((cs, successor))
+        # Check incoming edges (other -> cs)
+        for predecessor in self._graph.predecessors(cs):
+            transformations.append((predecessor, cs))
+
+        return transformations
+
+    def _get_elements_belonging_to_cs(self, cs: NgffCoordinateSystem) -> list[str]:
+        """
+        Get all elements belonging to a coordinate system.
+
+        Parameters
+        ----------
+        cs
+            The coordinate system to check.
+
+        Returns
+        -------
+        List of element names belonging to the coordinate system.
+        """
+        self.check_if_coordinate_system_exists(cs)
+
+        elements = []
+        for element_name, element_cs in self._element_to_cs_mapping.items():
+            if element_cs == cs:
+                elements.append(element_name)
+
+        return elements
+
+    def remove_coordinate_system(self, cs: NgffCoordinateSystem) -> None:
+        """
+        Remove a coordinate system from the transformation manager.
+
+        Parameters
+        ----------
+        cs
+            The coordinate system to remove.
+
+        Raises
+        ------
+        ValueError
+            If the coordinate system has associated transformations or elements.
+        KeyError
+            If the coordinate system does not exist.
+        """
+        self.check_if_coordinate_system_exists(cs)
+
+        # Check if coordinate system has any transformations
+        if len(list(self._graph.edges(cs))) > 0:
+            raise ValueError(f"Cannot remove coordinate system '{cs.name}' as it has associated transformations")
+
+        # Check if coordinate system has any associated elements
+        associated_elements = self._get_elements_belonging_to_cs(cs)
+        if associated_elements:
+            raise ValueError(
+                f"Cannot remove coordinate system '{cs.name}' as it has associated elements: {associated_elements}"
+            )
+
+        self._graph.remove_node(cs)
 
     def __repr__(self) -> str:
         return (
--- a/src/spatialdata/_core/transformation_manager/__init__.py
+++ b/src/spatialdata/_core/transformation_manager/__init__.py
@@ -375,8 +375,8 @@ class TransformationManager:
         with suppress_direct_internal_attribute_access_warning():
             transforms = []
             assert target_cs in self.graph[source_cs], TransformationNotFoundError(source_cs.name, target_cs.name)
-            for _edge_key, edge in self.graph[source_cs][target_cs]:
-                transform: BaseTransformation = edge[TRANSFORM_KEY]
+            for edge_data in self.graph[source_cs][target_cs].values():
+                transform: BaseTransformation = edge_data[TRANSFORM_KEY]
                 transforms.append(transform)
             return transforms
 
@@ -415,7 +415,7 @@ class TransformationManager:
             )
             self.graph.remove_edge(source_cs, target_cs, key=expected_edge_key)
 
-    def remove_all_transformations(
+    def remove_all_transformations_between_coordinate_systems(
         self,
         source_cs: NgffCoordinateSystem,
         target_cs: NgffCoordinateSystem,
@@ -441,12 +441,13 @@ class TransformationManager:
         # also checks if source_cs and target_cs exist
         with suppress_direct_internal_attribute_access_warning():
             assert len(self.graph[source_cs][target_cs]), TransformationNotFoundError(source_cs.name, target_cs.name)
-            for edge_key, _edge in self.graph[source_cs][target_cs]:
+            for edge_key in list(self.graph[source_cs][target_cs].keys()):
+                # need to covert keys() to list to freeze it, else it will change during the following removal
                 self.graph.remove_edge(source_cs, target_cs, key=edge_key)
 
     def _get_transformation_sequences_from_path_after_disambiguation(
         self,
-        paths: Sequence[list[NgffCoordinateSystem]],
+        paths: list[list[NgffCoordinateSystem]],
         expected_intermediate_transformations: list[BaseTransformation] | None,
     ) -> list[list[BaseTransformation]]:
         """
@@ -471,11 +472,12 @@ class TransformationManager:
                 self._get_edge_key_from_transform(it) for it in expected_intermediate_transformations
             }
         all_sequences = []
-        for path in paths:
+        deduplicated_paths = list({repr(x): x for x in paths}.values())
+        for path in deduplicated_paths:
             sequence = []
             for i in range(len(path) - 1):
                 edge_data = self.graph[path[i]][path[i + 1]]
-                if len(edge_data) >= 1:
+                if len(edge_data) > 1:
                     # when there are multiple edges between a pair of coordinate systems in the path
                     intermediate_transformation_key_here = intermediate_transformation_edge_keys & set(edge_data.keys())
                     if len(intermediate_transformation_key_here) == 0:
@@ -486,7 +488,9 @@ class TransformationManager:
                     # choosing the first one arbitrarily
                     sequence.append(edge_data[edge_key_to_use][TRANSFORM_KEY])
                 else:
-                    sequence.append(edge_data[0][TRANSFORM_KEY])
+                    # Only one edge, no ambiguity
+                    edge_key = next(iter(edge_data.keys()))
+                    sequence.append(edge_data[edge_key][TRANSFORM_KEY])
             all_sequences.append(sequence)
         return all_sequences
 
@@ -528,7 +532,7 @@ class TransformationManager:
         """
         with suppress_direct_internal_attribute_access_warning():
             try:
-                paths = list(nx.shortest_simple_paths(self.graph, source=source_cs, target=target_cs))
+                paths = list(nx.all_shortest_paths(self.graph, source=source_cs, target=target_cs))
 
             except nx.NetworkXNoPath as nxe:
                 raise TransformationPathNotFoundError(source_cs.name, target_cs.name) from nxe
--- a/src/spatialdata/_core/transformation_manager/__init__.py
+++ b/src/spatialdata/_core/transformation_manager/__init__.py
@@ -5,9 +5,17 @@ import warnings
 import networkx as nx
 
 from spatialdata._core.transformation_manager.exceptions import (
+    CannotRemoveCoordinateSystemError,
+    CoordinateSystemAlreadyExistsError,
+    CoordinateSystemHasElementsError,
+    CoordinateSystemHasTransformationsError,
     CoordinateSystemNotFoundError,
+    ElementAlreadyExistsError,
     ElementNotFoundError,
+    InternalAttributeAccessWarning,
     TransformationNotFoundError,
+    TransformationPathNotFoundError,
+    suppress_direct_internal_attribute_access_warning,
 )
 from spatialdata.transformations.ngff.ngff_coordinate_system import NgffCoordinateSystem
 from spatialdata.transformations.transformations import BaseTransformation
@@ -31,7 +39,21 @@ class TransformationManager:
         Returns
         -------
         The MultiDiGraph containing coordinate systems and transformations.
+
+        Note
+        ----
+        Direct manipulation of the graph is discouraged. Please use the
+        TransformationManager methods whenever possible for better maintainability
+        and to ensure proper validation and error handling.
         """
+        warnings.warn(
+            "Direct access to the internal graph is discouraged. "
+            "Please use TransformationManager methods whenever possible for better "
+            "maintainability and to ensure proper validation and error handling.",
+            InternalAttributeAccessWarning,
+            stacklevel=2,
+        )
+
         return self._graph
 
     @property
@@ -42,10 +64,23 @@ class TransformationManager:
         Returns
         -------
         A dictionary mapping element names to their coordinate systems.
+
+        Note
+        ----
+        Direct manipulation of internal element_to_cs_mapping is discouraged. Please use the
+        TransformationManager methods whenever possible for better maintainability
+        and to ensure proper validation and error handling.
         """
+        warnings.warn(
+            "Direct access to the internal element_to_cs_mapping is discouraged. "
+            "Please use TransformationManager methods whenever possible for better "
+            "maintainability and to ensure proper validation and error handling",
+            InternalAttributeAccessWarning,
+            stacklevel=2,
+        )
         return self._element_to_cs_mapping
 
-    def check_if_element_exists(self, element_name: str) -> None:
+    def check_if_element_exists_else_raise_error(self, element_name: str) -> None:
         """
         Check if an element exists in the transformation manager.
 
@@ -59,10 +94,11 @@ class TransformationManager:
         ElementNotFoundError
             If the element does not exist.
         """
-        if element_name not in self.element_to_cs_mapping:
-            raise ElementNotFoundError(element_name)
+        with suppress_direct_internal_attribute_access_warning():
+            if element_name not in self.element_to_cs_mapping:
+                raise ElementNotFoundError(element_name)
 
-    def check_if_coordinate_system_exists(self, cs: NgffCoordinateSystem) -> None:
+    def check_if_coordinate_system_exists_else_raise_error(self, cs: NgffCoordinateSystem) -> None:
         """
         Check if a coordinate system exists in the graph.
 
@@ -76,31 +112,80 @@ class TransformationManager:
         CoordinateSystemNotFoundError
             If the coordinate system does not exist.
         """
-        if cs not in self.graph:
-            raise CoordinateSystemNotFoundError(cs.name)
+        with suppress_direct_internal_attribute_access_warning():
+            if cs not in self.graph:
+                raise CoordinateSystemNotFoundError(cs.name)
 
-    def check_if_edge_exists(self, input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem) -> None:
+    def check_if_edge_exists_else_raise_error(
+        self, source_cs: NgffCoordinateSystem, target_cs: NgffCoordinateSystem
+    ) -> None:
         """
         Check if an edge exists between coordinate systems.
 
         Parameters
         ----------
-        input_cs
+        source_cs
             The input coordinate system.
-        output_cs
+        target_cs
             The output coordinate system.
 
         Raises
         ------
+        CoordinateSystemNotFoundError
+            If either coordinate system does not exist.
         TransformationNotFoundError
             If the edge does not exist.
         """
-        if not self.graph.has_edge(input_cs, output_cs):
-            raise TransformationNotFoundError(input_cs.name, output_cs.name)
+        self.check_if_coordinate_system_exists_else_raise_error(source_cs)
+        self.check_if_coordinate_system_exists_else_raise_error(target_cs)
+        with suppress_direct_internal_attribute_access_warning():
+            if not self.graph.has_edge(source_cs, target_cs):
+                raise TransformationNotFoundError(source_cs.name, target_cs.name)
+
+    def check_if_coordinate_system_has_no_transformations_else_raise_error(self, cs: NgffCoordinateSystem) -> None:
+        """
+        Check if a coordinate system has associated transformations.
+
+        Parameters
+        ----------
+        cs
+            The coordinate system to check.
+
+        Raises
+        ------
+        CoordinateSystemNotFoundError
+            If either coordinate system does not exist.
+        CoordinateSystemHasTransformationsError
+            If the coordinate system has associated transformations.
+
+        """
+        transformations = self._get_transformations_associated_with_cs(cs)
+        # also checks if cs exists
+        if transformations:
+            raise CoordinateSystemHasTransformationsError(cs.name)
+
+    def check_if_coordinate_system_has_no_elements_else_raise_error(self, cs: NgffCoordinateSystem) -> None:
+        """
+        Check if a coordinate system has elements that belong to it.
+
+        Parameters
+        ----------
+        cs
+            The coordinate system to check
+
+        Raises
+        ------
+        Co
+
+        """
+        elements = self._get_elements_belonging_to_cs(cs)
+        # also checks if cs exists
+        if elements:
+            raise CoordinateSystemHasElementsError(cs.name, elements)
 
     def add_coordinate_system(self, cs: NgffCoordinateSystem) -> None:
         """
-        Register a new coordinate system.
+        Add a coordinate system to the transformation manager.
 
         Parameters
         ----------
@@ -109,12 +194,45 @@ class TransformationManager:
 
         Raises
         ------
-        ValueError
+        CoordinateSystemAlreadyExistsError
             If the coordinate system already exists.
         """
-        if cs in self.graph:
-            raise ValueError(f"Coordinate system '{cs.name}' already exists in the transformation manager")
-        self.graph.add_node(cs)
+        try:
+            self.check_if_coordinate_system_exists_else_raise_error(cs)
+        except CoordinateSystemNotFoundError:
+            with suppress_direct_internal_attribute_access_warning():
+                self.graph.add_node(cs)
+                return
+
+        raise CoordinateSystemAlreadyExistsError(cs.name)
+
+    def remove_coordinate_system(self, cs: NgffCoordinateSystem) -> None:
+        """
+        Remove a coordinate system from the transformation manager.
+
+        Parameters
+        ----------
+        cs
+            The coordinate system to remove.
+
+        Raises
+        ------
+        CoordinateSystemHasTransformationsError
+            If the coordinate system has associated transformations.
+        CoordinateSystemHasElementsError
+            If the coordinate system has associated elements.
+        CoordinateSystemNotFoundError
+            If the coordinate system is not found
+        """
+        try:
+            self.check_if_coordinate_system_has_no_transformations_else_raise_error(cs)
+            # also checks if cs exists
+            self.check_if_coordinate_system_has_no_elements_else_raise_error(cs)
+        except (CoordinateSystemHasTransformationsError, CoordinateSystemHasElementsError) as err:
+            raise CannotRemoveCoordinateSystemError(cs.name) from err
+
+        with suppress_direct_internal_attribute_access_warning():
+            self.graph.remove_node(cs)
 
     def list_coordinate_systems(self) -> list[NgffCoordinateSystem]:
         """
@@ -124,7 +242,8 @@ class TransformationManager:
         -------
         A list of coordinate system objects.
         """
-        return list(self.graph.nodes())
+        with suppress_direct_internal_attribute_access_warning():
+            return list(self.graph.nodes())
 
     def add_element(self, element_name: str, coordinate_system: NgffCoordinateSystem) -> None:
         """
@@ -137,33 +256,20 @@ class TransformationManager:
         coordinate_system
             The coordinate system to which the element belongs.
 
-        Warnings
-        --------
-        UserWarning
-            If the coordinate system does not exist.
+        Raises
+        ------
+        CoordinateSystemNotFoundError
+            If the coordinate system is not found.
         """
         try:
-            self.check_if_element_exists(element_name)
-        except ElementNotFoundError as _enfe:
-            warnings.warn(
-                f"Cannot add element with name '{element_name}') as it already "
-                f"exists in the transformation manager. Skipping",
-                UserWarning,
-                stacklevel=2,
-            )
+            self.check_if_element_exists_else_raise_error(element_name)
+        except ElementNotFoundError:
+            self.check_if_coordinate_system_exists_else_raise_error(coordinate_system)
+            with suppress_direct_internal_attribute_access_warning():
+                self.element_to_cs_mapping[element_name] = coordinate_system
+                return
 
-        try:
-            self.check_if_coordinate_system_exists(coordinate_system)
-        except CoordinateSystemNotFoundError as _csnfe:
-            warnings.warn(
-                f"Cannot set coordinate system ('{coordinate_system.name}') to element as the "
-                f"coordinate system does not exist.",
-                UserWarning,
-                stacklevel=2,
-            )
-            return
-
-        self.element_to_cs_mapping[element_name] = coordinate_system
+        raise ElementAlreadyExistsError(element_name)
 
     def get_element_coordinate_system(self, element_name: str) -> NgffCoordinateSystem:
         """
@@ -176,14 +282,14 @@ class TransformationManager:
 
         Returns
         -------
-        The coordinate system or None if not found
+        The coordinate system
 
         Raises
         ------
         ElementNotFoundError
             If the element does not exist.
         """
-        self.check_if_element_exists(element_name)
+        self.check_if_element_exists_else_raise_error(element_name)
         return self.element_to_cs_mapping[element_name]
 
     def unset_element(self, element_name: str) -> None:
@@ -200,20 +306,21 @@ class TransformationManager:
         ElementNotFoundError
             If the element has not been registered to any coordinate system.
         """
-        self.check_if_element_exists(element_name)
-        del self.element_to_cs_mapping[element_name]
+        self.check_if_element_exists_else_raise_error(element_name)
+        with suppress_direct_internal_attribute_access_warning():
+            del self.element_to_cs_mapping[element_name]
 
     def add_transformation(
-        self, input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem, transformation: BaseTransformation
+        self, source_cs: NgffCoordinateSystem, target_cs: NgffCoordinateSystem, transformation: BaseTransformation
     ) -> None:
         """
         Add a transformation between coordinate systems.
 
         Parameters
         ----------
-        input_cs
+        source_cs
             The input coordinate system.
-        output_cs
+        target_cs
             The output coordinate system.
         transformation
             The transformation to add.
@@ -223,22 +330,23 @@ class TransformationManager:
         CoordinateSystemNotFoundError
             If either coordinate system does not exist.
         """
-        self.check_if_coordinate_system_exists(input_cs)
-        self.check_if_coordinate_system_exists(output_cs)
+        self.check_if_coordinate_system_exists_else_raise_error(source_cs)
+        self.check_if_coordinate_system_exists_else_raise_error(target_cs)
 
-        self.graph.add_edge(input_cs, output_cs, **{TRANSFORM_KEY: transformation})
+        with suppress_direct_internal_attribute_access_warning():
+            self.graph.add_edge(source_cs, target_cs, **{TRANSFORM_KEY: transformation})
 
     def get_existing_transformation(
-        self, input_cs: NgffCoordinateSystem, output_cs: NgffCoordinateSystem
+        self, source_cs: NgffCoordinateSystem, target_cs: NgffCoordinateSystem
     ) -> BaseTransformation:
         """
         Retrieve a transformation defined between coordinate systems.
 
         Parameters
         ----------
-        input_cs
+        source_cs
             The input coordinate system.
-        output_cs
+        target_cs
             The output coordinate system.
 
         Returns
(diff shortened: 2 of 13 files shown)


src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+class TransformationManager:
+    """Centralized storage for managing coordinate systems (cs), transformations and element-cs mapping."""
+
+    def __init__(self) -> None:
+        """Initialize a Scene with empty mappings."""
+        self._coordinate_systems: dict[str, NgffCoordinateSystem] = {}
+        # mapping names of coordinate system to coordinate systems
+        self._coordinate_transforms: dict[tuple[str, str], BaseTransformation] = {}
```
Tomaz-Vieira: This likely should be a graph so we can find indirect paths between coordinate systems, and also so we could have multiple different edges between the same 2 CSs. I see below that you've made `nx` into an optional dependency, but it seems to me like we might need it as a hard dependency.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+from spatialdata.transformations.ngff.ngff_coordinate_system import NgffCoordinateSystem
+from spatialdata.transformations.transformations import BaseTransformation
+
+
+class TransformationManager:
+    """Centralized storage for managing coordinate systems (cs), transformations and element-cs mapping."""
+
+    def __init__(self) -> None:
+        """Initialize a Scene with empty mappings."""
+        self._coordinate_systems: dict[str, NgffCoordinateSystem] = {}
```
Tomaz-Vieira: I wonder if this extra level of indirection (i.e. CS name -> CS) is the best approach. I makes our APIs "stringly" typed, where they take and return `str` (like in `_get_transformations_associated_with_cs`) rather than working with something more semantically clear, which is the `NgffCoordinateSystem` itself.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+        # mapping names of coordinate system to coordinate systems
+        self._coordinate_transforms: dict[tuple[str, str], BaseTransformation] = {}
+        # mapping a tuple of coordinate system names, (<source coordinate system name>, <target coordinate system name>)
+        # to a transformation object representing the transformation between them
+        self._element_to_cs_mapping: dict[tuple[ELEMENT_TYPE, str], str] = {}
```
Tomaz-Vieira: I'll be sounding like a broken record, but the fact that we use strings for everything here (CS name and Element name) is a bit confusing.  It seems we could just straight up have a

`dict[ELEMENT_TYPE, NgffCoordinateSystem]`

which reads easier and type checks more tightly, since there can never be mixups between element names and CS names. We do have to make sure that all elements have reasonable `__hash__` and `__eq__`, though.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+            A list of tuples representing the transformations associated with the coordinate system.
+        """
+        associated_transformations = []
+        for input_cs, output_cs in self._coordinate_transforms:
+            transformation_key = (input_cs, output_cs)
+            if (input_cs == cs_name or output_cs == cs_name) and transformation_key not in associated_transformations:
+                associated_transformations.append(transformation_key)
+        return associated_transformations
+
+    def _get_elements_associated_with_cs(self, cs_name: str) -> list[tuple[ELEMENT_TYPE, str]]:
```
Tomaz-Vieira: Maybe we need a word that is better than "associated" to mean that the element _belongs_  or _exists in_ the coordinate system. It's a stronger relation than the transforms that are "associated" with the coordinate systems, because an element can only belong so a single CS, but a CS can have many transforms.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+            If there are transformations or elements associated with the coordinate system.
+        """
+        if cs_name not in self._coordinate_systems:
+            raise KeyError(f"Coordinate system with name '{cs_name}' not found.")
+
+        associated_transformations = self._get_transformations_associated_with_cs(cs_name)
+        associated_elements = self._get_elements_associated_with_cs(cs_name)
+
+        # Raise error if there are associated transformations or elements
+        if len(associated_transformations) or len(associated_elements):
```
Tomaz-Vieira: I would raise a different exception (or at least an exception with different text) for each situation.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+            raise ValueError(
```
Tomaz-Vieira: Personally, I like declaring custom exceptions so that users can catch them explicitly, A `ValueError` is very general and doesn't say much about the kind of failure that happened. If a user wants to try removing a CS and fails, it would be nice if they could `except CoordSystemInUse` or `except CoordSystemMissing` to handle the different cases.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+        ------
+        KeyError
+            If the transformation does not exist.
+        """
+        key = (input_cs, output_cs)
+        if key not in self._coordinate_transforms:
+            raise KeyError(f"Transformation from '{input_cs}' to '{output_cs}' not found.")
+        del self._coordinate_transforms[key]
+
+    def build_nx_graph(self) -> Any:  # type: ignore[unresolved-reference]  # noqa: F821
```
Tomaz-Vieira: Maybe we could some shenanigans like

```python
if TYPE_CHECKING:
    import networkx as nx

...

def build_nx_graph(self) -> "nx.DiGraph":
    import networkx as nx # it's unfortunate that we need to import again here, though =/
    ...
```

But I suspect we'd need a hard dependency on this anyway to be compatible with all ways in which ngff can transform.

Tomaz-Vieira (commented):
I left a bunch of mostly typing-related comments, as you might imagine 😆 

Also, this is something we might need to decide as a team, but it's a bit funny that often the docstring is way longer than the code itself, and often there isn't much more to say, like:

```
souce_cs
    The name of the source coordinate system
```

I often feel like the typing signature is more concise and more informative than the docstrings, and that sometimes we could survive just with the headline and the "raises" part.


But anyway, good stuff, it's nice to see we're moving already =)

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: @ajkswamy in this case we don't need to have a tuple of 2 items because the element_name is guaranteed to be unique across the object, and also because the only reason we have element_type is legacy: we choose to put all the points in a folder, all the images in a different folders, etc. and in the past there was no uniqueness across these different folders. But this on-disk structure makes NGFF interop more difficult, because any arbitrary OME-Zarr store with images and labels doesn't contain the 'images/my_image', 'labels/my_labels' folder naming convention. So long story short, we were thinking of making the spatialdata object fully flat on disk, and use only the element name (indepedently of the type) to refer to elements.

LucaMarconato: Curerntly during IO we have helper functions like _get_element_type_from_element_name(), so we need both element_type and element_name, but once we move to a flat storage, element_type would not be needed anymore.

LucaMarconato: For internal variables like this we can do what is most convenient for us (for instance passing the element object instead of the element name), but I would prefer to avoid having the user having ever to instantiate a NgffCoordinateSystem unless they really want to. For most users a str would do the job well. Let's keep this in mind in the context of the user-facing APIs and then make a decision.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: Since these are "private" variables I'm fine with both. Either we construct a `nx` object on the fly each time, or we use this directly for the graph representation. In both cases, the internal variables are used to represent a graph.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: Added the comment also somewhere, writing it also here: since transformation manager is used internally let's do what is most convenient and clearer for us. I agree that using directly the CS objects instead of str is more robust. For user facing API calls we need to think it through, since some users come from single-cell and they have no experience (and maybe no interests) in working with NGFF classes. In those cases, passing the name of the `cs` would lead to the same plots/results than the full CS object.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: I agree, I like belongs for instance, but no hard opinions.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: I agree. Around the codebase we use standard exceptions with different text each time. I will comment on standard exceptions vs custom spatialdata exceptions below (where you talk above that).

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: I agree. In the codebase we use standard exception and with catch them by comparing the string, but a cleaner approach would be to define custom exception. We could use custom exceptions here, open an issue to track that at some point we should use custom exceptions elsewhere, and do the transition gradually. If we are going to do the large refactoring, we need to ensure that we subclass the existing exceptions because we don't want to break already implemented try except blocks from user libraries.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: The reason why we do local import was that we found out using profimp that the import time of nx appeared to be slow. But otherwise, hard dependency is the way. We can actually move again to global import and we rebenchmark the import time after we are done with the refactoring. The biggest culprit is dask.

LucaMarconato:
(to be moved to a separate issue)
We can probably restore networkx as a global import, as you can see from running profimp, dask.dataframe is what is driving the slow import time.

image

Importing networkx is actually 273 ms.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        List all registered coordinate system names, ordered as they were added.
+
+        Returns
+        -------
+        A list of coordinate system names.
+        """
+        return list(self._coordinate_systems.keys())
+
+    def add_element(self, element_type: ELEMENT_TYPE, element_name: str, coordinate_system: str) -> None:
```
LucaMarconato: Reminder: belong vs associated to be made consistent. And also element_type here is redundant.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+        key = (input_cs, output_cs)
+        self._coordinate_transforms[key] = transformation
+
+    def get_existing_transformation(self, input_cs: str, output_cs: str) -> BaseTransformation | None:
+        """
+        Retrieve a transformation defined between coordinate systems.
+
+        Parameters
+        ----------
+        input_cs
```
LucaMarconato: TODO: check if the specs allows multi edges between the coordinate systems. There was some discussion regarding this at the Heidelberg hackathon and I don't remmber the outcome.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Raises
+        ------
+        KeyError:
+            if no transformation exists from input_cs to output_cs.
+        """
+        key = (input_cs, output_cs)
+        return self._coordinate_transforms.get(key)
+
+    def remove_transformation(self, input_cs: str, output_cs: str) -> None:
```
LucaMarconato: if we have a multigraph, we need to change the signature to let the user remove a specific transformation.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+        import networkx as nx
+
+        g = nx.DiGraph()
+        for cs_name in self._coordinate_systems:
+            g.add_node(cs_name)
+        for (input_cs, output_cs), transformation in self._coordinate_transforms.items():
+            g.add_edge(input_cs, output_cs, transformation=transformation)
+        return g
+
+    def get_shortest_transformation_sequence(self, source_cs: str, target_cs: str) -> list[BaseTransformation]:
```
LucaMarconato: We will sometimes have multiple shortest paths between 2 coordinate systems (this may happen even if we don't have a multigraph). What we do in spatialdata is shown here. It's not the cleanest approach, but it could be used also here to get the job done: https://github.com/scverse/spatialdata/blob/eb4fb3d2398cd59298049dcafb03feda839b7ecd/src/spatialdata/transformations/operations.py#L218. Note one can still construct edge cases that fail, the clean approach may be to pass the full list of cs as a path.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+            # nx lazily imported
+        except nx.NetworkXNoPath as nxe:  # type: ignore[name-defined, unresolved-reference]  # noqa: F821
+            raise ValueError(f"No path found from {source_cs} to {target_cs}") from nxe
+
+        transformations = []
+        for i in range(len(path) - 1):
+            transformations.append(g[path[i]][path[i + 1]]["transformation"])
+        return transformations
+
+    def get_all_transformation_sequences(self, source_cs: str, target_cs: str) -> list[list[BaseTransformation]]:
```
LucaMarconato: We could return directly a list[spatialdata.transformation.Sequence]

LucaMarconato: Ideas:
- @ajkswamy suggested to let the user choose the return type list of list vs list of Sequence. This has the con of complicating typing. 

Comments:
- Pro: of returning a list of list: the user can easily extract sub transformations
- Con: if you have a Sequence you can directly use it to get an affine, to transform things. At the cost that you have to call my_sequence.transformations to get the subcomponents (but it's not that bad)

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,342 @@
...
+
+        all_sequences = []
+        for path in paths:
+            sequence = []
+            for i in range(len(path) - 1):
+                sequence.append(g[path[i]][path[i + 1]]["transformation"])
+            all_sequences.append(sequence)
+        return all_sequences
+
+    def __repr__(self) -> str:
```
LucaMarconato, line 627: For later: it would be very nice to let the user be able to plot this. An option could be a hard coded HTML rendering with no additional dependencies, but this comes with extra code to maintain.

LucaMarconato: If the user already import matplotlib (via spatialdata-plot), we could plot via networkx + matplotlib.

src/spatialdata/__init__.py
```diff
@@ -57,6 +57,8 @@
     # _core.query.spatial_query
     "bounding_box_query": "spatialdata._core.query.spatial_query",
     "polygon_query": "spatialdata._core.query.spatial_query",
+    # _core.transformation_manager
+    "TransformationManager": "spatialdata._core.transformation_manager",
```
LucaMarconato: No need for the user to instantiate it. It can be removed.

src/spatialdata/__init__.py
```diff
@@ -115,6 +117,8 @@
     # _core.query.spatial_query
     "bounding_box_query",
     "polygon_query",
+    # _core.transformation_manager
+    "TransformationManager",
```
LucaMarconato: Same here, to be removed.

src/spatialdata/__init__.py
```diff
@@ -208,6 +212,9 @@ def __dir__() -> list[str]:
     # _core.spatialdata
     from spatialdata._core.spatialdata import SpatialData
 
+    # _core.transformation_manager
+    from spatialdata._core.transformation_manager import TransformationManager
```
LucaMarconato: Same here.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I don't see anything in the specification that forbids multiple edges between coordinate systems. Also, `transformnd` uses nx.MultiDiGraph and does not check if an edge already exists between two nodes when adding edges to graph.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: How about "element set in a coordinate system"? the API would be like

`class TransformationManager`:
  - `__ init__`
 - .......
 - ........
- `set_element_in_coordinate_system(....)`
- `unset_element_from_coordinate_system(....)`
- `get _element_coordinate_system (....)   "Get the coordinate system in which an element is set"`  
- `get_elements_set_in_cs(....)`

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Implemented internal `nx.MultiDiGraph`

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Graph nodes are now `NgffCoordinateSystem` and edges contain spatialdata transforms as attributes. In addition `repr` of spatialdata transformations as "edge keys", to be able to distinguish for the case when there are multiple edges between nodes

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: done, we have `self.element_to_cs_mapping: dict[str, NgffCoordinateSystem] = {}`

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Currently, I have tried to stick to "transformations associated with a cs" and "elements belonging to a cs". Only exception is "unset_element" for which I don't have a better idea. Any ideas?

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Done, only custom exceptions and warning used

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Done, only custom exceptions and warning used

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Using global import for now

src/spatialdata/__init__.py
ajkswamy: Done

src/spatialdata/__init__.py
ajkswamy: Done

src/spatialdata/__init__.py
ajkswamy: Done

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Done (see [related comment thread](https://github.com/scverse/spatialdata/pull/1164#discussion_r3621172857))

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: done

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Thanks for the idea Luca. To solve this issue, I have implemented something more in line with what networkx does in such cases, which is to return all shortest paths and let the caller decide (`get_all_shortest_transformation_sequences`). Might be okay as these are internal APIs

For the cases where there are multiple edges between a pair of nodes in a path, then an error is raised by default ("TransformationPathAmbiguousError"). Caller can specify which of the multiple edges is to be selected by passing the corresponding transform into the list argument "expected_intermediate_transformations"

The same logic for disambiguation also applies to `get_all_transformation_sequences`

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: done, returning list[spatialdata.transformation.Sequence]

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        -------
+        The MultiDiGraph containing coordinate systems and transformations.
+
+        Note
+        ----
+        Direct manipulation of the graph is discouraged. Please use the
+        TransformationManager methods whenever possible for better maintainability
+        and to ensure proper validation and error handling.
+        """
+        warnings.warn(
```
Tomaz-Vieira: I think we shouldn't have this method. If someone wants to be daring, they can access the private field, and that it is immediately visible in the client code that they are doing something risky (it is even flagged by `pyright`), whereas this accessor method makes it feel like accessing the graph is fair game. This applies to all other methods that expose a private field

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+from spatialdata.transformations.ngff.ngff_coordinate_system import NgffCoordinateSystem
+from spatialdata.transformations.transformations import BaseTransformation, Sequence
+
+TRANSFORM_KEY = "transformation"
+
+
+class TransformationManager:
+    def __init__(self) -> None:
+        """Initialize a TransformationManager with empty graph and mappings."""
+        self._graph: nx.MultiDiGraph = nx.MultiDiGraph()
```
Tomaz-Vieira: `MultiDiGraph` is generic over the type of node, the node properties and the edge properties, but it doesn't have a default type for the node type. Could you add it here like `nx.MultiDiGraph[NgffCoordinateSystem] ?`

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        warnings.warn(
+            "Direct access to the internal element_to_cs_mapping is discouraged. "
+            "Please use TransformationManager methods whenever possible for better "
+            "maintainability and to ensure proper validation and error handling",
+            InternalAttributeAccessWarning,
+            stacklevel=2,
+        )
+        return self._element_to_cs_mapping
+
+    def check_if_element_exists_else_raise_error(self, element_name: str) -> None:
```
Tomaz-Vieira: I find that this is a lot of control flow encoded into a method name. Something like

```python
def get_element(self, *, name: str) -> SpatialElement | None:
    ...
```

feels simpler; it communicates that the element might not be there and it never raises, so the client can do whatever they feel is appropriate if they don't get an element.

The same applies to the other methods of the in the form `check_..._else_raise`

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+
+        Raises
+        ------
+        CoordinateSystemAlreadyExistsError
+            If the coordinate system already exists.
+        """
+        try:
+            self.check_if_coordinate_system_exists_else_raise_error(cs)
+        except CoordinateSystemNotFoundError:
+            with suppress_direct_internal_attribute_access_warning():
```
Tomaz-Vieira: The proposed changes above also means we don't have to use these warning-suppressing context managers, which I'd argue we should really really really avoid

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+            If the coordinate system has associated elements.
+        CoordinateSystemNotFoundError
+            If the coordinate system is not found
+        """
+        try:
+            self.check_if_coordinate_system_has_no_transformations_else_raise_error(cs)
+            # also checks if cs exists
+            self.check_if_coordinate_system_has_no_elements_else_raise_error(cs)
+        except (CoordinateSystemHasTransformationsError, CoordinateSystemHasElementsError) as err:
+            raise CannotRemoveCoordinateSystemError(cs.name) from err
```
Tomaz-Vieira, line 218: Catching `CoordinateSystemHasTransformationsError` and `CoordinateSystemHasElementsError`, and then raising a `CannotRemoveCoordinateSystemError` erases the error information. In fact, `CannotRemoveCoordinateSystemError` isn't event listed in the `Raises` part of the docstring........ which is another reason I like returning exceptions: the type checker would have caught this :D

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+
+    def list_coordinate_systems(self) -> list[NgffCoordinateSystem]:
+        """
+        List all registered coordinate systems.
+
+        Returns
+        -------
+        A list of coordinate system objects.
+        """
+        with suppress_direct_internal_attribute_access_warning():
```
Tomaz-Vieira: We should just use `self._graph` here and ditch the context manager

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        coordinate_system
+            The coordinate system to which the element belongs.
+
+        Raises
+        ------
+        CoordinateSystemNotFoundError
+            If the coordinate system is not found.
+        """
+        try:
+            self.check_if_element_exists_else_raise_error(element_name)
```
Tomaz-Vieira: Another argument for not doing this amount of control flow encoded in the method name (`check...if..else...raise`) is that it makes this control flow here a bit awkward to follow. We expect the check to fail with an exception, which we catch and do the normal operation, whereas when it _doesn't_ fail with an exception, then we manually raise an exception

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+            If the element has not been registered to any coordinate system.
+        """
+        self.check_if_element_exists_else_raise_error(element_name)
+        with suppress_direct_internal_attribute_access_warning():
+            del self.element_to_cs_mapping[element_name]
+
+    @staticmethod
+    def _get_edge_key_from_transform(transform: BaseTransformation) -> str:
+
+        return repr(transform)
```
Tomaz-Vieira: I think it's a bit dangerous to trust the `__repr__` of `BaseTransform` to be enough to cater to all the needs of the graph. In particular, it's possible that two transformations that are equal but not identical (i.e. `a ==  b and id(a) != id(b)` would produce the same key. Maybe something like `(id(transform), transform)` instead?

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+
+    def add_transformation(
+        self, source_cs: NgffCoordinateSystem, target_cs: NgffCoordinateSystem, transformation: BaseTransformation
```
Tomaz-Vieira: I'd remove the `_cs` suffix from `source_cs` and `target_cs`, since their types already contain that information. Also, I'd make them keyword-only arguments, because they have he same types and this makes it clear which is which.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        """
+        self.check_if_coordinate_system_exists_else_raise_error(source_cs)
+        self.check_if_coordinate_system_exists_else_raise_error(target_cs)
+
+        with suppress_direct_internal_attribute_access_warning():
+            edge_key = self._get_edge_key_from_transform(transformation)
+            edge_attributes = {TRANSFORM_KEY: transformation}
+            self.graph.add_edge(source_cs, target_cs, key=edge_key, **edge_attributes)
+
+    def get_existing_direct_transformations(
```
Tomaz-Vieira: I think we could remove the `existing` from the method name, as it is rather implicit

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+
+        Returns
+        -------
+        List of transformations
+
+        Raises
+        ------
+        CoordinateSystemNotFoundError
+            If either coordinate system does not exist.
+        TransformationNotFoundError
```
Tomaz-Vieira: I suspect returning `None` (or maybe even an empty list?) might be more elegant than raising an exception, as I think it might be useful to tell that two nodes aren't connected as a normal usage of using spatial data, and it'd be nicer to do that without having to catch exceptions

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        self.check_if_any_edge_exists_else_raise_error(source_cs, target_cs)
+        # also checks if source_cs and target_cs exist
+        with suppress_direct_internal_attribute_access_warning():
+            expected_edge_key = self._get_edge_key_from_transform(transformation)
+            assert expected_edge_key in self.graph[source_cs][target_cs], TransformationNotFoundError(
+                source_cs.name, target_cs.name, expected_edge_key
+            )
+            self.graph.remove_edge(source_cs, target_cs, key=expected_edge_key)
+
+    def remove_all_transformations_between_coordinate_systems(
```
Tomaz-Vieira, line 401: maybe rename this method to `remove_all_transformations(self, *, from: ..., to: ...)` ?

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+            The output coordinate system.
+
+        Raises
+        ------
+        CoordinateSystemNotFoundError
+            If either coordinate system does not exist.
+        TransformationNotFoundError
+            If no transformation exists between the coordiante systems
+        """
+        self.check_if_any_edge_exists_else_raise_error(source_cs, target_cs)
```
Tomaz-Vieira: clearing the connections between two unconnected nodes is arguably not an error. Something that we could do is to return the removed connections; If the caller cares about this, they can check if the list was empty (or None or whatever)

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        intermediate_transformation_edge_keys = set()
+        if expected_intermediate_transformations is not None:
+            intermediate_transformation_edge_keys |= {
+                self._get_edge_key_from_transform(it) for it in expected_intermediate_transformations
+            }
+        all_sequences = []
+        deduplicated_paths = list({repr(x): x for x in paths}.values())
+        for path in deduplicated_paths:
+            transformation_list = []
+            for i in range(len(path) - 1):
```
Tomaz-Vieira, line 474: what if `path` has only one node?

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        CoordinateSystemNotFoundError
+            If either coordinate system does not exist.
+        """
+        self.check_if_coordinate_system_exists_else_raise_error(source_cs)
+        self.check_if_coordinate_system_exists_else_raise_error(target_cs)
+
+        with suppress_direct_internal_attribute_access_warning():
+            edge_key = self._get_edge_key_from_transform(transformation)
+            edge_attributes = {TRANSFORM_KEY: transformation}
+            self.graph.add_edge(source_cs, target_cs, key=edge_key, **edge_attributes)
```
Tomaz-Vieira: Can't we just put all the edge information in `key`? At least we can keep track of the key type, by setting it in the generic type params of the graph. If we use edge attributes we completely lose the type information.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+            assert len(self.graph[source_cs][target_cs]), TransformationNotFoundError(source_cs.name, target_cs.name)
+            for edge_key in list(self.graph[source_cs][target_cs].keys()):
+                # need to covert keys() to list to freeze it, else it will change during the following removal
+                self.graph.remove_edge(source_cs, target_cs, key=edge_key)
+
+    def _get_transformation_sequences_from_path_after_disambiguation(
+        self,
+        paths: list[list[NgffCoordinateSystem]],
+        expected_intermediate_transformations: list[BaseTransformation] | None,
+    ) -> list[Sequence]:
```
Tomaz-Vieira: missing the generic type parameter for `Sequence`

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+            }
+        all_sequences = []
+        deduplicated_paths = list({repr(x): x for x in paths}.values())
+        for path in deduplicated_paths:
+            transformation_list = []
+            for i in range(len(path) - 1):
+                edge_data = self.graph[path[i]][path[i + 1]]
+                if len(edge_data) > 1:
+                    # when there are multiple edges between a pair of coordinate systems in the path
+                    intermediate_transformation_key_here = intermediate_transformation_edge_keys & set(edge_data.keys())
```
Tomaz-Vieira: I suspect that the fact that we don't consume the transform in `intermediate_transformation_edge_keys` could cause some errors here, because candidate edges could be matched against the same expected intermediate edge multiple times

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+                    if len(intermediate_transformation_key_here) == 0:
+                        # transformation was not specified in `intermediate_transformations` for disambiguation
+                        raise TransformationPathAmbiguousError(path[i].name, path[i + 1].name)
+
+                    edge_key_to_use = list(intermediate_transformation_key_here)[0]
```
Tomaz-Vieira: isn't it conceivable that `len(intermediate_transformation_key_here) > 1` ?

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+                    edge_key = next(iter(edge_data.keys()))
+                    transformation_list.append(edge_data[edge_key][TRANSFORM_KEY])
+            all_sequences.append(Sequence(transformation_list))
+        return all_sequences
+
+    def get_all_shortest_transformation_sequences(
+        self,
+        source_cs: NgffCoordinateSystem,
+        target_cs: NgffCoordinateSystem,
+        expected_intermediate_transformations: list[BaseTransformation] | None = None,
```
Tomaz-Vieira: consider `expected_intermediate_transformations: Sequence[BaseTransform] = ()`

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+                )
+            except TransformationPathAmbiguousError as tpae:
+                raise tpae from TransformationPathAmbiguousError(source_cs.name, target_cs.name)
+
+    def get_all_transformation_sequences(
+        self,
+        source_cs: NgffCoordinateSystem,
+        target_cs: NgffCoordinateSystem,
+        expected_intermediate_transformations: list[BaseTransformation] | None = None,
+    ) -> list[Sequence]:
```
Tomaz-Vieira: missing generic type argument

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        if expected_intermediate_transformations is not None:
+            intermediate_transformation_edge_keys |= {
+                self._get_edge_key_from_transform(it) for it in expected_intermediate_transformations
+            }
+        all_sequences = []
+        deduplicated_paths = list({repr(x): x for x in paths}.values())
+        for path in deduplicated_paths:
+            transformation_list = []
+            for i in range(len(path) - 1):
+                edge_data = self.graph[path[i]][path[i + 1]]
```
Tomaz-Vieira: I suspect a lot of this logic could be simplified by using `all_simple_edge_paths`, specially if we store the edge information in the edge key rather than on the edge attributes

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+                raise TransformationPathNotFoundError(source_cs.name, target_cs.name) from nxe
+
+            try:
+                return self._get_transformation_sequences_from_path_after_disambiguation(
+                    paths, expected_intermediate_transformations
+                )
+            except TransformationPathAmbiguousError as tpae:
+                raise tpae from TransformationPathAmbiguousError(source_cs.name, target_cs.name)
+
+    def _get_transformations_associated_with_cs(self, cs: NgffCoordinateSystem) -> list[BaseTransformation]:
```
Tomaz-Vieira: consider renaming to `_get_transformations(self, *, cs: NgffCoordinateSystem)`

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+        """
+        Get all transformations associated with a coordinate system.
```
Tomaz-Vieira: semantically this feels a bit weird because the returned transformations are both incoming and outgoing. Also, it seems this is only used to check whether a CS can be removed, and in that case we don't even need to go through the trouble of collecting all transforms; simply knowing that one exists would suffice.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,662 @@
...
+                transformations_outgoing = self.get_existing_direct_transformations(cs, successor)
+                transformations += transformations_outgoing
+            # Check incoming edges (other -> cs)
+            for predecessor in self.graph.predecessors(cs):
+                transformations_incoming = self.get_existing_direct_transformations(source_cs=predecessor, target_cs=cs)
+                transformations += transformations_incoming
+
+            return transformations
+
+    def _get_elements_belonging_to_cs(self, cs: NgffCoordinateSystem) -> list[str]:
```
Tomaz-Vieira, line 599: I would rather get the elements themselves rather than their names.

Also, have you considered keeping the element literally inside the nodes?

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -1,342 +1,662 @@
...
+import networkx as nx
+
+from spatialdata._core.transformation_manager.exceptions import (
+    CannotRemoveCoordinateSystemError,
+    CoordinateSystemAlreadyExistsError,
+    CoordinateSystemHasElementsError,
+    CoordinateSystemHasTransformationsError,
+    CoordinateSystemNotFoundError,
+    ElementAlreadyExistsError,
+    ElementNotFoundError,
```
LucaMarconato: In the case in which there are other parts of the code where we get an exception due to a "element not found", we can handle this in two ways:
- we reuse the same ElementNotFoundError exception and differentiate the two exception via the message string. This means that we would move the exception to a central file, outside `transformation_manager`. 
- we allow to have a different exception class named `ElementNotFoundError`, defined in a different path. The user would then handle the `spatialdata._core.transformation_manager.exceptions.ElementNotFoundError` and `spatialdata._core.somewhere_else.ElementNotFoundError` explicitly. 

The second approach may be cleaner (but also the first looks cleaner but hides the complexity in the string). Tagging @Tomaz-Vieira for comments.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: Read access is fine, but the problem is that there the object is mutable, and editing would mess things up.

We could maybe return a deep copy (it's cheap anyway)? Feels a bit hacky though unless we call it something like `graph_readonly()`?

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -1,342 +1,662 @@
...
+        Returns
+        -------
+        A dictionary mapping element names to their coordinate systems.
+
+        Note
+        ----
+        Direct manipulation of internal element_to_cs_mapping is discouraged. Please use the
+        TransformationManager methods whenever possible for better maintainability
+        and to ensure proper validation and error handling.
+        """
```
LucaMarconato: Same problem than with the graph. Read is fine, but editing could mess things up. But here I'd just remove the property and keep the private method.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: Or we just keep the private graph.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: Good catch!

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: `(id(transform), transform)` should do the trick.

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -1,342 +1,662 @@
...
+            If the element does not exist.
         """
-        mapping_key = (element_type, element_name)
-        return self._element_to_cs_mapping.get(mapping_key)
+        self.check_if_element_exists_else_raise_error(element_name)
+        with suppress_direct_internal_attribute_access_warning():
+            return self.element_to_cs_mapping[element_name]
 
-    def add_transformation(self, input_cs: str, output_cs: str, transformation: BaseTransformation) -> None:
+    def unset_element(self, element_name: str) -> None:
```
LucaMarconato: We need to think if we want to be able to have an element in no coordinate system (as opposed to a coordinate system for the element itself). We can keep it like this and iterate, but let's keep in mind that trying to write an element that belongs to no coordinate system would probably not follow the specs.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: This is the `Sequence` transformation, not the `collection.abc.Sequence`.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: I'd go for `list` because:
- `transformations.Sequence` is not a parametrizable type
- `transformations.Sequence` would be a new object that does not exist anywhere in the graph, while `list[BaseTransform]` will list objects that are existing somewhere.

The distinction between transforms that exist and that are created anew is important in the case in which we check intermediate transformations with `id(transformation)` instead of `== transformation`.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: `transformation.Sequence` is not a parametrizable type.

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: not specifiable for `transformations.Sequence`

src/spatialdata/_core/transformation_manager/__init__.py
LucaMarconato: The user could replace an element (we do it sometimes in the codebase), leading to a change in `id`. So here having the element in the graph would be risky.

This would be especially common when creating a `SpatialData` object in memory, and where the in-memory object may change. For example you could start by loading an image from a TIFF, but at some point writing that image (to a previously written SpatialData object) and reloading only that image from disk. The `id` would change.

src/spatialdata/transformations/transformations.py
```diff
@@ -509,7 +509,7 @@ def __eq__(self, other: Any) -> bool:
 class Affine(BaseTransformation):
     def __init__(
         self,
-        matrix: list[Number] | ArrayLike,
+        matrix: list[list[Number]] | ArrayLike,
```
LucaMarconato, line 512: Correct. The specs before was just a list of numbers, now it's a list of lists.

LucaMarconato: https://github.com/ome/ngff/issues/146#issuecomment-1270117494

src/spatialdata/_core/transformation_manager/__init__.py
Tomaz-Vieira: Ah, I was wrong everywhere else, thinking that `transformations.Sequence` was the builtin `collections.Sequence`, but in this case I actually meant `collections.Sequence`, because it does take a type argument (just like `list`) and it allows for `expected_intermediate_transformations` to always be a `collections.Sequence[BaseTransformation]` instead of an union (`list[...] | None`).

That is, unless there is different semantics to the empty `list` VS `None`; If empty means "I will tolerate nothing" and `None` means "I don't care", then it should still be the union. If there is no distinction, then it could be a `collections.Sequence[BaseTransformation]` defaulting to `()`, which is immutable and therefore safe to use as a default argument.

> The distinction between transforms that exist and that are created anew is important

That does sound a bit similar to an issue in `transformnd`; It might be that we need two different concepts: one is just the mathematical object that transforms vectors, and another is an edge on the graph. The edge on the graph could either have an identity, which is what would go into query methods like this one, or it could be aware of the Coordinate Systems (and their IDs, probably), so that they could be safely compared to other edges on a path. I'm sure an `Identity` transform would otherwise match a whole bunch of unintended edges 🤔

src/spatialdata/_core/transformation_manager/__init__.py
Tomaz-Vieira: > differentiate the two exception via the message string

In haven't thought a lot about this, but I suppose that we don't need all exceptions to be absolutely unique, because the caller already has some context of what is going on, since they are calling the function to begin with, and they should handle the `ElementNotFound` differently depending on what they were trying to do.

On the other hand, were you thinking about some case in which a function call can produce an `ElementNotFoundError` for two different failure modes?

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I agree on keeping the graph private for now as prroviding a read-only copy will need a lot of costly deep copying. It would be better in the future to make the transformation classes immutable (for example using frozen dataclasses) and use [nx.generic_graph_view](https://networkx.org/documentation/stable/reference/classes/generated/networkx.classes.graphviews.generic_graph_view.html#generic-graph-view) to return a read-only view of the graph. This exposes attributes, but with our transformation classes immutable, it won't be an issue.

Made the graph private.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Done

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: These methods are just helper functions that assert pre-requisites. They make code in other methods simpler and avoid repeating code. I have renamed them to directly convey what they assert.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, done

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I've updated the documentation to clearly define the errors raised and their chaining. 

However, error information is not lost as chained error messaged gets printed:

```
except (CoordinateSystemHasTransformationsError, CoordinateSystemHasElementsError) as err:
            raise CannotRemoveCoordinateSystemError(cs.name) from err
```

```
  File "..../spatialdata/src/spatialdata/_core/transformation_manager/__init__.py", line 199, in remove_coordinate_system
    self.assert_coordinate_system_has_no_transformations(cs)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^
  File "..../spatialdata/src/spatialdata/_core/transformation_manager/__init__.py", line 139, in assert_coordinate_system_has_no_transformations
    raise CoordinateSystemHasTransformationsError(cs.name)
spatialdata._core.transformation_manager.exceptions.CoordinateSystemHasTransformationsError: Coordinate System ('cs1') has transformations.

The above exception was the direct cause of the following exception:

  File "..../spatialdata/tmp/example_error_chaining.py", line 44, in <module>
    main()
    ~~~~^^
  File "..../spatialdata/tmp/example_error_chaining.py", line 38, in main
    tm.remove_coordinate_system(cs1)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "..../spatialdata/src/spatialdata/_core/transformation_manager/__init__.py", line 203, in remove_coordinate_system
    raise CannotRemoveCoordinateSystemError(cs.name) from err
spatialdata._core.transformation_manager.exceptions.CannotRemoveCoordinateSystemError: Cannot remove coordinate system with name cs1.

```

Programmatically, this can be handled as follows:

```
try:
    tm.remove_coordinate_system(cs1)
except CannotRemoveCoordinateSystemError as e:
    # Check if it was raised from a specific error
    if isinstance(e.__cause__, CoordinateSystemHasTransformationsError):
        pass # do something
    elif isinstance(e.__cause__, CoordinateSystemHasElementsError):
        pass # do something else
    else:
        pass # so something else still
```

Here is the full script for reference: [example_error_chaining.py](https://github.com/user-attachments/files/30583921/example_error_chaining.py)

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, done

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree that this pattern is awkward. Have added and used a new method to improve it:

`assert_element_does_not_exist`

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Considering all other comments here, I have defined edge key to be 

`f"{source_cs.name}_{target_cs.name}_{id(transform)}"`.

This makes the key unique across the graph (for the edge case that someone tries to add the same transform between the same coordinate systems repeatedly, it just gets silently overwritten, which I think is okay)

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I would rather keep it to make usage better for a user writing without type hints:

```
tm.add_transformation(source_cs=..., target_cs=...., transformation=....)

vs 

tm.add_trasnsformation(source=...., target=...., transformation=....)
```

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, done.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, done

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I see your point. I've made the method private. I think we would only need to use this private method when removing an element from a SpatialData object.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: [Docs](https://networkx.org/documentation/stable/reference/classes/generated/networkx.MultiDiGraph.add_edge.html#networkx.MultiDiGraph.add_edge) say:

```
key
   hashable identifier, optional (default=lowest unused integer)
   Used to distinguish multiedges between a pair of nodes.
```

Our edge information is not hashable currently, and even if update them to be hashable, putting it in the `key` would be a hack and could lead to unexpected behaviour due to future updates of `networkx`

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, returning an empty list.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I think `remove_all_transformation(...)` might be interpreted as removing all transformations of the graph

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, not raising an error and silently doing nothing in this case. I think we don't need to return removed edges for now, we could add this feature if we need it later.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, throwing error for this case

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I see the simplification coming from using `all_simple_edge_paths`. However, `networkx` algorithms for finding shortest paths only return paths and hence we would need this complexity for that case.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I have updated the API to require passing edge definitions to `expected_intermediate_edges` (see code below). In addition, I have also asserted that we expected to disambiguate only simple paths in this method. These facts along with the fact that we encode this edge definition into the edge key ensure that we avoid this issue.

```
for path in paths:
            assert len(set(path)) == len(path), TransformationPathNotSimple(path)
```


```
def get_all_shortest_transformation_sequences(
        self,
        source_cs: NgffCoordinateSystem,
        target_cs: NgffCoordinateSystem,
        expected_intermediate_edges: Sequence[EDGE_DEF] = (),
    ) -> list[sd_transforms.Sequence]:
        """
        Get all shortest sequences of transformations between two coordinate systems.

        Parameters
        ----------
        source_cs
            The source coordinate system.
        target_cs
            The target coordinate system.
        expected_intermediate_edges
            list of intermediate edges expected.
            An edge is defined as a tuple, (source_cs, target_cs, transform).
            Used to choose an edge when multiple edges are found between the same coordinate systems.
....
....
```

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, raising a new custom error for this case

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: I've replaced the type definition to be `expected_intermediate_edges: Sequence[EDGE_DEF] = ()`, where `EDGE_DEF = tuple[NgffCoordinateSystem, NgffCoordinateSystem, sd_transforms.BaseTransformation]`. I have also encoded `these` edge defining tupes into edge keys as `f"{source_cs.name}_{target_cs.name}_{id(transform)}"`. All these together must solve the issues mentioned.

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: the method has been Removed, based on your comment below

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, refactored to just check if there are incoming or outgoing edges

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Agree, another point, I would like to keep TransformationManager independent of other attributes and containers within SpatialData.

src/spatialdata/_core/transformation_manager/__init__.py
Tomaz-Vieira: Maybe "erases the information" is a bit too strong, but I'd still argue that we'd be expecting too much of the clients of this method if we require them to go chasing for exceptions inside `__cause__`. The main point is that the docstring says `raises X` but one cannot `except X`. Strictly speaking, this only raises `CannotRemoveCoordinateSystemError`, even if that exception happens to have some particular things inside `__cause__`.

And if we're fine requiring that users call `isinstance(e, SomeException)`, then we might as well go all in and return the exceptions, so they show up in the method signature and get type checker support 🙃

src/spatialdata/_core/transformation_manager/__init__.py
```diff
@@ -0,0 +1,623 @@
...
+    TransformationPathAmbiguousMultipleEdgeExpectedError,
+    TransformationPathAmbiguousNoEdgeExpectedError,
+    TransformationPathNotFoundError,
+    TransformationPathNotSimple,
+)
+from spatialdata.transformations import transformations as sd_transforms
+from spatialdata.transformations.ngff.ngff_coordinate_system import NgffCoordinateSystem
+
+TRANSFORM_KEY = "transformation"
+EDGE_DEF = tuple[NgffCoordinateSystem, NgffCoordinateSystem, sd_transforms.BaseTransformation]
```
Tomaz-Vieira: depending on the python version we want to support, this should probably either be

`EDGE_DEF: TypeAlias = ...` # up to 3.11

or 

`type EDGE_DEF = ...` # from 3.12 onwards

src/spatialdata/_core/transformation_manager/__init__.py
Tomaz-Vieira: even with the keuyword-only params?

src/spatialdata/_core/transformation_manager/__init__.py
ajkswamy: Thanks, 
- I've corrected the syntax here to use `type EDGE_DEF = `
- `spatialdata` [no longer supports python v3.11](https://github.com/scverse/spatialdata/pull/1151), [since scverse follows SPEC-0](https://github.com/scverse/spatialdata/issues/1153)

Issue #1152, opened 2026-06-22
Missing values in AnnData `.uns` DataFrame become string `"nan"` after SpatialData Zarr round trip


AlbertoFabbri93:
## Description

When writing a `SpatialData` object to Zarr and reading it back with `sd.read_zarr`, missing values in a pandas `DataFrame` stored inside an AnnData table’s `.uns` are converted from real missing values (`NaN` / `pd.NA`) into the literal string `"nan"`.

This changes the semantics of the metadata after a write/read round trip. In downstream code, the string `"nan"` is then treated as a real category rather than a missing value.

## Reproducible check from my object

```python
col = "TNBCtype4_n235_notPreCentered"

before = sdata_filtered.tables["table"].uns["patients"][col]

print("BEFORE WRITE")
print(before.dtype)
print(before.value_counts(dropna=False))
print("string nan:", (before == "nan").sum())
print("real NA:", before.isna().sum())

sdata_filtered.write(config.DATA_DIR / "TNBC_filtered2.zarr")

sdata_reloaded = sd.read_zarr(config.DATA_DIR / "TNBC_filtered2.zarr")
after = sdata_reloaded.tables["table"].uns["patients"][col]

print("\nAFTER READ")
print(after.dtype)
print(after.value_counts(dropna=False))
print("string nan:", (after == "nan").sum())
print("real NA:", after.isna().sum())
```

## Observed output

```text
BEFORE WRITE
object
TNBCtype4_n235_notPreCentered
BL1    39
BL2    25
NaN    24
LAR    22
M      20
Name: count, dtype: int64
string nan: 0
real NA: 24

AFTER READ
object
TNBCtype4_n235_notPreCentered
BL1    39
BL2    25
nan    24
LAR    22
M      20
Name: count, dtype: int64
string nan: 24
real NA: 0
```

## Expected behavior

The missing values should round-trip as missing values.

Expected after reading:

```text
string nan: 0
real NA: 24
```

Instead, all 24 real missing values are converted to the literal string `"nan"`.

## Context

The affected object is an AnnData table stored inside a SpatialData object:

```python
sdata_filtered.tables["table"].uns["patients"]
```

The `patients` entry is a pandas `DataFrame` containing patient-level metadata. It is generated before writing and stored in `.uns`.

My script also sets:

```python
ad.settings.allow_write_nullable_strings = True
```

but the affected column has dtype `object` both before and after the round trip.

## Question

[...]

PR #1170, opened 2026-07-31
AnnData obs/var string to categoricals

ajkswamy:
# Issue:

scverse/spatialdata#1152

# Summary:

* Writing AnnData table to zarr file now uses different methods depending on zarr format being written and anndata version installed
* A new custom Deprecation warning is thrown when writing any element to zarr format v2.
* A new hatch env config to help run all tests for combination of anndata/pandas versions

# How to test:

* Use [this notebook](<https://github.com/user-attachments/files/30595657/NaNs_spatialdata_checking_obs_and_uns.ipynb>), modified from the one provided by [LazDaria](<https://github.com/LazDaria>) in scverse/spatialdata#1152 to cover all cases there
* Use the hatch env added here and call `hatch run test-anndata-pandas: test {args}`. Check config before running, every script is run multiple times based on the matrix defined.

# Related PRs

* https://github.com/scverse/spatialdata/pull/1106#issuecomment-4876580849

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -266,8 +266,16 @@ anndata-pandas = [
 [tool.hatch.envs.test-anndata-pandas.overrides]
 matrix.anndata-pandas.extra-dependencies = [
     # every option where the if-condition is True gets included
+
+    # anndata 0.13
     {value="anndata~=0.13", if = ["0.13-2", "0.13-3"]},
+
+    # anndata 0.12
     {value="anndata>=0.12,<0.13", if = ["0.12-2"]},
+
+    # pandas 2
     {value="pandas>=2.3,<3", if = ["0.13-2", "0.12-2"]},
+
+    # pandas 3
     {value="pandas~=3.0", if = ["0.13-3"]},
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -253,17 +253,20 @@ dependency-groups = ["test"]
 [tool.hatch.envs.test-anndata-pandas]
 template = "test"
 extra-dependencies = ["zarr>=3"]
+scripts.test = ["pip list|grep anndata && pip list|grep pandas && pytest {args}"]
 scripts.test-readwrite = ["pip list|grep anndata && pip list|grep pandas && pytest tests/io/test_readwrite.py"]
 scripts.test-all = ["pip list|grep anndata && pip list|grep pandas && pytest ."]
 
 [[tool.hatch.envs.test-anndata-pandas.matrix]]
-anndata-pandas = ["0.13-2", "0.13-3"]
+anndata-pandas = ["0.13-2", "0.13-3", "0.12-2"]
 
 [tool.hatch.envs.test-anndata-pandas.overrides]
 matrix.anndata-pandas.extra-dependencies = [
-    # every option when if is True gets included
+    # every option where the if-condition is True gets included
     {value="anndata~=0.13", if = ["0.13-2"]},
     {value="pandas>=2.3,<3", if = ["0.13-2"]},
     {value="anndata~=0.13", if = ["0.13-3"]},
     {value="pandas~=3.0", if = ["0.13-3"]},
+    {value="anndata>=0.12,<0.13", if = ["0.12-2"]},
+    {value="pandas>=2.3,<3", if = ["0.12-2"]},
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ extra  = [
 [dependency-groups]
 dev = [
     "bump2version",
+    "prek>=0.4.11",
 ]
 test = [
     "pytest",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,6 +65,7 @@ extra  = [
 dev = [
     "bump2version",
     "prek>=0.4.11",
+    "ruff>=0.16.0",
 ]
 test = [
     "pytest",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,8 +64,6 @@ extra  = [
 [dependency-groups]
 dev = [
     "bump2version",
-    "prek>=0.4.11",
-    "ruff>=0.16.0",
 ]
 test = [
     "pytest",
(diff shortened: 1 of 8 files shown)


src/spatialdata/_io/io_table.py
```diff
@@ -56,16 +60,43 @@ def write_table(
...
-    tables_group.attrs["region"] = region
-    tables_group.attrs["region_key"] = region_key
-    tables_group.attrs["instance_key"] = instance_key
-    tables_group.attrs["version"] = element_format.spatialdata_format_version
+    # Ensure the table group exists
+    table_group = group.require_group(name=name)
+
+    assert element_format in TablesFormats.values(), FormatVersionUnknownError(
+        element_type="table", version_encountered=element_format
+    )
```
LucaMarconato: This assertion gets silenced when running code with `python -O`. An `if` `raise` is preferable. Please change.

LucaMarconato: also here there is no `raise`, so even in the assertion is evaluated we would get an `AssertionError`, not ` FormatVersioUnknownError`

src/spatialdata/_io/io_table.py
```diff
@@ -56,16 +60,43 @@ def write_table(
...
+        # solution of passing resolved store directly roughly based on:
+        # https://github.com/scverse/anndata/issues/1548#issuecomment-2199801855
+
+        # resolve the store from the group
+        resolved_store = _resolve_zarr_store(table_group)
+
+        # Write the table to the path of the table group
+        table.write_zarr(store=resolved_store, consolidate_metadata=False)
+
+        table_group = group[name]
```
LucaMarconato: do we need to redefine `table_group`? If yes please explain in a comment why. If not please remove.

src/spatialdata/_io/io_table.py
```diff
@@ -56,16 +60,43 @@ def write_table(
...
+    else:
+        table.strings_to_categoricals()
```
LucaMarconato: Here there is a problem: `table.strings_to_categoricals()` changes the dtype of the string column to a categorical. An in-place operation from a write operation is an unexpected side effect.

We need to find a different solution. I suggest to add an extra assert to `test_sdata_with_nan_in_obs()` to verify that the `dtype` of the obs columns of the input object does not change after write, and to run this against all the combinations of pandas and anndata via the hatch config you created.

pyproject.toml
```diff
@@ -64,6 +64,8 @@ extra  = [
 [dependency-groups]
 dev = [
     "bump2version",
+    "prek>=0.4.11",
+    "ruff>=0.16.0",
```
LucaMarconato: Not something to do now, but after moving to the scverse cookiecutter these dev packages may change. We'll figure this out later.

pyproject.toml
```diff
@@ -244,3 +246,27 @@ memray-flame = "memray flamegraph --temporal"
...
+
+[tool.hatch.envs.test-anndata-pandas.overrides]
+matrix.anndata-pandas.extra-dependencies = [
+    # every option where the if-condition is True gets included
+    {value="anndata~=0.13", if = ["0.13-2"]},
+    {value="pandas>=2.3,<3", if = ["0.13-2"]},
+    {value="anndata~=0.13", if = ["0.13-3"]},
+    {value="pandas~=3.0", if = ["0.13-3"]},
+    {value="anndata>=0.12,<0.13", if = ["0.12-2"]},
+    {value="pandas>=2.3,<3", if = ["0.12-2"]},
```
LucaMarconato: why isn't this included in the tests `anndata>=0.12,<0.13` with `pandas~=3.0`? I would either include or explain why it's not in a comment.

LucaMarconato: Also instead of the comment `# every option where the if-condition is True gets included`

I would consider adding these comments, to make the 3 cases immediate. But up to you.

```
# case 1
    {value="anndata~=0.13", if = ["0.13-2"]},
    {value="pandas>=2.3,<3", if = ["0.13-2"]},
# case 2
    {value="anndata~=0.13", if = ["0.13-3"]},
    {value="pandas~=3.0", if = ["0.13-3"]},
# case 3
    {value="anndata>=0.12,<0.13", if = ["0.12-2"]},
    {value="pandas>=2.3,<3", if = ["0.12-2"]},
# case 4: anndata xxx pandas yyy not added because zzz (or add missing case)
```

src/spatialdata/_io/io_table.py
```diff
@@ -56,16 +60,43 @@ def write_table(
...
-    tables_group.attrs["instance_key"] = instance_key
-    tables_group.attrs["version"] = element_format.spatialdata_format_version
+    # Ensure the table group exists
+    table_group = group.require_group(name=name)
+
+    assert element_format in TablesFormats.values(), FormatVersionUnknownError(
+        element_type="table", version_encountered=element_format
+    )
+
+    if element_format.zarr_format == 3 and version("anndata") >= "0.13":
```
LucaMarconato: The second boolean condition performs a string comparison that is evaluated lexicografically, not parsing the version numbers:

```
version("anndata") >= "0.2.0"
Out[5]: False
version("anndata")
Out[6]: '0.12.7'
```

LucaMarconato: To fix you can wrap each string with `packaging.version.Version()`

LucaMarconato:
One last thing, please rename the PR to a more descriptive text since upon release the automatically-generated release notes will use the PR title.

src/spatialdata/_io/io_table.py
ajkswamy: Agree, done

src/spatialdata/_io/io_table.py
ajkswamy: you are right, I have removed them

pyproject.toml
ajkswamy: Done

src/spatialdata/_io/io_table.py
ajkswamy: Right, done

pyproject.toml
ajkswamy: I've added a comment why the case of `anndata>=0.12,<0.13` with `pandas~=3.0` .

I have simplified the conditionals defining the dependency versions. For now, I prefer to keep them organized per dependency and version. We can modify them if needed later.

src/spatialdata/_io/io_table.py
ajkswamy: Yes, right, I added the extra assert statements to `test_sdata_with_nan_in_obs()` and indeed there is a side effect caused by `strings_to_categoricals()`.

`anndata.AnnData.write_zarr` has an argument `convert_strings_to_categorical` which is set to True by default. Hence the default round trip for nullable stings columns works by default with `anndata`, but with this side effect. Here, I have added an argument to `io_table.write_table`  along with all functions in the call chain from `Spatialdata.write` that allows users the same options to convert strings in AnnData objects to categorical before writing. The main difference that we set the default value to be false, i.e., we do not convert by default and users need to explicitly set this to be able to do round trip with nullable string/object columns with pandas 2.0.

LucaMarconato: Thanks for explaining! So if I understood correctly the side effect happens already with `anndata`? If so could you please report to the `anndata` team (GitHub issue) with a minimal reproducible example (only using `anndata`, not `spatialdata`)? The bug to show is that a write operation changes the input in-memory object silently.

ajkswamy: I've created an issue in anndata [here](https://github.com/scverse/anndata/issues/2610)

PR #1180, opened 2026-08-18
docs: add interoperability page

jan-glx:
Closes scverse/spatialdata#1176. Preview: [https://scverse-spatialdata--1180.org.readthedocs.build/en/1180/interoperability.html](<https://scverse-spatialdata--1180.org.readthedocs.build/en/1180/interoperability.html>)

Adds a short `Interoperability` page listing the non-Python interfaces to the SpatialData on-disk format, following the structure of the [equivalent page in mudata](<https://github.com/scverse/mudata/commit/f97ad24e981cb2e5e5149c511073195c4771a05d>) and the one in [anndata](<https://github.com/scverse/anndata/blob/main/docs/interoperability.md>), as requested by @Zethson so the style is consistent across the three data structures.

Sections: **R** ([spatialdataR](<https://github.com/HelenaLC/spatialdataR>)), **JavaScript and TypeScript** ([SpatialData.js](<https://github.com/Taylor-CCB-Group/SpatialData.js>), [Vitessce](<https://vitessce.io/docs/data-file-types/#spatialdatazarr>)), and a short **File format** note (to be expanded into its osn tab later).

`napari-spatialdata` is not listed as it's python, relies on the spatialdata python package and does not read on-disk data directly

PR #1181, opened 2026-08-19
Fix numba JIT disable patch targeting wrong config attribute

Tomaz-Vieira:
Test suite tried to force-disable numba JIT when already imported by patching `NUMBA_DISABLE_JIT` on numba.core.config, but numba internally reads the env var into an attribute named `DISABLE_JIT` (no prefix).

We should find a less hacky solution to this, but for now, running the tests sequentially (`pytest -x`) always pass.


Tomaz-Vieira:
`pre-commit` is flagging a `mypy` error that has nothing to do with this PR, so I don't think we should pollute it with a fix

LucaMarconato:
Thanks for the fix, well spotted!

> pre-commit is flagging a mypy error that has nothing to do with this PR, so I don't think we should pollute it with a fix

Yes, clean solution is to fix it in another PRs, but practically those pre-commit fail happen routinely so we usually just address them at the first occasion (or sometimes merge to main with red CI and fix it later, if we know that we are going to fix it soon anyway). I fixed it here right before merging: https://github.com/scverse/spatialdata/pull/1180.

PR #1182, opened 2026-08-19
NGFF v06 transformations

Tomaz-Vieira:
Mirrors Ngff* classes in the `graph` module, parses ome-zarr-models-py classes.

# `CoordSystem` (based on `NgffCoordinateSystem`)

- Made immutable, so we can `__hash__` (needed for usage in graph, but we might still have the identity issue when searching for verts)
- Can be marked "virtual" so that we know not to serialize it (that is, it has been created for the sake of putting it in a graph and maintaining the invariant of always having an input and output, but it should not show up in the zarr metadata)

# `*Edge` (based on `NgffTransformation`)

- All transformations always have input and output, which diverge from the raw NGFF
- Don't do serialization to raw JSON anymore, but rather convert to and from `omz-models-py` classes
- Stronger invariants; checked in `__init__` and nowhere else
- A few corner cases fixed

# io_raster.py

- Adds function to read multiscales from v06, applying the scale/translation transformations as a `RangeIndex` to the resulting xarrays

# Next steps

## Roundtripping

This PR only does input dfor now; The full roundtrip is showing some issues that would delay this further, so I've removed all output for now and will submit it in another PR

## Using unorderd axes in coordinate systems

Since our arrays are labeled, we probably shouldn't be using a `Sequence` of axes in our coordinate systems. This will probably simplify runtime quit a lot, but might create some more work during IO


pyproject.toml
```diff
@@ -35,7 +35,8 @@ dependencies = [
     "networkx",
     "numba>=0.55.0",
     "numpy",
-    "ome_zarr>=0.16.0",
+    # "ome_zarr>=0.16.0",
+    "ome_zarr @ git+https://github.com/ome/ome-zarr-py/@e859b55425b740335309876bb023fe52977a3045",
```
LucaMarconato: this commit is ~40 commits old wrt main; I'd update already.

src/spatialdata/_io/io_raster.py
```diff
@@ -160,6 +166,83 @@ def _prepare_storage_options(
     return prepared_options
 
 
+def try_read_ngff06_multiscale(store: Path) -> tuple[DataTree, Sequence[BaseTransfEdge]]:
+    multiscale = oz.OMEZarrMultiscale.from_ome_zarr(str(store))
+    assert isinstance(multiscale, oz.OMEZarrMultiscale)  # disambiguate from OMEZarrLabel
+
+    name_to_cs: dict[str, CoordSystem] = {}
```
LucaMarconato, line 173: Reminder: later when we work with `Scene`s, the coordinate system name is not enough for uniquely identifying a CS. It will be the combniation of path where the cs is defined, and the name.

src/spatialdata/_io/io_raster.py
```diff
@@ -160,6 +166,83 @@ def _prepare_storage_options(
...
+        transformed_stop = seq.transform_points((ds_shape - 1)[np.newaxis, :])[0]
+
+        coords = xr.Coordinates()
+        for low, high, ax, extent in zip(transformed_start, transformed_stop, out_cs.axes, ds_shape, strict=True):
+            if ax.type == "channel" and channel_names is not None:
+                coords.merge({ax.name: channel_names})
+                continue
+            coords = coords.merge(
+                xr.Coordinates.from_xindex(
+                    RangeIndex.linspace(
```
LucaMarconato, line 223: Looking good

src/spatialdata/_io/io_raster.py
```diff
@@ -160,6 +166,83 @@ def _prepare_storage_options(
...
+    data_tree = xr.DataTree()
+    for scale_idx, (ds_md, ds) in enumerate(zip(multiscale.metadata.datasets, multiscale.images, strict=True)):
+        transf = ds_md.coordinateTransformations[0]
+
+        out_cs = name_to_cs[multiscale.metadata.intrinsic_coordinate_system.name]
+        assert transf.input is not None
+        assert transf.input.path is not None
+        in_cs = CoordSystem(name=str(transf.input.name), axes=[Axis(name=ax.name, type=ax.type) for ax in out_cs.axes])
+
+        ozm_seq = ozm06trans.Sequence(transformations=ds_md.coordinateTransformations)
```
LucaMarconato, line 210: IIRC, according to the specs, here we either have a Sequence (of Scale + Translation) or a Scale. We need to double check.

LucaMarconato: Anyway the logic of doing `.transform_points()` is correct in both cases

src/spatialdata/_io/io_raster.py
```diff
@@ -160,6 +166,83 @@ def _prepare_storage_options(
...
+                        num=extent,
+                        endpoint=True,
+                        dim=ax.name,
+                    )
+                )
+            )
+
+        data_tree[f"scale{scale_idx}"] = xr.Dataset(
+            {
+                "image": xr.DataArray(
```
LucaMarconato, line 235: Not for this PR (we can "resolve the conversation"), but a reminder. This string and the `name="image"` below needs to be documented somewhere, e.g. in docstrings of model classes/contribution guide for developers using spaitaldata/in-memory design doc.

E.g. will the users expect to have always `image` or any string, but always a len(dataset) == 1? The syntax to retrieve the `DataArray` from the `DataTree` will change. 

CC @jan-glx

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+
+from collections.abc import Sequence
+from typing import Final, Literal
+
+import ome_zarr.classes.image as ozi
+import pydantic as pyd
+
+
+class AxisParsingException(Exception):
+    pass
```
LucaMarconato, line 11: I like it here. Just for awareness, in the transformation manager codebase the exceptions are in a separate `exceptions.py` file. https://github.com/scverse/spatialdata/pull/1164/changes

We can choose to keep things local unless there are many exceptions, in that case it is better to place in a dedicated files.

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+
+    @classmethod
+    def try_from_model(cls, model: ozi.Axis) -> Axis:
+        name = model.name
+        if name is None:
+            raise AxisParsingException("Axis doesn't have a name")
+        if model.type != "channel" and model.type != "space":
+            raise AxisParsingException(f"Can't handle axis of type {model.type}")
+        if not isinstance(model.unit, str):
+            raise AxisParsingException("Can't handle axis unit")
```
LucaMarconato: Something to keep in mind for when we'll try to read legacy data. In the old CS we were settings `unit="unit"`. This is not allowed by the specs (having "unit" is a SHOULD, but if they are specified they need to be real units).

LucaMarconato: But also here, shouldn't we allow for model.unit to be `None`?

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+        return Axis(
+            name=name,
+            type=model.type,
+            unit=model.unit,
+            long_name=model.longName,
```
LucaMarconato, line 75: why is here camel case, and elsewhere with underscore?

LucaMarconato: ah it's the ozm one!

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+
+    name: Final[str]
+    axes: Final[tuple[Axis, ...]]
+
+    virtual: Final[bool]
+    """A virtual coordinate system exists as an intermediate step between
+    non-virtual coordinate systems and is usually ignored during serialization"""
+
+    class LegacyAxes:
+        pass
```
LucaMarconato, line 112: unused for now; placeholder for later?

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def __init__(self, name: str, axes: Sequence[Axis], virtual: bool = False):
+        self.name = name
+        self.axes = tuple(axes)
+        self.virtual = virtual
+        if len(self.axes) != len({axis.name for axis in self.axes}):
+            raise ValueError("Axes names must be unique")
```
LucaMarconato, line 119: Can we delegate all the validation to ozm, including this one?

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+            if isinstance(parsed := Axis.try_from_model(axis), Exception):
+                raise CoordSystemParsingException(parsed)  # FIXME
+            axes.append(parsed)
+        return CoordSystem(
+            name=model.name,
+            axes=axes,
+        )
+
+    @classmethod
+    def try_from_model_or_default[T](cls, model: ozi.CoordinateSystem | None, *, default: T) -> CoordSystem | T:
```
LucaMarconato, line 140: Who is going to use this/how?

LucaMarconato: The logic for ensuring that `default` has unique names is something that would live outside this class. I wonder if then even the whole 3-line implementation of this function should live outside.

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Final, Literal
+
+import ome_zarr.classes.image as ozi
```
LucaMarconato, line 6: How about being more explicit and import directly from ome-zarr-models-py? WDYT?

https://github.com/ome/ome-zarr-py/blob/025e83da618ec9e74f4b41f05814fce07b029356/ome_zarr/classes/image.py#L16

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+        return ozi.CoordinateSystem(
+            name=self.name,
+            axes=tuple(ax.to_model() for ax in self.axes),
+        )
+
+    def to_model_cs_ident(self) -> ozi.CoordinateSystemIdentifier | None:
+        input = self.to_model()
+        if input is None:
+            return None
+        return ozi.CoordinateSystemIdentifier(name=input.name)
```
LucaMarconato, line 157: ok good, this is already the foundation for when we'll deal with `Scene`s

src/spatialdata/_io/io_raster.py
LucaMarconato: I saw later in that in `vert.py` there is already some logic needed for this (the `CoordinateSystemIndentifier` usage.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+import numpy as np
+import ome_zarr_models.v06.coordinate_transforms as ozm06trans
+import pydantic as pyd
+import xarray as xr
+
+from spatialdata._types import ArrayLike
+from spatialdata.transformations.graph.vert import Axis, CoordSystem
+
+
+class GarbledInput(Exception):
```
LucaMarconato: :D

LucaMarconato: good to have the exception

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+
+        Notes
+        -------
+        This function will check if the dimensionality of the input and output coordinate systems of the
+        transformation are compatible with the given points.
+        """
+
+    @abstractmethod
+    def to_affine(self) -> AffineEdge:
+        """Convert the transformation to an affine transformation, whenever the conversion can be made."""
```
LucaMarconato, line 66: Up to you if doing it now or later, but I think it is going to be very beneficial to already play around with the displacement class to test code paths where a walk in the transformation graph contains a "non-affinable-transformation".

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+
+    def _validate_transform_points_shapes(self, points: xr.DataArray | xr.DataTree | ArrayLike) -> None:
```
LucaMarconato: why datatree and dataarray here? I think datatree should not be here, and dataarray should not be used for images. Maybe it is to use this as an arraylike with annotated coordinates (e.g. `n` x `dims`, with `dims` for example `['x', 'y', 'k', 'u']`)?

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        """
+        Validate if the shape of the points (coordinats to be transformed) are consistent with the input size of the
```
LucaMarconato: mini typo

LucaMarconato: coordinats -> coordinates

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        transformation.
+        """
+        input_size = len(self.input.axes)
+        if len(points.shape) != 2 or points.shape[1] != input_size:
+            raise ValueError(
+                f"points must be a tensor of shape (n, d), where n is the number of points and d is the "
+                f"the number of spatial dimensions. Points shape: {points.shape}, input size: {input_size}"
```
LucaMarconato: NGFF is more general: the matrix also contains the non-spatial dimensions. The dimensionality constraint to check is the same, but we should not tell the user about spatial dimensions, just dimensions

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+
+        Returns
+        -------
+        The compoesed transformation.
+
+        Notes
+        -------
+        Self is applied first, then the transformation passed as argument.
+        """
+        return SequenceEdge([self, transformation], name=None)  # FIXME: no name?
```
LucaMarconato: I think I had this in the old NGFF classes but never used. Also, nitpick, this would lead to a transformation like [[[t0, t1], t2], t3]; which is fine, but composing in one go from outside may be more appealing [t0, t1, t2, t3].

Anyway, no big deal, with can keep.

LucaMarconato: Regarding the name: it is optional, so we can keep it optional (or let the user pass it as optional argument to compose_with())

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+class AffineEdge(BaseTransfEdge):
+    """The Affine transformation from the NGFF specification."""
+
+    linear: Final[ArrayLike]
+    translation: Final[ArrayLike]
+    affine: Final[ArrayLike]
+
+    def __init__(
+        self,
+        name: str | None,
```
LucaMarconato: why not `= None?`

LucaMarconato: after all the coordinate system name is crucial, the transformation name is not a big deal

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        *,
+        linear: ArrayLike,
+        translation: ArrayLike | None = None,
```
LucaMarconato, line 123: This removes all the ambiguity of having to pass the homogenous affine matrix vs not.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+    @classmethod
+    def from_affine_matrix(
+        cls,
+        *,
+        name: str | None,
+        affine_matrix: ArrayLike,
+        input: CoordSystem,
+        output: CoordSystem,
+    ) -> AffineEdge:
+        """Creates an AffineEdge from a raw affine matrix
```
LucaMarconato: I'd add "in homogenous coordinates"

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        The compoesed transformation.
+
+        Notes
+        -------
+        Self is applied first, then the transformation passed as argument.
+        """
+        return SequenceEdge([self, transformation], name=None)  # FIXME: no name?
+
+    @abstractmethod
+    def to_model(self) -> ozm06trans.AnyTransform:
```
LucaMarconato, line 107: shall we add `try_from_model` as abstractmethod as well? this would mirror what done in vert.py

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        """
+        for out_ax, inp_ax in output_to_input.items():
+            if not input.has_axis(inp_ax):
+                raise ValueError(f"input has no axis named {inp_ax}")
+            if not output.has_axis(out_ax):
+                raise ValueError(f"output has no axis named {out_ax}")
+        if not (len(output_to_input) == output.num_axes == input.num_axes):
+            raise ValueError("input_to_output, input and output must have the same number of axes entries")
+        if len(set(output_to_input.values())) != len(output_to_input):
+            raise ValueError("input_to_output must map unique inputs to unique outputs")
```
LucaMarconato: Do we want to do this validation here instead of relying on ozmp? Added discussion in Zulip. Let's follow up there.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+
+        self.output_to_input = output_to_input
+        super().__init__(input=input, output=output, name=name)
+
+    def __repr__(self) -> str:
+        s = super().__repr__() + "\n"
+        s += "\n".join(f"    {out} <- {inp}\n" for out, inp in self.output_to_input.items())
+        return s
```
LucaMarconato, line 337: Rough edge: MapAxis is basically an affine (made of zeros and a few ones to permute the data). When we do __repr__ on the affine we print the matrix and not the axes (which will be handled by the coordinate systems). Here we do the opposite, we print the axes and don't care about the affine. 

It's not incorrect, but I see that this goes more towards the philosophy of spatialdata transformations, while the affine stays "true" to the NGFF transformations. 

For clarify we could actually always print both: the "index-based/matrix representation" (in this case index 0 goes to 1, index 3 goes to 4 etc; in the affine case the matrix); and then the axes annotation (x -> y, y -> z here; in the case of affine the input and output axes, as we do when we print spatialdata Affine transformations).

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            mapAxis.append(in_idx)
+        return ozm06trans.MapAxis(
+            name=self.name,
+            mapAxis=tuple(mapAxis),
+            input=self.input.to_model_cs_ident(),
+            output=self.output.to_model_cs_ident(),
+        )
+
+
+class TranslationEdge(BaseTransfEdge):
```
LucaMarconato, line 465: Up to you if doing it here or in a follow-up PR. projectAxis is a crucial transform and we should start using it/play around with it soon.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        return AffineEdge(
+            linear=affine_matrix[:-1, :-1],
+            translation=affine_matrix[-1, :-1],
+            input=input,
+            output=output,
+            name=name,
+        )
+
+    def inverse(self) -> BaseTransfEdge:
+        inv = np.linalg.inv(self.affine)
```
LucaMarconato: worth adding a check on the stability by inspecting the conditioning number. We can reuse the code of decompose affine.

LucaMarconato: In general inverting matrices is not a good idea, but I think the matrices coming from spatial alignments should be generally fine to be inverted.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            Output coordinate system of the transformation.
+        """
+        expected_shape = (output.num_axes, input.num_axes)
+        if linear_matrix.shape != expected_shape:
+            raise ValueError(f"linear matrix should have shape {expected_shape}")
+        if input.num_axes != output.num_axes:
+            raise ValueError("input and output should have the same numbe rof axes")
+        if not np.isclose(np.linalg.det(linear_matrix), 1.0):
+            raise ValueError("det(linear_matrix) should be ~= 1")
+        linear_matrix.flags.writeable = False
```
LucaMarconato, line 648: what's this?

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            The transformations which compose the sequence.
+        """
+        if len(transformations) == 0:
+            raise ValueError("Empty transformation list")
+        previous_transf = transformations[0]
+        for current_transf in transformations[1:]:
+            if previous_transf.output != current_transf.input:
+                raise ValueError(f"Mismatched input/output from {previous_transf} to {current_transf}")
+            previous_transf = current_transf
+        self.transformations = transformations
```
LucaMarconato: NGFF transformations in a NGFF sequence may not specify the input and output (sometimes). So we should relax this check.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: It would look nice and symmetrical, but it's not sound from a typing perspective; If we make it an abstract class method, then all child methods have to have a compatible signature, but each class deserializes from a specific model (e.g.: `ScaleEdge` deserializes from `omz.Scale`, `IdentityEdge` deserializes from `omz.Identity`, etc.), but child methods can't have stricter parameters than their parents.

Serializing in `to_model` is fine, though, because the child methods are free to have more precise return types than the parent method.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: Even though `self.rotation` is `Final`, there is nothing preventing it from being modified internally (potentially through another reference that was kept outside of `__init__`). This flags raises an exception if anything tries writing to the array.

Still, for extra paranoia we should probably make a copy of `linear_matrix` so we don't accidentally freeze an array that a user might be meaning to recycle 🤔

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: They don't need to specify input and output in their serialized, `ome-zarr-models-py` form, but in our runtime representation (`*Edge`) we **gurantee** that this is always the case (the type of `BaseTransfEdge.input` is a non-optional `CoordSystem`). If the inner transformations are being read in from a file, then they will be given virtual coordinate systems for input and output.

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: Another issue with delegating to `ozm` is that it can only validate based on their own models, so we'd have to convert e.g. our `Axis` type into _their_ `ozm.Axis` to do the validation, which is also kinda clunky 🤔

src/spatialdata/transformations/graph/edge.py
LucaMarconato: Ok great!

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        composed = self.transformations[0].to_affine().affine
+        for t in self.transformations[1:]:
+            a = t.to_affine()
+            composed = a.affine @ composed
+        return AffineEdge.from_affine_matrix(
+            affine_matrix=composed, input=self.input, output=self.output, name=self.name and f"{self.name}__affine"
+        )
+
+    def transform_points(self, points: ArrayLike) -> ArrayLike:
+        return self.to_affine().transform_points(points)  # FIXME
```
LucaMarconato, line 763: if all the transformations can be turn into affine, this is a good way. But if we have transformations that cannot, we could transform step by step. Another way is to have a function to represent a walk in the transformation graph as a list of [affine, non-affine, affine, non-affine, ...]

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        return out
+
+    def inverse(self) -> BaseTransfEdge:
+        inverse_transformations = [t.inverse() for t in self.transformations]
+        return ByDimensionEdge(
+            transformations=inverse_transformations,
+            input=self.output,
+            output=self.input,
+            name=self.name and f"{self.name}__inverse",
+        )
```
LucaMarconato, line 850: If the ByDimension is not invertible we'll now from the constructor and we'll get a ValueError. I suggest to have a is_invertible property to capture these exceptions and tell users if a transfomraiton is invertible beforehand. Or we could have a NonInvertible exception.

LucaMarconato: (same for the other transformations)

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+    output: CoordSystem | CsGen,
+) -> SequenceEdge:
+    parsed_inners: list[BaseTransfEdge] = []
+
+    base_name = "intermediate" + ("" if not model.name else f"_for_{model.name}")
+    cs_gen: CsGen = output if isinstance(output, CsGen) else CsGen(base_name=base_name)
+    parsed = parse_ngff_transf(
+        input=input,
+        model=model.transformations[0],
+        output=cs_gen if len(model.transformations) > 1 else output,
```
LucaMarconato, line 1054: the transformations in model.transformations don't always specify input/output, but sometimes they do. Here cs_gen is used all the time. The code could be changed to first detect if the input is present (and matches the previous outputs), and then if output is present. If not, cs_gen would be used.

LucaMarconato: I can't find generate being called here in parse_sequence, or in the Sequence class, so it could be that this part is still work in progress.

LucaMarconato (commented):
Finished adding comments. In general it looks really good, great work!

src/spatialdata/transformations/graph/vert.py
ajkswamy: Also, this will currently import `Image` and `Axis` that implement NGFF v0.5. Please mention the development branches of dependencies against which this PR is being developed in the PR description.

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+    unit
+        unit of the axis. For a set of valid options see https://ngff.openmicroscopy.org/
+    long_name:
+        a longer, human-friendly name for this axis
+    """
+
+    name: Final[str]
+    type: Final[Literal["space", "channel"]]
+    unit: Final[str | None]
+    long_name: Final[str | None]
```
ajkswamy, line 33: Maybe we could use [Frozen Data Classes](https://docs.python.org/3/library/dataclasses.html#frozen-instances) with [post-init processing](https://docs.python.org/3/library/dataclasses.html#post-init-processing) here? 

### Benefits I see:
- >... (frozen) dataclasses will add [\_\_setattr__()](https://docs.python.org/3/reference/datamodel.html#object.__setattr__) and [\_\_delattr__()](https://docs.python.org/3/reference/datamodel.html#object.__delattr__) methods to the class. These methods will raise a [FrozenInstanceError](https://docs.python.org/3/library/dataclasses.html#dataclasses.FrozenInstanceError) when invoked.
- We avoid boiler plate in `__init__`

### Drawbacks I see:
- >There is a tiny performance penalty when using frozen=True: [__init__()](https://docs.python.org/3/reference/datamodel.html#object.__init__) cannot use simple assignment to initialize fields, and must use object.__setattr__().

src/spatialdata/transformations/graph/vert.py
ajkswamy: Maybe a custom Exception would help here? Also would be good to add some contextual information like the name of the coordinate system and name of the axes to the exception. This would help users to understand/communicate about this error when a user loads an NGFF 0.6 ome-zarr file and gets this exception. 

I guess such contextual information would help for all our custom exceptions.

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+        self.long_name = long_name
+
+    def cloned_with(self, *, unit: str | None) -> Axis:
+        return Axis(name=self.name, type=self.type, unit=unit or self.unit, long_name=self.long_name)
+
+    def __hash__(self) -> int:
+        return hash((self.name, self.type, self.unit, self.long_name))
+
+    def __repr__(self) -> str:
+        return f"NgffAxis(name={self.name}, type={self.type})"
```
ajkswamy, line 50: Since this is stricter than OZMP 0.6 Axis, better to not have "NgffAxis"

src/spatialdata/transformations/graph/vert.py
```diff
@@ -0,0 +1,201 @@
...
+            return False
+        return (
+            self.name == value.name
+            and self.type == value.type
+            and self.unit == value.unit
+            and self.long_name == value.long_name
+        )
+
+    @classmethod
+    def try_from_model(cls, model: ozi.Axis) -> Axis:
```
ajkswamy: would be good to make the source model clear in the name, like `ozmp_model`

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+
+class GarbledInput(Exception):
+    def __init__(self, message: str, input: pyd.JsonValue) -> None:
+        import json
+
+        super().__init__(message + "\n" + json.dumps(input, indent=4))
+        self.input = input
+
+
+class BaseTransfEdge(ABC):
```
ajkswamy, line 25: I'd recommend "BaseTransformationEdge"

src/spatialdata/transformations/graph/edge.py
ajkswamy: Might this be a good place to add further validation checks that you have documented above under `transform_points`?

>This function will check if the dimensionality of the input and output coordinate systems of the
        transformation are compatible with the given points.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            raise ValueError(f"linear's shape is {linear.shape}. Expected f{(num_outputs, num_inputs)}")
+        expected_translation_shape = (num_outputs,)
+        if translation.shape != expected_translation_shape:
+            raise ValueError(f"translation's shape is {translation.shape}. Expected {expected_translation_shape}")
+
+        self.linear = linear
+        self.translation = translation
+
+        self.affine = np.zeros((num_outputs + 1, num_inputs + 1))
+        self.affine[:-1, :-1] = self.linear
```
ajkswamy, line 161: might be useful to create helper function to convert affine matrix to it's linear and translation parts and vice versa, as this is snippet will be used a lot

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            output=self.input,
+            name=self.name and f"{self.name}__affine",
+        )
+
+    def transform_points(self, points: ArrayLike) -> ArrayLike:
+        self._validate_transform_points_shapes(points)
+        p = np.vstack([points.T, np.ones(points.shape[0])])
+        q = self.affine @ p
+        res = q[: self.output.num_axes, :].T
+        assert isinstance(res, np.ndarray)
```
ajkswamy, line 238: If I understand correctly, this check raises an error for the case when `points` is not `numpy.ndarry`, which in some cases will lead to `res` not being an `numpy.ndarray`. For the case when this happens and `points` is very large, the user will need to wait for a while and then get an error. Would it be possible to catch such cases and throw error early on, before the matrix multiplication?

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            input=self.output,
+            output=self.input,
+            name=self.name and f"{self.name}__affine",
+        )
+
+    def transform_points(self, points: ArrayLike) -> ArrayLike:
+        self._validate_transform_points_shapes(points)
+        p = np.vstack([points.T, np.ones(points.shape[0])])
+        q = self.affine @ p
+        res = q[: self.output.num_axes, :].T
```
ajkswamy, line 237: I think we should avoid the need to transpose points before and after the matrix multiplication. Since this function will not be exposed to users, the complexity would still remains with us, but moved elsewhere or or possibly eliminated. 

Among the solutions to avoid this complexity here, I like the solution that doesn't compose `self.affine` in `__init__`, just keeps `linear` and `translation` parts as read from NGFF, defines and uses helper functions to create the appropriate matrix:
- when converting to OZMP
- named `affine_matrix_t ` of size `(self.input.num_axes + 1, self.output.num_axes +1)` here to do `points_with_additional_column @ affine_matrix_t`. This would be cleaner and perform better for large `points`.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+    def inverse(self) -> BaseTransfEdge:
+        return IdentityEdge(input=self.output, output=self.input, name=self.name and f"{self.name}__inverse")
+
+    def transform_points(self, points: ArrayLike) -> ArrayLike:
+        self._validate_transform_points_shapes(points)
+        return points
+
+    def to_affine(self) -> AffineEdge:
+        return AffineEdge(
+            linear=np.eye(self.input.num_axes),
```
ajkswamy, line 290: In TranslationEdge.to_affine, I see that you have used `np.identity`. Although these two are nearly identical for the use cases here, it would be to use `np.identity`in both places to be consistent.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        if input.num_axes != output.num_axes:
+            raise ValueError("input and output must have same number of dimensions")
+        self.scale = scale
+        super().__init__(input=input, output=output, name=name)
+
+    def __repr__(self) -> str:
+        return super().__repr__() + str(self.scale)
+
+    def inverse(self) -> ScaleEdge:
+        if any(s == 0 for s in self.scale):
```
ajkswamy, line 575: Maybe I'm missing it, why do we need to allow scale values to be zero in our classes, in contrast to NGFF spec?

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+            input=self.input.to_model_cs_ident(),
+            output=self.output.to_model_cs_ident(),
+            scale=tuple(self.scale),
+        )
+
+
+class RotationEdge(BaseTransfEdge):
+    """The Rotation transformation from the NGFF specification."""
+
+    rotation: Final[ArrayLike]
```
ajkswamy, line 609: Frozen Data classes could be useful here as well, and in all these subclasses. Please see https://github.com/scverse/spatialdata/pull/1182/changes#r3863756681 for details.

One additional caveat here is that you will need to NOT freeze the base class and freeze the subclasses (please see [this discussion](https://discuss.python.org/t/frozen-dataclass-and-inheritance/100079) for more details). 

```bash
In [1]: from dataclasses import dataclass

In [2]: class Base:
   ...:     a: int = 0
   ...:     b: int = 0
   ...:
   ...: base = Base()
   ...:
   ...: assert base.a == 0
   ...: assert base.b == 0

In [3]: @dataclass(frozen=True)
   ...: class Child(Base):
   ...:     c: int = 0
   ...:
   ...: child = Child()
   ...: assert child.a == 0
   ...: assert child.b == 0
   ...: assert child.c == 0

In [4]: child.a = 1
---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
Cell In[4], line 1
----> 1 child.a = 1

File <string>:15, in __create_fn__.<locals>.__setattr__(self, name, value)
     13 'Could not get source, probably due dynamically evaluated source code.'

FrozenInstanceError: cannot assign to field 'a'

In [5]: child.b = 1
---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
Cell In[5], line 1
----> 1 child.b = 1

File <string>:15, in __create_fn__.<locals>.__setattr__(self, name, value)
     13 'Could not get source, probably due dynamically evaluated source code.'

FrozenInstanceError: cannot assign to field 'b'

In [6]: child.c = 1
---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
Cell In[6], line 1
----> 1 child.c = 1

File <string>:15, in __create_fn__.<locals>.__setattr__(self, name, value)
     13 'Could not get source, probably due dynamically evaluated source code.'

FrozenInstanceError: cannot assign to field 'c'

In [7]: child.d = 2
---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
Cell In[7], line 1
----> 1 child.d = 2

File <string>:15, in __create_fn__.<locals>.__setattr__(self, name, value)
     13 'Could not get source, probably due dynamically evaluated source code.'

FrozenInstanceError: cannot assign to field 'd'
```

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        input
+            Input coordinate system of the transformation.
+        output
+            Output coordinate system of the transformation.
+        """
+        num_inputs = input.num_axes
+        num_outputs = output.num_axes
+        translation = np.zeros(num_outputs) if translation is None else translation
+
+        expected_linear_shape = (num_outputs, num_inputs)
```
ajkswamy, line 146: Validation code for linear shape and translation shape could be refactored out to static functions, as they are also used in `RotationEdge` and `TranslationEdge` classes. If we use frozen data classes, they would need to go in the base. Else they could go here in `AffineEdge`, and `RotationEdge` and `TranslationEdge` could inherit from it.

Another idea is to make an `AffineBase` dataclass that is not frozen, from which `AffineEdge`, `ScaleEdge`, `TranslationEdge` and `RotationEdge` would inherit, have additional attributed and be frozen data classes. This could also help streamline the overloading to `to_affine`  method, which is very similar across these classes.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        return RotationEdge(
+            linear_matrix=self.rotation.T,
+            input=self.output,
+            output=self.input,
+            name=self.name and f"{self.name}__inverse",
+        )
+
+    def transform_points(self, points: ArrayLike) -> ArrayLike:
+        self._validate_transform_points_shapes(points)
+        res = (self.rotation @ points.T).T
```
ajkswamy, line 667: `res = points @ self.rotation.T` would be cleaner here and would perform much better for large `points`.

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+
+        Parameters
+        ----------
+        transformations
+            The transformations which compose the sequence.
+        """
+        if len(transformations) == 0:
+            raise ValueError("Empty transformation list")
+        previous_transf = transformations[0]
+        for current_transf in transformations[1:]:
```
ajkswamy: maybe we could use itertools.pairwise here

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        super().__init__(input=input, output=output, name=name)
+
+    def __repr__(self) -> str:
+        s = super().__repr__() + "\n"
+        s += "\n".join(str(row) for row in self.rotation)
+        return s
+
+    def inverse(self) -> BaseTransfEdge:
+        return RotationEdge(
+            linear_matrix=self.rotation.T,
```
ajkswamy, line 659: Maybe I'm missing something, but this will be wrong if `self.rotation` has scale and shear components

src/spatialdata/transformations/graph/edge.py
ajkswamy: Would also good to catch LinAlgError and throw a custom exception with more context?

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+                )
+                for axis in other.axes
+            ],
+            virtual=True,
+        )
+        self._cs_count += 1
+        return out
+
+
+def parse_identity(
```
ajkswamy, line 949: Could these `parse....` functions be something like `from_omzp` classmethods in respective classes? Also, could we have `out`/`output` to be of type `CoordSystem | None = `None`?

src/spatialdata/transformations/graph/edge.py
```diff
@@ -0,0 +1,948 @@
...
+        transformations=piecewise_transforms,
+    )
+
+
+def parse_ngff_transf(
+    input: CoordSystem,
+    model: ozm06trans.AnyTransform,
+    output: CoordSystem | CsGen,
+) -> BaseTransfEdge:
+    if isinstance(model, ozm06trans.Identity):
```
ajkswamy, line 1111: Maybe we could have methods `get_ozmp_model_class` in respective classes above and a list of classes of transformations we support and something like 

```
for transformation_class in list_of_supported_transformation_classes:
    if isinstance(model, transformation_class.get_ozmp_model_class()):
         return transformation_class.from_omzp(model, input, output)
else:
    raise NotImplementedError(....)
```

ajkswamy (commented):
Great work @Tomaz-Vieira. I learnt so much reading this PR. Designing these new classes is very complicated and you have done an excellent job picking and choosing the best aspects of spatialdata transformation classes and NGFF specs.

None of the points raised here are blockers and I would love to hear your perspective if the points I raised are wrong or incomplete.

A few comments at the PR level:

1. the folder `graph` could be something_like `graph_io/ngff`

Looking forward to the next draft!

src/spatialdata/transformations/graph/edge.py
LucaMarconato: This is now part of the `spatialdata.transformations.Affine` APIs. We can add it also here. For the "full" decomposition, we could move the math (numpy operations) on a helper functions so that we can reimplement across classes (new NGFF Affine and/or spatialdata Affine) if needed.

src/spatialdata/transformations/graph/edge.py
LucaMarconato: A way forward could be to assume that these vectors are small (just used in tests) and do the performant implementations in the `transform()` API.

src/spatialdata/transformations/graph/edge.py
LucaMarconato: As per the comment here, another approach could be to entirely remove the performance objective from these classes: https://github.com/scverse/spatialdata/pull/1182#discussion_r3881831090. 

Happy to hear your thoughts.

LucaMarconato: But also, I think these transpositions operations are pretty fast.

src/spatialdata/transformations/graph/edge.py
LucaMarconato: Does NGFF restrict the value to non-zeros? If so, we need to be more general than there, because if you try to decompose a generic square affine matrix as scale + translation you will get some scale values equal to zero. And AFAIK NGFF doesn't put a restriction on the rank of affine matrices, so we could encounter that case when decomposing into scale + translation.

src/spatialdata/transformations/graph/edge.py
LucaMarconato: Here the spec guarantees that the columns are orthonormal, so the code will work https://ngff.openmicroscopy.org/rfc/5/index.html#rotation

src/spatialdata/_io/io_raster.py
Tomaz-Vieira: > This string and the name="image" below needs to be documented somewhere

Yup, I agree. I am replicating the current behavior, but I'd really rather have it not use magic strings at all, even if well documented

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: > shouldn't we allow for model.unit to be None?

good catch!

Regarding saving `unit="unit"` in the legacy model, I suppose we could read it in as `None`

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: yup, good idea. I was abusing `ValueError` quite a bit

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: yeah, this is a wonky attempt to simplify code elsewhere. I'll see if I can clean it up

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: It's all supposed to be on top of 0.6, but I'll add it to the PR description as well

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I was probably thinking of using the xarray's labels here but ended up abandoning. I'll just leave it as `ArrayLike` at least for now.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: > I think I had this in the old NGFF classes but never used

It's ripped straight out of there =)

> but composing in one go from outside may be more appealing

I'll also consider a `.flatten()` method

src/spatialdata/transformations/graph/edge.py
ajkswamy: I'm confused. I thought transform() API would call this function to transform points?

src/spatialdata/transformations/graph/edge.py
ajkswamy: In that case, we would need additional checks before casting to OZMP class in `to_model` below

src/spatialdata/transformations/graph/edge.py
ajkswamy: I agree that this will be guaranteed when reading from OME-Zarr compliant with NGFF >= 0.6. However it would be good to have this validation in `__init__` of this class to cover other use cases, for example, when we want to add a user specified rotation transformation to an element or between two elements.

src/spatialdata/transformations/graph/edge.py
ajkswamy: I tried to benchmark and you are right. There is very little reduction in performance due to transpositions when working with large points.

V1: implementation with transpositions; V2: implementation without. 

```
================================================================================
STATISTICS OVER 50 ITERATIONS (times in ms)
================================================================================

Points: 100
  V1: mean=0.0360, median=0.0309, std=0.0185, min=0.0199, max=0.1284
  V2: mean=0.0215, median=0.0167, std=0.0112, min=0.0111, max=0.0645
  Speedup: mean=1.72x, median=1.71x

Points: 1,000
  V1: mean=0.0427, median=0.0337, std=0.0229, min=0.0215, max=0.1295
  V2: mean=0.0404, median=0.0333, std=0.0199, min=0.0208, max=0.0790
  Speedup: mean=1.06x, median=1.04x

Points: 10,000
  V1: mean=0.3186, median=0.2538, std=0.1760, min=0.1139, max=0.8643
  V2: mean=0.3705, median=0.3102, std=0.1965, min=0.1260, max=0.8943
  Speedup: mean=0.88x, median=0.86x

Points: 100,000
  V1: mean=3.5154, median=2.4191, std=1.9560, min=2.0646, max=8.6407
  V2: mean=3.5622, median=2.7137, std=1.6593, min=2.3121, max=7.3514
  Speedup: mean=0.97x, median=0.92x

Points: 1,000,000
  V1: mean=40.1754, median=34.6711, std=12.3971, min=29.8887, max=72.1030
  V2: mean=39.5384, median=33.5323, std=14.4447, min=26.2169, max=73.5493
  Speedup: mean=1.04x, median=1.00x
```

<details>

<summary>Code to reproduce benchmarking</summary>

```python
import numpy as np
import time


def transform_points_v1(linear_part, translation_part, points):
    """Apply affine transform to points (original approach, corrected)."""
    d = linear_part.shape[0]
    affine = np.eye(d + 1)
    affine[:d, :d] = linear_part
    affine[:d, d] = translation_part
    p = np.vstack([points.T, np.ones(len(points))])
    q = affine @ p
    return q[:d, :].T


def transform_points_v2(linear_part, translation_part, points):
    """Apply affine transform without transposing points."""
    d = linear_part.shape[0]
    affine = np.eye(d + 1)
    affine[:d, :d] = linear_part
    affine[:d, d] = translation_part
    p_homogeneous = np.column_stack([points, np.ones(len(points))])
    q = p_homogeneous @ affine.T
    return q[:, :d]


def benchmark():
    d = 3  # 3D points
    linear_part = np.random.randn(d, d)
    translation_part = np.random.randn(d)

    num_points_list = [100, 1_000, 10_000, 100_000, 1_000_000]
    num_iterations = 50
    num_repeats = 10  # Inner repeats for stable timing per iteration

    # Store all timing results: {n: {'v1': [times], 'v2': [times]}}
    results = {n: {"v1": [], "v2": []} for n in num_points_list}

    print("Running 50 iterations of benchmark...")
    for _ in range(num_iterations):
        for n in num_points_list:
            points = np.random.randn(n, d)

            # Benchmark V1 (with transpositions)
            start = time.perf_counter()
            for _ in range(num_repeats):
                transform_points_v1(linear_part, translation_part, points)
            time_v1 = (time.perf_counter() - start) / num_repeats * 1000  # ms
            results[n]["v1"].append(time_v1)

            # Benchmark V2 (no transpositions)
            start = time.perf_counter()
            for _ in range(num_repeats):
                transform_points_v2(linear_part, translation_part, points)
            time_v2 = (time.perf_counter() - start) / num_repeats * 1000  # ms
            results[n]["v2"].append(time_v2)

    # Print statistics
    print("\n" + "=" * 80)
    print("STATISTICS OVER 50 ITERATIONS (times in ms)")
    print("=" * 80)

    for n in num_points_list:
        v1_times = np.array(results[n]["v1"])
        v2_times = np.array(results[n]["v2"])
        speedups = v1_times / v2_times

        print(f"\nPoints: {n:,}")
        print(
            f"  V1: mean={np.mean(v1_times):.4f}, median={np.median(v1_times):.4f}, "
            f"std={np.std(v1_times):.4f}, min={np.min(v1_times):.4f}, max={np.max(v1_times):.4f}"
        )
        print(
            f"  V2: mean={np.mean(v2_times):.4f}, median={np.median(v2_times):.4f}, "
            f"std={np.std(v2_times):.4f}, min={np.min(v2_times):.4f}, max={np.max(v2_times):.4f}"
        )
        print(f"  Speedup: mean={np.mean(speedups):.2f}x, median={np.median(speedups):.2f}x")


if __name__ == "__main__":
    benchmark()
```

</details>

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: The  `assert` is just coming form the old Ngff code and I'd say it's probably because the function used to deal with more types than just `np.ndarray`, and it was there just as a sanity check, and should never trigger outside of development time.

My type checker is happy assuming that `res` is always a numpy array, so I;ve removed the assert.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: Interesting, I completely missed the observation on zeros in NGFF. Though to be fair, the spec says it "SHOULD be non-zero", so strictly speaking this is still valid.

I'll take this into account on the next PR when I do the serialization

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: I've moved those classes to frozen dataclasses. I don't quite like how this intrudes in `__init__`, but for such simple classes I guess I'm ok with it if it might take away a foot gun from a user

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: It would be nice to have them as class methods, but any inheritance on classes that implement that method wouldn't be sound from a typing perspective. Since each function takes different arguments, they would be overriding the parent class in an invalid way.

As for `output: CoordSystem | None`... maybe. The idea of using the coodinate system generator was two-fold:

1- Move common logic into a common place (the `CsGen` methods);
2- Give an opportunity to the caller to control how the intermediate coord systems are to be created, which we don't get by simply passing `None` in.

I think there is some work do be done here still, as I myself don't quite like `CsGen`, but this might get clearer as we progress.

One idea would be to have the parsing produce something like `Generator[..., CoordSystem, ...]` so that the caller can decide on the intermediate coord systems at their own pace, but I haven't invested much time into this yet.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: The problem is that the type checker can't see into this style; the proposed `get_ozmp_model_class` would return either a `Type[BaseTransformationEdge]` or a union of all `*Edge` classes. From there, the only way to call a `parse_model` method would be if those methods took a more general model type (e.g. `models.AnyTransforml`), rather than exactly the model type that can be deserialized (e.g. `models.Scale`), so we lose some more typing information there. 

Also, this style allows for an exhaustive check; Right now the type checker is aware that we don't have all transformations yet, which is why I `raise NotImplemented`. But the moment we implement them all, we'll replace that `raise NotImplemented` with an `assert_never(model)`, and if ever NGFF adds a new transformation, the type checker will let us know, by detecting that we haven't exhaustively checked for every possible type of transform.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: Removed transpositions of points. Code looks a lot cleaner like that as well =)

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I'll leave it for now. I am considering making many transforms subclasses of `AffineEdge`, and so that would mean that `Affine` could not be made frozen anyway =/

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I agree; it looks a lot like many transfomrs should just be subclasses of `AffineEdge`. The reason I don't want to do it right now is because I want to see if they stay similar in all of their usages. If they do, then it's easy to contract everything into children of `Affine`. If they don't, it's rather annoying to split them off of their parent. But i suspect the inheirtance will come in the future

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: Agree. Done.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: Even though we indirectly depend on itertools, I didn't want to tie us to the dependency for a few lines.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I was checking for `det(M)==1`, but i was not checking for orthonormality. It's done now, so this is guaranteed to work

src/spatialdata/_core/transformation_manager/exceptions.py
```diff
@@ -227,3 +229,65 @@ class TransformationManagerWarning(UserWarning):
...
+class DeterminantDifferentFromOne(Exception):
+    def __init__(self, matrix: ArrayLike) -> None:
+        self.matrix = matrix
+        super().__init__("Matrix does not have det(M) == 1")
+
+
+class NotOrthonormalError(Exception):
+    def __init__(self, matrix: ArrayLike) -> None:
+        self.matrix = matrix
+        super().__init__("Matrix is not orthonormal")
```
LucaMarconato, line 276: In these 2 exception matrix is passed but it is not used.

src/spatialdata/_io/io_raster.py
LucaMarconato: RangeIndex is available from xarray 2025.04.0 (https://github.com/pydata/xarray/commit/38169010846e772a085158c356ef2ed6b8f2086a), so we need to bump our min dependencies in pyproject.toml

src/spatialdata/_io/io_raster.py
Tomaz-Vieira: Yup, tat will be part of the next pR =)

src/spatialdata/_io/io_raster.py
Tomaz-Vieira: nice catch O.O

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: I left it in vert.py at least for now to avoid circular dependencies in exceptions.py

src/spatialdata/transformations/graph/vert.py
Tomaz-Vieira: Yup, was a placeholder, removed for now

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: `to_affine` now returns an optional, making it clear that this path is fallible and must be handled, but I'll add the transform itself to a later PR

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I think I'd rather avoid the indexed output since we have everything labeled (axes have names, arrays have dims), and hopefully we eventually completely hide the fact that indices have any meaning for us

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: added `ProjectAxisEdge` =)

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I've stopped raising errors and started returning a None instead so the fallibility of the method is more obvious. Also added the condition check like what we do in decomposition.

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: I'll do it on the next PR, when I introduce new transforms that cannot be transformed to affine. The type checker will make sure that I don't forget it once `to_affine` returns an optional =)

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: Maybe this is from before `inverse()` started returning `SomeTransform | None`? Because now that that fallibility is visible at the type level.

src/spatialdata/_core/transformation_manager/exceptions.py
Tomaz-Vieira: yeah, it's just context for the exception in case someone that catches the exception might want to inspect the matrix, but I can put it in the message as well

src/spatialdata/transformations/graph/edge.py
Tomaz-Vieira: This is kind of ugly, but the alternatives are kinda uglier. The reason I don't check `model.input` and `model.output` is because It does away with the resolution of the coordinate system; by the time `parse_*` is called, the caller must have determined what the `input` is and what the output is, if any. To determine those coordinate systems otherwise would require passing in a

`Mapping[CoordinateSystemIdentifier, CoordSystem]`

that could be checked for `model.input` and `model.output`. But since `model.input` and `model.output ` can be `None` (in sequences), the `parse_*`, functions also have to take in an `input` parameter, which is either always resolved (`input: CoordSystem`), or isn't (`input: CoordSystem | None`), which pollutes the code with more code paths. So I just opted for doing the resolving outside of the function and not passing in the coordinate system map at all; this resolved input is either something that is coming from the coord systems defined in the NGFF or is just the output CS of the previous item in a sequence. 

Also, `prase_sequence` is now using the coordinate generator as well, so the last comment is probably just outdated.

PR #1184, opened 2026-08-19
Do not erase the anndata encoding metadata from the table group

### LINKED ISSUE #1183: Tables written on main lose the anndata encoding-type/encoding-version attributes
While testing Thyra (our MSI converter) against current main I noticed that the
`tables/<name>` group comes out without the anndata `encoding-type` and
`encoding-version` attributes. On 0.8.0 they are there. Is this intended, or a
side effect of #1170? I could not find it mentioned in the PR or the changelog,
and the tables format version did not change, so I am guessing it is accidental,
but wanted to check before assuming.

Repro:

```python
import json, numpy as np, pandas as pd
from anndata import AnnData
from spatialdata.models import TableModel

obs = pd.DataFrame({"region": pd.Categorical(["r"] * 4),
                    "instance_id": np.arange(4)})
table = TableModel.parse(AnnData(X=np.ones((4, 3)), obs=obs),
                         region="r", region_key="region",
                         instance_key="instance_id")
SpatialData(tables={"t": table}).write("bug.zarr")

attrs = json.load(open("bug.zarr/tables/t/zarr.json"))["attributes"]
assert "encoding-type" in attrs, sorted(attrs)
```

On 0.8.0 this passes. On main (21abd1e) it fails with
`['instance_key', 'region', 'region_key', 'spatialdata-encoding-type', 'version']`.

`SpatialData.read` still works, but anndata level readers dispatch on those
attrs, so `anndata.io.read_elem` on the table group now returns a plain dict
instead of an AnnData, and `read_lazy` does the same when given the group.
Nothing raises, you just silently get the wrong type. Per the anndata on-disk
spec both fields are required on every element.

I think the cause is the ordering in `write_table`: the handle comes from
`group.require_group(name)` before `table.write_zarr(...)`, and the spatialdata
attrs are set through that handle afterwards. zarr v3 keeps all attributes in
one zarr.json and `update_attributes` writes back the handle's cached view,
which was captured while the group was still empty, so it overwrites what
anndata just wrote. Three lines show it:

```python
handle = root.require_group("t")                       # caches empty attrs
zarr.open_group(store + "/t").attrs["encoding-type"] = "anndata"
handle.attrs["x"] = "y"                                # encoding-type is gone
```

If that is right, fetching the handle after the write (like 0.8.0 did with
`group[name]`) should fix it. Manually restoring the two attrs on an affected
store makes `read_elem` return an AnnData again, so they seem to be the whole
difference.

spatialdata main @ 21abd1e, anndata 0.13.2, zarr 3.1.6, python 3.12, Windows.

Happy to open a PR if useful, and let me know if you need any more info.

Tomatokeftes:
Fixes #1183.

`write_table` acquires the table group handle with `group.require_group(name)`
before the anndata write and sets the spatialdata attributes through that
handle afterwards. zarr v3 writes attributes as a whole document from the
handle's cached view, which was captured while the group was still empty, so
the first attribute assignment erased the `encoding-type` and
`encoding-version` attributes that anndata had just written. As a result
`anndata.io.read_elem` and `anndata.experimental.read_lazy` silently returned
a plain dict instead of an AnnData when given the table group.

The fix re-fetches the handle after the write, which is the ordering 0.8.0
used (`tables_group = group[name]` after `write_adata`). It sits after the
if/else so it covers both the `write_zarr` and the `write_adata` branch.

Also adds a regression test asserting the anndata encoding metadata and the
spatialdata attributes are both present on the written table group, and that
`read_elem` on the group returns an AnnData. The test fails on current main
with `KeyError: 'encoding-type'` and passes with the fix.

Verified against the downstream converter that hit this (Thyra): its
previously failing read_lazy contract test passes with this branch, and its
converted stores are byte-identical to 0.8.0 output again.

Note on local testing: on my Windows machine some io tests flake with
`PermissionError: [WinError 5]` on the atomic `.zattrs` rename in zarr v2
raster writes. That happens at the same rate on unmodified main, so it is
unrelated, but worth knowing when comparing local runs to CI.

Let me know if you would like a changelog entry or any changes.

### DIFF
--- a/src/spatialdata/_io/io_table.py
+++ b/src/spatialdata/_io/io_table.py
@@ -119,6 +119,12 @@ def write_table(
 
         write_adata(group, name, table)
 
+    # Re-fetch the group before setting the attributes below: the handle obtained above was cached while the group
+    # was still empty, and zarr writes attributes as a whole document based on the handle's cached view, so writing
+    # through the stale handle would erase the `encoding-type`/`encoding-version` metadata that anndata just wrote
+    # (https://github.com/scverse/spatialdata/issues/1183).
+    table_group = group[name]
+
     table_group.attrs["spatialdata-encoding-type"] = group_type
     table_group.attrs["region"] = region
     table_group.attrs["region_key"] = region_key
--- a/tests/io/test_readwrite.py
+++ b/tests/io/test_readwrite.py
@@ -16,6 +16,7 @@ import pyarrow.parquet as pq
 import pytest
 import zarr
 from anndata import AnnData
+from anndata.io import read_elem
 from numpy.random import default_rng
 from packaging.version import Version
 from pandas.testing import assert_series_equal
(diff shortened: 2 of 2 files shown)


ajkswamy (approved):
LGTM

PR #1185, opened 2026-08-19
docs: link API entries to their source on GitHub

jan-glx:
scanpy and anndata already link API entries to their source on GitHub, via `scanpydoc` + `sphinx.ext.linkcode`. spatialdata enables neither `linkcode` nor `viewcode`, so its API reference has no `[source]` links at all, see, e.g., [`api/io.html`](https://spatialdata.scverse.org/en/stable/api/io.html).

This PR follows scanpy's approach:
- `sphinx.ext.linkcode` emits the links
- [`scanpydoc.rtd_github_links`](https://icb-scanpydoc.readthedocs-hosted.com/en/latest/scanpydoc.rtd_github_links.html) builds the URLs from the `html_context` already in `docs/conf.py` (See also discussion in [cookiecutter-scverse#103](https://github.com/scverse/cookiecutter-scverse/issues/103))
- `docs/extensions/git_ref.py`, taken from [scanpy](https://github.com/scverse/scanpy/blob/main/docs/extensions/git_ref.py), resolves the current tag so released docs link to their own code.

Will need revisiting if [theislab/scanpydoc#238](https://github.com/theislab/scanpydoc/issues/238) lands.


jan-glx:
52ce669 additionally adds an exception for PR doc builds so you can see it working live here: https://scverse-spatialdata--1185.org.readthedocs.build/en/1185/api/io.html

LucaMarconato (approved):
Thanks! Tested on the readsthedocs deployment for this PR and works great.

docs/extensions/git_ref.py
```diff
@@ -0,0 +1,72 @@
...
+    Uses branch/tag name if found, otherwise uses commit hash.
+    """
+    # On Read the Docs pull request previews, HEAD is a local branch such as
+    # "external-1185" that does not exist on GitHub, so use the commit instead.
+    # GitHub serves a pull request's head commit from the base repository, also
+    # when the branch itself lives on a fork.
+    if os.environ.get("READTHEDOCS_VERSION_TYPE") == "external" and (
+        commit := os.environ.get("READTHEDOCS_GIT_COMMIT_HASH")
+    ):
+        return commit
```
LucaMarconato, line 47: Seems useful also for scanpy/scversedocs. CC-ing @flying-sheep

docs/extensions/git_ref.py
```diff
@@ -0,0 +1,72 @@
...
+
+def set_ref(app: Sphinx, config: Config) -> None:
+    """`config-inited` hook to set `html_context["github_version"]`."""
+    config["html_context"]["github_version"] = get() or "main"
+
+
+def setup(app: Sphinx) -> ExtensionMetadata:
+    """App setup hook."""
+    app.connect("config-inited", set_ref)
+    return {"parallel_read_safe": True}
```
LucaMarconato: I found `ExtensionMetadata(parallel_read_safe=True)` clearer from a types point of view. Why did you change?

docs/extensions/git_ref.py
jan-glx: yes, potentially. But how often do you need to go to the PR source from a PR's docs build? More importantly, `git_ref.py` should move to scanpy/scversedocs.

docs/extensions/git_ref.py
jan-glx: For watermarking ;)

I agree, the "reduced runtime dependencies" are rather imaginary and definitely not worth the deviation. I should have done the diff to the original `git_ref.py` with a difftool rather than by eye. [Added it below](https://github.com/scverse/spatialdata/pull/1185/changes/176ded9fa9bafef9f7eb7b7bdee59fe338b13814) now and will revert the unnecessary changes that escaped my eye.

Happy to rebuild a clean history unless we squash anyway.

LucaMarconato: Squashing, no big deal.

jan-glx:
Now live https://spatialdata.scverse.org/en/latest/api/io.html

PR #1186, opened 2026-08-20
Improve affine decomposition

LucaMarconato:
*Claude generated PR description, manually verified*

Port the RQ-based decomposition algorithm from `Spatial-Innovation-Team/transfo`
to replace the old 2D-only, theta-based rotation logic in
`_decompose_transformation`. The new algorithm is dimension-generic, warns on
ill-conditioned linear parts, and encodes the reflection as a single
axis flip instead of an arbitrary diagonal matrix.

Drop the spatialdata-specific restrictions that no longer apply: the `z` axis
and the `c` channel are now decomposed like any other axis, and axes are
allowed to come out in a different order between input and output (checked
by axis set rather than exact tuple order).

Split `_decompose_transformation` into `_decompose_transformation_simple` and
`_decompose_transformation_full`, each returning a fixed-size tuple of the
component transformations instead of a `Sequence`, so callers get a
concretely-typed tuple (2 vs. 5 components) rather than a `Union`.

Reorganize the tests into `TestSimpleDecomposition`/`TestFullDecomposition`
classes, consolidating round-trip coverage into a single
`DECOMPOSE_TRANSFORMATION_CASES` table (with descriptive ids) shared by both
classes, and keeping only the checks that can't be expressed as a
matrix-in/reconstructed-matrix-out case (warnings, component types/arity,
algorithmic invariants).

### DIFF
--- a/src/spatialdata/transformations/transformations.py
+++ b/src/spatialdata/transformations/transformations.py
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
     return Affine(matrix, input_axes=input_axes, output_axes=output_axes)
 
 
-def _decompose_transformation(
-    transformation: BaseTransformation, input_axes: tuple[ValidAxis_t, ...], simple_decomposition: bool = True
-) -> Sequence:
+def _validate_square_affine_for_decomposition(
+    transformation: BaseTransformation, input_axes: tuple[ValidAxis_t, ...]
+) -> tuple[ArrayLike, ArrayLike, ArrayLike]:
     """
-    Decompose a given 2D transformation into a sequence of predetermined types of transformations.
+    Validate that a transformation can be decomposed, and extract the parts of its affine matrix.
 
     Parameters
     ----------
     transformation
         The transformation to decompose. It is assumed to be of a type that can be represented as a single affine
-        transformation. It should leave the input axes unmodified, and it should not transform the c channel, if this
-        is present.
+        transformation. It should leave the set of input axes unmodified (adding, dropping or renaming an axis is
+        not allowed), but the axes are allowed to come out in a different order: the matrix is always queried back
+        in ``input_axes`` order before being decomposed. There is no restriction on which axes are present: spatial
+        axes (``x``, ``y``, ``z``) and the ``c`` channel axis are all decomposed uniformly, as the matrix is
+        treated as a generic square affine.
     input_axes
-        The axes of the data the transformation is to be applied to
-    simple_decomposition
-        If true, decomposes a transformation into it's linear part (affine without translation) and translation part,
-        otherwise decomposes it into a sequence of reflection, rotation, shear, scale, translation.
+        The axes of the data the transformation is to be applied to.
 
     Returns
     -------
-    sequence
-        Returns a sequence of transformations (class :class:`~spatialdata.transformations.Sequence`) which operates only
-        on the spatial part (no c channel). The output sequence will contain either 2 either 5 transformations in the
-        following order (the first is applied first).
-        Case `simple_decomposition = True`.
+    A tuple ``(matrix, translation_part, linear_part)`` where ``matrix`` is the full homogeneous affine matrix (with
+    both rows and columns ordered as ``input_axes``), ``translation_part`` is its last column (excluding the
+    homogeneous row), and ``linear_part`` is the square matrix obtained by removing the last row and column of
+    ``matrix``.
 
-            1. Linear part (affine): linear part of the affine transformation, represented as a
-            :class:`~spatialdata.transformations.Affine` transformation.
-            2. Translation. Represented as a :class:`~spatialdata.transformations.Translation` transformation.
-
-        Case `simple_decomposition = False`.
-
-            1. Reflection. Represented as :class:`~spatialdata.transformations.Scale` transformation with elements in
-                {1, -1}.
-            2. Rotation. Represented as an :class:`~spatialdata.transformations.Affine` transformation which in its
-                matrix form presents itself as an homogeneous affine matrix with no translation part and determinant 1.
-                Please look at the source code of this function if you need to recover the angle theta.
-            3. Shear. Represented as an :class:`~spatialdata.transformations.Affine` transformation which in its matrix
-                form presents itself as an homogeneous affine matrix with no translation part. The matrix is upper
-                triangular with diagonal elements all equal to 1.
-            4. Scale. Represented as a :class:`~spatialdata.transformations.Scale` transformation with positive
-            elements.
-            5. Translation. Represented as a :class:`~spatialdata.transformations.Translation` transformation.
-
-        Note that some of these transformations may be identity transformations.
+    Raises
+    ------
+    ValueError
+        If the transformation changes the set of input axes (as opposed to merely reordering them).
+    RuntimeWarning
+        If the linear part of the affine has a large condition number, in which case the decomposition may be
+        numerically inaccurate.
     """
     output_axes = _get_current_output_axes(transformation=transformation, input_axes=input_axes)
-    if input_axes != output_axes:
-        raise ValueError("The transformation should leave the input axes unmodified.")
-    if "z" in input_axes:
-        raise ValueError("The transformation should not transform the z axis.")
-    affine = transformation.to_affine(input_axes=input_axes, output_axes=output_axes)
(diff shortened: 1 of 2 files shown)


LucaMarconato:
@Tomaz-Vieira can you please review?

Also, the GitHub diff view is not the most useful for this PR since the code is now split into functions. I suggest to use a standard diff view, so that most of the code is easily shown as untouched.

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
     return Affine(matrix, input_axes=input_axes, output_axes=output_axes)
 
 
-def _decompose_transformation(
-    transformation: BaseTransformation, input_axes: tuple[ValidAxis_t, ...], simple_decomposition: bool = True
-) -> Sequence:
+def _validate_square_affine_for_decomposition(
+    transformation: BaseTransformation, input_axes: tuple[ValidAxis_t, ...]
```
Tomaz-Vieira: Could we change parameters to just `transformation: Affine`?  The docstring even says _"It is assumed to be of a type that can be represented as a single affine"_, so we might as well just  require proof of that, by asking straight for an instance of `Affine`. Then we can just check that it's square, and use exactly `transformation.input_axes` and `transformation.output_axes` rather than another `input_axes` argument and the fallible `_get_current_output_axes`. This way we move a bunch of errors out of this function

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
+) -> tuple[ArrayLike, ArrayLike, ArrayLike]:
```
Tomaz-Vieira: Could we return `tuple[Affine, Scale, Translation]` instead? I think it signals our intent even better than the raw matrices.

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
-            input_axes=axes,
-            output_axes=axes,
-        )
-        rotation = _compose_affine_from_linear_and_translation(
-            linear=adjusted_rotation_matrix,
-            translation=np.zeros(rotation_matrix.shape[0]),
-            input_axes=axes,
-            output_axes=axes,
+    translation_part = matrix[:-1, -1]
+    linear_part = matrix[:-1, :-1]
```
Tomaz-Vieira: Could these be fields/properties/methods of `Affine`? Like `my_affine.linear` and `my_affine.translation`?

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
+    Parameters
+    ----------
+    transformation
+        The transformation to decompose. See :func:`_validate_square_affine_for_decomposition`.
+    input_axes
+        The axes of the data the transformation is to be applied to.
+
+    Returns
+    -------
+    A tuple ``(linear, translation)``, applied in this order (``linear`` first), whose composition equals
```
Tomaz-Vieira: I think the intention was something like

>A tuple ``(linear, translation)``, whose composition equals `transformation`, applied in the following order:

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
         )
-        inversion = Scale(np.diag(adjusted_qq), axes=axes)
-        translation = Translation(translation_part, axes=axes)
-        sequence = Sequence([inversion, rotation, shear, scale, translation])
-    check_m = sequence.to_affine_matrix(input_axes=input_axes, output_axes=input_axes)
+    return matrix, translation_part, linear_part
+
+
+def _decompose_transformation_simple(
+    transformation: BaseTransformation, input_axes: tuple[ValidAxis_t, ...]
```
Tomaz-Vieira: Could we change parameters to just `transformation: Affine`? In fact, could we even make this a method of `Affine` 🤔 ?

The docstring for `_validate_square_affine_for_decomposition` even says _"It is assumed to be of a type that can be represented as a single affine"_, so we might as well  require proof of that, by asking directly for an instance of `Affine`. Then we can just check that it's square, and use exactly `transformation.input_axes` and `transformation.output_axes` rather than another `input_axes` argument and the fallible `_get_current_output_axes`. This way we move a bunch of errors out of this function, plus we don't have this indirection on the param requirements, which requires the reader to go read `_validate_square_affine_for_decomposition`

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
+
+        1. Linear part (affine): linear part of the affine transformation, represented as a
+           :class:`~spatialdata.transformations.Affine` transformation.
+        2. Translation. Represented as a :class:`~spatialdata.transformations.Translation` transformation.
+
+    Note that some of these transformations may be identity transformations.
+    """
+    matrix, translation_part, linear_part = _validate_square_affine_for_decomposition(transformation, input_axes)
+
+    linear = _compose_affine_from_linear_and_translation(
```
Tomaz-Vieira: I think this part would also be eliminated by requiring the `transformation` param to be only a square `transformation: Affine`

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
+        linear=linear_part,
+        translation=np.zeros(linear_part.shape[0]),
+        input_axes=input_axes,
+        output_axes=input_axes,
+    )
+    translation = Translation(translation_part, axes=input_axes)
+
+    check_m = Sequence([linear, translation]).to_affine_matrix(input_axes=input_axes, output_axes=input_axes)
```
Tomaz-Vieira: To be super pedantic, this could be moved into an `if __debug__`, since this value is only created for the sake of the assertion

src/spatialdata/transformations/transformations.py
```diff
@@ -859,136 +859,196 @@ def _compose_affine_from_linear_and_translation(
...
+    assert np.allclose(check_m, matrix)
+    return linear, translation
+
+
+def _decompose_transformation_full(
+    transformation: BaseTransformation, input_axes: tuple[ValidAxis_t, ...]
```
Tomaz-Vieira: same argument for at last making `transformation` be of type `Affine`

Tomaz-Vieira (commented):
The decomposition is a very nice maneuver =)

My comments are more on the code organization side of things; in particular, I argue for expressing the requirements of the functions more clearly, and leaving most of the responsibility of meeting those requirements to the callers of the functions.

src/spatialdata/transformations/transformations.py
LucaMarconato: Yes, your wording is better. I will rephrase.

src/spatialdata/transformations/transformations.py
LucaMarconato: Great point, I'll do.

src/spatialdata/transformations/transformations.py
LucaMarconato: Yes, they can be property of `Affine`, I'll do.

src/spatialdata/transformations/transformations.py
LucaMarconato: Good point!

src/spatialdata/transformations/transformations.py
LucaMarconato: Done. And I turned the `_validate_square_affine_for_decomposition()` into a real validation function (no return). The 3 values that previously were returned are now computed easily in the caller functions.

LucaMarconato:
@Tomaz-Vieira thanks for the review, really helpful! I have addressed all the review comments. I have also renamed the two functions form `decompose_affine_simple`/`full` to:
- `_decompose_into_linear_and_translation`
- `_decompose_into_5_simple_transformations` (not sure if it's a good name).

I will merge as it is required for the bounding box PR; if we need to change something (like the name of the functions) we can do it in a follow up PR; they are private and unreleased anyway.

LucaMarconato:
Had to do a manual rebase as I stacked PRs complained about the merge commit.

PR #1193, opened 2026-08-25
fix: #1162; preserving order of obs after _filter_table_by_elements

### LINKED ISSUE #1162: Relational query reorders table `obs` by region in 0.8
Since #1131, filtering a table that annotates **multiple regions** reorders its `obs` rows by region instead of preserving the original order. This shifts table rows relative to element geometries and broke an ordering assertion in a `spatialdata-plot` test.

**Cause:** `_filter_table_by_elements` was changed from a boolean mask (order-preserving) to `join_spatialelement_table(how="left")`, which does `obs.groupby(region)` and thus regroups the rows.

### MRE

```python
import numpy as np
import pandas as pd
from anndata import AnnData
from geopandas import GeoDataFrame
from shapely.geometry import Point
import spatialdata as sd
from spatialdata.models import ShapesModel, TableModel


def circles(n):
    gdf = GeoDataFrame({"geometry": [Point(i, i) for i in range(n)], "radius": [1.0] * n})
    return ShapesModel.parse(gdf)


# table annotates two regions, obs rows INTERLEAVED: a,b,a,b,a,b
obs = pd.DataFrame(
    {
        "region": pd.Categorical(["a", "b", "a", "b", "a", "b"]),
        "instance_id": [0, 0, 1, 1, 2, 2],
        "label": ["a0", "b0", "a1", "b1", "a2", "b2"],
    }
)
table = TableModel.parse(
    AnnData(X=np.zeros((6, 1)), obs=obs, var=pd.DataFrame(index=["g0"])),
    region=["a", "b"],
    region_key="region",
    instance_key="instance_id",
)
sdata = sd.SpatialData(shapes={"a": circles(3), "b": circles(3)}, tables={"table": table})

before = list(sdata["table"].obs["label"])
after = list(sdata.filter_by_coordinate_system("global")["table"].obs["label"])
assert after == before, f"obs reordered {before} -> {after}"
```

### Result
```
┌─────────┬───────────────────────────────────────┐
│ version │ obs after filter_by_coordinate_system │
├─────────┼───────────────────────────────────────┤
│ 0.7.2   │ a0, b0, a1, b1, a2, b2 (preserved) ✅ │
├─────────┼───────────────────────────────────────┤
│ 0.8     │ a0, a1, a2, b0, b1, b2 (regrouped) ❌ │
└─────────┴───────────────────────────────────────┘

### PR DESCRIPTION
**Pull request by LucaMarconato**
Closes #1162

### DIFF
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -422,8 +422,7 @@ def _left_exclusive_join_spatialelement_table(
                 group_df = groups_df.get_group(name)
                 table_instance_key_column = group_df[instance_key]
                 if element_type in ["points", "shapes"]:
-                    mask = np.full(len(element), True, dtype=bool)
-                    mask[table_instance_key_column.values] = False
+                    mask = ~np.isin(element.index, table_instance_key_column.values)
                     masked_element = element.loc[mask, :] if mask.sum() != 0 else None
                     element_dict[element_type][name] = masked_element
                 else:
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -469,6 +469,12 @@ def _left_join_spatialelement_table(
         # if nan were present, the dtype would have been changed to float
         if joined_indices.dtype == float:
             joined_indices = joined_indices.astype(int)
+        # `groupby(region)` above collects the matching table rows grouped by region, which does not
+        # preserve the original `table.obs` row order when a table annotates multiple interleaved
+        # regions. For `match_rows="no"` there is no element-driven ordering to honor, so
+        # restore the original table row order, as would be expected for a semi-join.
+        if match_rows == "no":
+            joined_indices = joined_indices.sort_values()
     joined_table = table[joined_indices.tolist(), :].copy() if joined_indices is not None else None
     _inplace_fix_subset_categorical_obs(subset_adata=joined_table, original_adata=table)
     if joined_table is not None:
--- a/src/spatialdata/_core/query/relational_query.py
+++ b/src/spatialdata/_core/query/relational_query.py
@@ -255,6 +255,13 @@ def _right_exclusive_join_spatialelement_table(
     match_rows: Literal["left", "no", "right"],
     filter_label_pixels: bool | None = None,
 ) -> tuple[dict[str, Any], AnnData | None]:
+    if match_rows == "left":
+        warnings.warn(
+            "Matching rows 'left' is not supported for 'right_exclusive' join; it will be treated as 'no'.",
+            UserWarning,
+            stacklevel=2,
+        )
+        match_rows = "no"
     regions, region_column_name, instance_key = get_table_keys(table)
     if isinstance(regions, str):
         regions = [regions]
@@ -300,7 +307,12 @@ def _right_join_spatialelement_table(
     filter_label_pixels: bool | None = None,
 ) -> tuple[dict[str, Any], AnnData]:
     if match_rows == "left":
-        warnings.warn("Matching rows 'left' is not supported for 'right' join.", UserWarning, stacklevel=2)
+        warnings.warn(
+            "Matching rows 'left' is not supported for 'right' join; it will be treated as 'no'.",
+            UserWarning,
+            stacklevel=2,
+        )
+        match_rows = "no"
     regions, region_column_name, instance_key = get_table_keys(table)
     if isinstance(regions, str):
         regions = [regions]
@@ -412,6 +424,13 @@ def _left_exclusive_join_spatialelement_table(
     match_rows: Literal["left", "no", "right"],
     filter_label_pixels: bool | None = None,
 ) -> tuple[dict[str, Any], AnnData | None]:
+    if match_rows == "right":
+        warnings.warn(
+            "Matching rows 'right' is not supported for 'left_exclusive' join; it will be treated as 'no'.",
+            UserWarning,
+            stacklevel=2,
+        )
+        match_rows = "no"
     regions, region_column_name, instance_key = get_table_keys(table)
     if isinstance(regions, str):
         regions = [regions]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by LucaMarconato** [comment]
The fix implements a suggestion from @jan-glx "in the filter function, add row numbers, do the join, then sort by those row numbers, then drop".

Please either @timtreis or @jan-glx review; whoever does it first 😁

**Comment by LucaMarconato** [comment]
For the maintainers: I created and added the label "todo-add-extra-comments-on-release-notes", so when we auto-generate release notes we remember to add a comment mentioning potential order implications to users.

File: src/spatialdata/_core/query/relational_query.py [hunk 1]
```diff
@@ -146,17 +146,29 @@ def _filter_table_by_elements(table: AnnData | None, elements_dict: dict[str, di
...
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore", UserWarning)
-        _, filtered = join_spatialelement_table(
-            spatial_element_names=list(elements_by_name.keys()),
-            spatial_elements=list(elements_by_name.values()),
-            table=table,
-            how="left",
-        )
-    return filtered if filtered is not None and len(filtered) > 0 else None
+    # `join_spatialelement_table(how="left")` groups rows by region, which does not preserve the original
```
jan-glx: true. but I think it how="left" match="no" *should* preserve order (its essentially a semi-join for the table side.

tests/core/query/test_relational_query.py
```diff
@@ -940,6 +940,41 @@ def test_filter_table_categorical_bug(shapes):
...
+
+    before = list(sdata["table"].obs["label"])
+    filtered = sdata.filter_by_coordinate_system("global")
+    after = list(filtered["table"].obs["label"])
+    assert after == before
+    # the original table must not be mutated by the filtering operation
+    assert list(sdata["table"].obs.columns) == ["region", "instance_id", "label"]
+
+
 def test_filter_table_non_annotating(full_sdata):
```
jan-glx, line 1130: Test LGTM

src/spatialdata/_core/query/relational_query.py
```diff
@@ -146,17 +146,29 @@ def _filter_table_by_elements(table: AnnData | None, elements_dict: dict[str, di
...
+                spatial_element_names=list(elements_by_name.keys()),
+                spatial_elements=list(elements_by_name.values()),
+                table=table,
+                how="left",
+            )
+    finally:
+        del table.obs[order_col]
+    if filtered is None or len(filtered) == 0:
+        return None
+    filtered = filtered[np.argsort(filtered.obs[order_col].to_numpy()), :].copy()
```
jan-glx: Looks correct, but probably upstream fix is more efficient (no double copy)

jan-glx (commented):
Looks correct, but I think now upstream fix is the way to go.

jan-glx:
In [d61c0ca](https://github.com/scverse/spatialdata/pull/1193/commits/d61c0ca01563cf41c670250ba385614a09e0bfa9) I proposed an alternative solution that restores the row order for the table already in the semi join performed in `_left_join_spatialelement_table`

src/spatialdata/_core/query/relational_query.py
LucaMarconato: Good point.

tests/core/query/test_relational_query.py
LucaMarconato: I expanded them, please can you re-review?

src/spatialdata/_core/query/relational_query.py
LucaMarconato: Thanks for the upstream fix; I extended it to the inner join.

src/spatialdata/_core/query/relational_query.py
```diff
@@ -438,7 +467,12 @@ def _left_join_spatialelement_table(
     filter_label_pixels: bool | None = None,
 ) -> tuple[dict[str, Any], AnnData]:
     if match_rows == "right":
-        warnings.warn("Matching rows 'right' is not supported for 'left' join.", UserWarning, stacklevel=2)
+        warnings.warn(
+            "Matching rows 'right' is not supported for 'left' join; it will be treated as 'no'.",
```
LucaMarconato, line 471: Truth be told, for left join we could support the right match_rows, and for right join we could support the left match_rows. But we can skip it for now (since also it was not supported before this PR), and eventually do it in the future. 

For left_exclusive, right match_rows does not make sense, so it is not supported. Same for right_exclusive: left match_rows does not make sense there.

tests/core/query/test_relational_query.py
```diff
@@ -940,6 +944,166 @@ def test_filter_table_categorical_bug(shapes):
...
+            "right": _JoinOutcome(
+                table_order=["b2", "b1", "a2", "a1", "a0", "b0"],
+                warns=True,
+                element_index={"a": [2, 1, 0], "b": [1, 2, 0]},
+            ),
+        },
+        "left_exclusive": {
+            # TODO: make this test more interesting by adding indices 5, 4 to "a" and 4, 6 to "b"
+            # by design, "left_exclusive" never returns a table (only filtered elements), regardless of
+            # match_rows or whether anything was actually excluded.
```
LucaMarconato: We can do in this PR or leave for the future. It is beyond the scope of the original fix anyway.

LucaMarconato:
Thanks @jan-glx for the review. I upstreamed the fix as suggested and added extensive tests for all the join and match_rows combinations for tables with interleaved annotations of the elements.

In doing so I found out that:
* the reported bug was affecting also the inner join
* I improved the disclosure to users of unsupported join x match_rows combinations (via clear warning messages)
* found and fixed a bug for match_rows = 'no' for 'inner' and 'right' join, which was not leaving the index order untouched
* found and fixed a bug in `_left_exclusive_join_spatialelement_table` that was leading to an exception.

I kindly ask you for a re-review.

tests/core/query/test_relational_query.py
jan-glx: I added the two unmatched table rows in f7d31f82a64751fdc1d0b3018a43fcff0b518314 as suggested.

jan-glx:
I additionally added a unmatched region "c" to the test case table in 2c9c419cb8ac268bc25ad47b49217b3ced0bd226. This reveals a bug in right exclusive joins (that "c" row is missing in the output there). But this is out of scope of this PR, so I xfail them in d358535aa852fe00b6a9d3b1f368035288545971 .

jan-glx:
I also checked that, on pre-fix code, the extended test suite:
  - fails 3 cases (`inner-no`, `right-no`, `right-left`),
  - xfails the right_exclusive tests.

  After applying the 3-line [`_get_masked_element`](https://github.com/scverse/spatialdata/pull/1193/changes#diff-2e3532910877ce6aed075425645b4cfe1a9bc779290695039724524b31f019f1R239-R241) fix, the suite passes cleanly, confirming the fix resolves that ordering bug without regressions.

jan-glx (approved):

jan-glx:
(rewrote clean history, same end result)

PR #1197, opened 2026-08-29
Adopt the scverse cookiecutter template

Zethson:
Links the repository to `cookiecutter-scverse` via `.cruft.json`, and brings packaging, tooling, CI, issue templates and docs configuration in line with the template.

Adopting the template's mypy hook surfaced 500 errors, among them these bugs:

- `validate_table_in_spatialdata` guarded on `dtype is str`, which is never true of a dtype object, so the instance-key dtype mismatch has never been reported.
- `_resolve_zarr_store` branched on `zarr.storage.ConsolidatedMetadataStore`, which zarr 3 removed, so a group backed by anything other than a Local or Fsspec store raised `AttributeError` instead of the `ValueError` on the next line.
- `_resolve_zarr_store` then branched on `isinstance(path, zarr.storage.StoreLike)`, a PEP 695 alias that `isinstance` rejects with `TypeError`, so an already-resolved store never reached its branch.
- Writing a shapes element with a string index in the v01 format passed `object_codec` to `create_array`, which zarr 3 removed, raising `TypeError`.
- `_aggregate_image_by_labels` called `.compute()` on the result of `xrspatial.stats`, which is a pandas frame for numpy-backed values and only a dask one for dask-backed values.
- `ImageTilesDataset` used `.A` and `to_circles` used `.A1`, the scipy sparse aliases removed in 1.14.

Merging needs PyPI trusted publishing and codecov OIDC enabled on the repo, since `release.yaml` and `test.yaml` no longer use `PYPI_API_TOKEN` and `CODECOV_TOKEN`.


pyproject.toml
```diff
@@ -1,284 +1,422 @@
...
 name = "spatialdata"
 description = "Spatial data format."
-authors = [
-    {name = "scverse"},
-]
+readme = "README.md"
+license = { file = "LICENSE" }
 maintainers = [
-    {name = "scverse", email = "giov.pll@gmail.com"},
+  { name = "scverse", email = "giov.pll@gmail.com" },
```
Zethson, line 11: This should be Luca + team

.cruft.json
```diff
@@ -0,0 +1,44 @@
+{
+  "template": "https://github.com/scverse/cookiecutter-scverse",
+  "commit": "eb4523fde2e18bcfc121ff2cd722c9037ff8b910",
+  "checkout": null,
```
Zethson, line 4: Should not be null but rather main or 0.8.0

.cruft.json
```diff
@@ -0,0 +1,44 @@
...
+  "context": {
+    "cookiecutter": {
+      "project_name": "spatialdata",
+      "package_name": "spatialdata",
+      "project_description": "Spatial data format.",
+      "author_full_name": "scverse",
+      "author_email": "giov.pll@gmail.com",
```
Zethson, line 11: Luca or whose email?

pyproject.toml
```diff
@@ -1,284 +1,391 @@
...
+  "ome-zarr>=0.16",
+  "pandas",
+  "pooch",
+  "pyarrow",
+  "rich",
+  "scikit-image",
+  "scipy!=1.17",
+  "scverse-misc[datasets]>=0.1",
+  # for debug logging (referenced from the issue template)
+  "session-info2",
```
Zethson, line 48: Maybe this shouldn't be a runtime dependency

pyproject.toml
```diff
@@ -1,284 +1,391 @@
...
+  "setuptools",
```
Zethson, line 49: Still required? In 2026?

README.md
```diff
@@ -71,10 +70,6 @@ Update Feb 2025: `spatialdata` cannot be currently be installed via `conda` beca
 mamba install -c conda-forge spatialdata napari-spatialdata spatialdata-io spatialdata-plot
 ```
 
-## Limitations
```
Zethson, line 74: We can keep this if you like. I always point Windows users to WSL and don't bother with Windows but that's totally up to you

Zethson (commented):
We can remove the vscode config. Comes optionally from the template. No strong opinion

.cruft.json
LucaMarconato: going for "main"

.cruft.json
LucaMarconato: will add my scverse email

README.md
LucaMarconato: kept, but reworded and now mentioning WSL.

pyproject.toml
LucaMarconato: I'll update the list of maintainers to include the current ones: me, @melonora, @Tomaz-Vieira, @ajkswamy

pyproject.toml
LucaMarconato: yeah

pyproject.toml
LucaMarconato: can be removed

README.md
LucaMarconato: Actually we now test on Windows thanks to a recent PR from Wouter. I had forgotten about this. So we can remove.

LucaMarconato (approved):
Thanks @Zethson. Reviewed and applied some changes. Waiting for green to be auto-merged.

Zethson:
@LucaMarconato there's some pending tests because we changed the Github actions workflows. This will resolve itself when this is merged but it means that auto-merge will probably never complete now.

Thanks!

LucaMarconato:
Right, changed the required green check to ") Test / Tests pass in all hatch environments (pull_request)".

PR #1247, opened 2026-09-11
Fixes basic typing issues in rasterize.py

Tomaz-Vieira:
Cleans up `rasterize.py` so that it passes basic type checking with pyright.

# Rationale behind `not isinstance(...)`

Unfortunately, because python has a crazy (multi-)inheritance model, code like this:

```python
a: int = 123
assert isinstance(a, list)
```

will be accepted by type checkers (when not in strict mode), and will infer `a: <subclass of int and list>`, so we can end up accidentally asserting a type that should never be asserted. If we instead assert that the type is _not_ what it isn't supposed to be, the type checker can infer that it must be the correct type. And as a bonus, if ever the type of the asserted value changes, this negative assertion won't hide this from us by forcing the type of the variable to whatever was asserted.

In the future, we can simplify this by configuring the type checker to be stricter, but one step at a time =)

### DIFF
--- a/src/spatialdata/_core/operations/rasterize.py
+++ b/src/spatialdata/_core/operations/rasterize.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple, assert_never, assert_type, cast
 
 import numpy as np
 from dask.array import Array as DaskArray
 from dask.dataframe import DataFrame as DaskDataFrame
 from geopandas import GeoDataFrame
+from numpy.typing import DTypeLike
 from shapely import Point
 from xarray import DataArray, DataTree
 
@@ -46,6 +47,12 @@ from spatialdata.transformations.transformations import (
 VALUES_COLUMN = "__values_column"
 
 
+class TargetDimensions(NamedTuple):
+    target_width: float
+    target_height: float
+    target_depth: float | None
+
+
 def _compute_target_dimensions(
     spatial_axes: tuple[str, ...],
     min_coordinate: ListOrNDArrayFloating,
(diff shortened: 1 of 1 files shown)


Zethson (approved):
Not sure why my review got requested here but random comments:
1. Nothing wrong with also being pyright compliant, but just mypy is good enough for now. As discussed on the template, there's is a likely future where scverse will move to pyright, pyrefly, or ty. We'll see.
2. Instead of having lots of asserts, I'd encourage you to raise informative errors. It's more "pythonic".
3. I'm all for more dataclasses/namedtuples with descriptive names. Well done!

> In the future, we can simplify this by configuring the type checker to be stricter, but one step at a time =)

Yeah! Please feel free to use mypy in the stricted possible mode. But also note that you will have to fight the typechecker from time to time as they're all broken in their own ways.

LucaMarconato:
> Not sure why my review got requested here but random comments:

Just texted you in Zulip, but you saw the notification before the message 😁

Tomaz-Vieira:
> Instead of having lots of asserts, I'd encourage you to raise informative errors. It's more "pythonic".

For these kinds of errors, I'd rather stick to the asserts, because they're not operational errors; They are bugs, and a sign that we don't know what's going on anymore. So the asserts are there for the sake of sanity checking and also for keeping the type checker happy, which enables further sanity checking =D

Tomaz-Vieira:
I was wrong about strictness of the type checker. I had pyright's `reportUnnecessaryIsInstance` in mind, but that doesn't quite work the way I had in mind (mypy is extremely less configurable and has nothing close to it). So far, the best I could find to make the assert in such a way that doesn't force the type and is still kinda readable is:

```python
assert not isinstance(my_image, DataTree)
typing.assert_type(my_image, DataArray)
```

so for readability, you still get to see that `my_image` is of type `DataArray`, but if ever `my_image`'s type is changed to something different, we'll still get a type error on the code following the negative assertion.

Zethson:
Sorry I fatfingered the wrong button on my phone. I just wanted to unsubscribe 🙃😂

